PR #4, opened 2022-01-16
Fix compiler warning: type qualifiers ignored on function return type

vog:
This fixes the following compiler warning:

```
mkiocccentry.c:768:13: error: type qualifiers ignored on function return type [-Wignored-qualifiers]
 static char const * const strnull(char const * const str);
             ^~~~~
```


lcn2:
Many thanks!

lcn2:
BTW: We also had to apply 7eceda5 ( https://github.com/ioccc-src/mkiocccentry/commit/7eceda54a0bc60fa6ad6ee23c558f013dfaa09c0 ) to address the other parts of your patch.

Thanks again.

PR #17, opened 2022-01-20
Fixed remind_user() to print the correct commands.

ilyakurdyukov:
```
Now that we have formed the compressed tarball file, you
can remove the entry directory we have formed by executing:

    rm -rf -- --123\ 456/test-1

If you are curious, you may examine the newly created compressed tarball
by running the following command:

    (cd -- --123\ 456 && /bin/tar -Jtvf entry.test.1.1642655744.txz)
```


ilyakurdyukov:
If the working directory doesn't start with '-', then it displays the commands in the old format.

PR #23, opened 2022-01-22
The mkiocccentry code has been updated to higher C++ standards.

ilyakurdyukov:
Mostly fixes for passing string constants to a char* pointer and assigning a type to the result of malloc().

Now compiled by the C++ compiler without warnings: 
`make clobber all CC="g++"`


ilyakurdyukov:
This is a C contest, but people who have also worked with C++ won't be annoyed of passing string constants to non-const pointers.

PR #24, opened 2022-01-22
__func__ is not supported prior to C99

ilyakurdyukov:
To compile with the `-std=gnu89` option.


lcn2:
Thanks @ilyakurdyukov

lcn2:
For some reason this pull request didn't take.  Humm ...

PR #25, opened 2022-01-22
Compatibility with other compilers.

ilyakurdyukov:
TenDRA:

`  [C90 6.5.2]: The type 'suseconds_t' hasn't been declared.`

Better to avoid this type, it's uncommon. But `int` type is enough to hold microseconds.

Little C compiler:

`mkiocccentry.c:984: warning: missing return value`

Also needed for C89 compilers that don't know that exit() doesn't return.

Little C compiler:

```
mkiocccentry.c:1512: warning: overflow in converting constant expression from `int' to `unsigned long'
```

```c
static size_t
file_size(char const *path)
```
Here:
```c
	return -1;
```
I did this:
```c
	return ~(size_t)0;
```

I can't still fully compile with TenDRA due to the missing `ssize_t`. And linking with LCC is failed, but at least it compiles.

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -240,7 +240,7 @@ struct info {
      * time
      */
     time_t tstamp;		/* seconds since epoch when .info json was formed (see gettimeofday(2)) */
-    suseconds_t usec;		/* microseconds since the tstamp second */
+    int usec;			/* microseconds since the tstamp second */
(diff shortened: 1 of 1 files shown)


lcn2:
Thank you @ilyakurdyukov 

The file_size really wants to return a ssize_t.  Made changes to accommodate that in 7cfb0184feb17eb20ba04b943a0b38505c8d49fa

PR #26, opened 2022-01-22
Changed "U+2013 EN DASH" to "U+002D HYPHEN-MINUS".

ilyakurdyukov:
The TenDRA compiler found another issue: 

`		      "    A–Z a–z 0–9 . _ - +",`

```
mkiocccentry.c, line 3868: Error:
  [C90 6.1.3.4]: Character out of range.
```

And finally mkioccccentry is compiled with TenDRA but I can't link it properly.


lcn2:
Interesting catch @ilyakurdyukov , thanks

PR #31, opened 2022-01-26
Answer file fixes

xexyl:
The problem with opening it first is that if the entry directory already exists and we have created the answers file then the answers file will have invalid data. Thus we open it after and then write the previously obtained information to the file.


lcn2:
Thanks @xexyl

xexyl:
Just happy to help!

PR #33, opened 2022-01-26
Check remarks md typo

xexyl:
When remarks.md could not be found the message referred to prog.c rather than remarks.md.


lcn2:
Thanks @xexyl

PR #34, opened 2022-01-27
Ignoring warnings with answer file fixes

xexyl:
If answer file specified and user ignores warnings write the
confirmation to the answers file.

When ignoring warnings and specifying the answers file we print a
warning that if they change this (fix the warnings or cause additional
warnings) that the results are likely to be incorrect and strongly
recommend that they re-enter the answers.

When showing how to use the answers file to update an entry warn them
that if they had ignored warnings and fixed them (or if new warnings are
triggered) then they are strongly recommended to re-enter the answers.

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -1032,17 +1032,19 @@ main(int argc, char *argv[])
     /*
      * finalize the answers file
      */
-    if (answerp != NULL && a_flag_used == true) {
-	errno = 0;			/* pre-clear errno for errp() */
-	ret = fprintf(answerp, "y\n");
-	if (ret <= 0) {
-	    warnp(__func__, "unable to write confirming y to the answers file");
-	}
-	errno = 0;			/* pre-clear errno for errp() */
-	ret = fflush(answerp);
-	if (ret != 0) {
-	    errp(6, __func__, "error in flushing data to the answers file");
-	    not_reached();
+    if (answerp != NULL) {
+	if (a_flag_used == true) {
+	    errno = 0;			/* pre-clear errno for errp() */
+	    ret = fprintf(answerp, "y\n");
+	    if (ret <= 0) {
+		warnp(__func__, "unable to write confirming y to the answers file");
+	    }
+	    errno = 0;			/* pre-clear errno for errp() */
+	    ret = fflush(answerp);
+	    if (ret != 0) {
+		errp(6, __func__, "error in flushing data to the answers file");
+		not_reached();
+	    }
 	}
 	ret = fclose(answerp);
 	if (ret != 0) {
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -229,6 +229,7 @@ struct info {
     bool wordbuf_warning;	/* true ==> word buffer overflow detected */
     bool ungetc_warning;	/* true ==> ungetc warning detected */
     bool Makefile_override;	/* true ==> Makefile rule override requested */
+    bool warnings_ignored;	/* true ==> warnings were ignored */
     /*
      * filenames
      */
@@ -664,7 +665,7 @@ static bool is_dir(char const *path);
 static bool is_read(char const *path);
 static bool is_write(char const *path);
 static ssize_t file_size(char const *path);
-static void check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char const *prog_c);
+static void check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char const *prog_c, FILE *answerp);
 static ssize_t readline(char **linep, FILE * stream);
 static char *readline_dup(char **linep, bool strip, size_t *lenp, FILE * stream);
 static void sanity_chk(struct info *infop, char const *work_dir, char const *tar,
@@ -676,7 +677,7 @@ static char *get_contest_id(struct info *infop, bool *testp);
 static int get_entry_num(struct info *infop);
 static char *mk_entry_dir(char const *work_dir, char const *ioccc_id, int entry_num, char **tarball_path, time_t tstamp);
 static bool inspect_Makefile(char const *Makefile);
-static void check_Makefile(struct info *infop, char const *entry_dir, char const *cp, char const *Makefile);
+static void check_Makefile(struct info *infop, char const *entry_dir, char const *cp, char const *Makefile, FILE *answerp);
 static void check_remarks_md(struct info *infop, char const *entry_dir, char const *cp, char const *remarks_md);
 static char *base_name(char const *path);
 static void check_extra_data_files(struct info *infop, char const *entry_dir, char const *cp, int count, char **args);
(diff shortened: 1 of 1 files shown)


ilyakurdyukov:
It doesn't work like that!

Because you might not have had warnings before when writing the answer file, and they appear when the entry is updated and the answer file is used.

So it's better to add an option to remove confirmations. To use it with an answer file. And show warnings at the end if this option is used.

ilyakurdyukov:
Even better would be to add the option to read answers from the specified file, instead of an option that removes confirmations.

xexyl:
Yeah I did think of that. But I actually do warn at the end as well as when telling them how to use the file. 

To have the possibility to remove confirmation of the answers seriously complicates the idea. You would have to parse the file rather than redirect it. Also in this case they might as well just re-enter the information. 

The problem is it doesn’t actually parse the answer file and there’s no need to. As you said one could use tee but the benefit of my idea is that it can tell the user how to use it and how to do it rather than just leave it to them to come up with it; it’s built in. 

I personally don’t see it worth the effort when these entries are highly unlikely to win especially the 0 size prog.c. I only added them to be complete: if they do ignore warnings they can cause invalid input if they make changes and that’s why they’re warned about it.

ilyakurdyukov:
> To have the possibility to remove confirmation of the answers seriously complicates the idea.

No, it's not. Just don't ask for confirmation if the option is enabled and read from stdin as before. And don't write confirmations to the answer file.

xexyl:
> > To have the possibility to remove confirmation of the answers seriously complicates the idea.
> 
> No, it's not. Just don't ask for confirmation if the option is enabled and read from stdin as before. And don't write confirmations to the answer file.

But it doesn't read from stdin. Or I should say: stdin is from a different source (the file). It's automatic which was the entire point of it. If you have to re-enter the information again it defeats the purpose of saving the answers. That's why there's the warning - just in case.

ilyakurdyukov:
The problem is that when you use a response file, updated files can trigger warnings and things don't go according to plan.

xexyl:
I did just have another thought though that might make it even better. I'll work on it in a bit on a new branch. Anyway summary of this change is that it allows for the answer file to work even if you ignore warnings. It was simply a bug fix.

ilyakurdyukov:
Maybe I can do a better solution than warnings, but not right now, I'll look into it in a while. Maybe you will do it before.

xexyl:
Maybe. I think it's wise to give a warning though when they're actually ignoring warnings and using the answers file at the same time. 

The idea here is to automate answering the question so any change in answers necessitates re-entering the answers. That's just how it goes. But without this change if you ignore warnings you couldn't use the answers file since the answers to ignore warnings were not written to the answers file. 

Still I have an idea that might make it easier to at least use the answers file.

ilyakurdyukov:
I suggest this:

Add an option for answers reading mode.

When this mode - do not ask for confirmation. If any part is wrong, then don't ask to re-enter that part, exit immediately with an error. Print the Rule2a/2b sizes and warnings about them at the end.

And the answers writer should write only final answers and no confirmations.

xexyl:
The funny thing is before I saw this I did implement a read answer file mode. I suppose you’re suggesting that (but this is just a guess) when it’s in read from file mode (which in my latest pull request is visible) if there’s a problem immediately exit with an error. This might be something.

It might also be something to in this mode (before starting) warn them that if they have ignored warnings and they’ve made changes that it could result in incorrect results. I’ll have to play with this another time - if the pull requests are accepted maybe I can go from there. 

I do think having a read mode is a great idea and as I said I did it before I even noticed this message. But the rest is playing the guessing game of ‘what’s the best for those who are ignoring warnings’ and I’m not sure I know how to answer that; I don’t think there is a single answer to this question. Ignoring the warnings is a risk to begin with.

If you elaborate on your last sentence though maybe it’ll be clearer what you’re getting at; I’m not sure I followed it.

xexyl:
I'll be gone for many hours now but if you can elaborate on the last part I'd appreciate it. 

I do want to reiterate though that right now if one ignores a warning in the tool then they cannot use the answers file because I didn't add confirmation of ignoring warnings as I hadn't noticed that the warnings were allowed to be ignored. That should be fixed.

That's what this change does and it's the most important part of the change: it fixes it so that one use the answers file even if they ignore warnings. The warning them that if they change their answers it will change the results of the answers file is just as much worth noting when they're not ignoring warnings. Maybe there should be a more general note on this; I don't know.

lcn2:
I'm curious how the conversation goes, so since this is not urgent, I'll watch to is if your @xexyl & @ilyakurdyukov (and others) might have / suggest.   If things stabilize in a day or so .. or if folks call it "good enough" I'll merge the code.

ilyakurdyukov:
@lcn2 What do you think should be a good solution for this?

lcn2:
> @lcn2 What do you think should be a good solution for this?

If by '**this**' you mean got suggestion in https://github.com/ioccc-src/mkiocccentry/pull/34#issuecomment-1023477782 your idea of "_... writer should write only final answers and no confirmations_" seems reasonable.

Else I'm not sure which  '**this**'  you refer to.  :-)

ilyakurdyukov:
So my suggestion is to strip these yes/Y/y confirmations from the answer file:

```
test
0
title
abstract
1
author name
cc
yes
test@example.com
http://example.com/index.html
@twitter
@github
affiliation
Y
y
```

... and now I think, what if something messes up with the fields? For example, there is an additional field.

Then mkiocccentry should try to read the extra line, and if it can read the extra line, then return an error.

So it should be the correct line numbers in the answer file, not anything more attached:

```
test
0
title
abstract
1
author name
cc
test@example.com
http://example.com/index.html
@twitter
@github
affiliation
```

This with the extra string "12345" at the end should return an error: 
```
test
0
title
abstract
1
author name
cc
test@example.com
http://example.com/index.html
@twitter
@github
affiliation
12345
```

I'm leaving this to Cody because I don't feel very good right now to program. But I'll probably do it next week if Cody doesn't do it sooner.

lcn2:
On the other hand ... reconsidering https://github.com/ioccc-src/mkiocccentry/pull/34#issuecomment-1024055952 

The problem with changing it to NOT ask for continuation is that the saved answers (under -a file) are NO LONGER symmetric with the input (proposed -i file) OR they are no longer equivalent to the original input.

We will think about it / sleep on it: for now we lean towards ALWAYS requiring confirmation: write Y's into the answers AND supply them on input.

lcn2:
Perhaps the mkiocccentry-test.sh should be expanded to include testing of -i answers?

xexyl:
> On the other hand ... reconsidering [#34 (comment)](https://github.com/ioccc-src/mkiocccentry/pull/34#issuecomment-1024055952)
> 
> The problem with changing it to NOT ask for continuation is that the saved answers (under -a file) are NO LONGER symmetric with the input (proposed -i file) OR they are no longer equivalent to the original input.
> 
> We will process the pull request as is. If there is a compelling request to change it .. then it could be reconsidered.

Thanks for merging these. There was indeed a problem with not merging it namely that IF the user ignored the warnings then the answers files will NOT work because the code (before this commit) did not write those answers (ignoring the warnings) to the file.

Perhaps more warnings can be given however. Ilya had the thought that whenever the '-I' (laptop changing it, I mean lower-case '-I') is used a warning about the sizes can be given (if say the warnings were ignored - which I included in the code already in the info struct). Then again these warnings are already given so maybe if they are just give a final warning about it. I'll take a look at that too along with the rest of the comments.

xexyl:
> Perhaps the mkiocccentry-test.sh should be expanded to include testing of -i answers?

I can let Ilya do that since that was (I believe) his idea; I thought of the answers file immediately after I noticed that you had to re-enter the input every time since I update my entries quite a lot (usually documentation as you probably remember from my previous winning entries - I do it during the contest and post contest). Then he told me that he had done something with the script and I took a very quick look at that.

But since he did this he knows it better than I do so it would be better that he does it I think (unless he wants me to do it I guess).

PR #37, opened 2022-01-29
Rewritten answers file writing/reading.

ilyakurdyukov:
1. The answers file contains no more (y/n) confirmations.

2. Reading the answers file doesn't ask for confirmation.

3. Reading the answers file doesn't repeat questions again, exits with an error if there were problems.

4. Added a marker for the end of answers file and checking for its presence.

5. Reading the answers file doesn't print hints.

6. Fixed the echo of the lines read from the answers file.

### DIFF
--- a/mkiocccentry-test.sh
+++ b/mkiocccentry-test.sh
@@ -32,6 +32,8 @@ int main() { puts("Hello, World!"); }
 EOF
 }
 
+echo -n > "$src_dir"/empty_prog.c
+
 # fake some requireed files
 #
 test -f "$src_dir"/Makefile || cat Makefile.example > "$src_dir"/Makefile
@@ -76,7 +78,20 @@ fi
 #
 find "$work_dir_esc" -mindepth 1 -depth -delete
 
-./mkiocccentry -i answers.txt -- "$work_dir" "$src_dir"/{prog.c,Makefile,remarks.md,extra1,extra2}
+./mkiocccentry -i answers.txt -- "$work_dir" "$src_dir"/{empty_prog.c,Makefile,remarks.md,extra1,extra2}
+status="$?"
+if [[ $status -ne 0 ]]; then
+    echo "$0: FATAL: /mkiocccentry non-zero exit code: $status" 1>&2
+    exit "$status"
+fi
+
+# cleanout the under work_dir area, again
+#
+find "$work_dir_esc" -mindepth 1 -depth -delete
+
+echo yes >> answers.txt
+
+cat answers.txt | ./mkiocccentry -- "$work_dir" "$src_dir"/{empty_prog.c,Makefile,remarks.md,extra1,extra2}
 status="$?"
 if [[ $status -ne 0 ]]; then
     echo "$0: FATAL: /mkiocccentry non-zero exit code: $status" 1>&2
--- a/mkiocccentry-test.sh
+++ b/mkiocccentry-test.sh
@@ -61,9 +61,22 @@ y
 EOF
 }
 
+rm -f answers.txt
+
 # run the test, looking for an exit
 #
-answers | ./mkiocccentry -- "$work_dir" "$src_dir"/{prog.c,Makefile,remarks.md,extra1,extra2}
+answers | ./mkiocccentry -a answers.txt -- "$work_dir" "$src_dir"/{prog.c,Makefile,remarks.md,extra1,extra2}
+status="$?"
+if [[ $status -ne 0 ]]; then
+    echo "$0: FATAL: /mkiocccentry non-zero exit code: $status" 1>&2
+    exit "$status"
+fi
+
+# cleanout the under work_dir area, again
+#
+find "$work_dir_esc" -mindepth 1 -depth -delete
+
+./mkiocccentry -i answers.txt -- "$work_dir" "$src_dir"/{prog.c,Makefile,remarks.md,extra1,extra2}
 status="$?"
 if [[ $status -ne 0 ]]; then
     echo "$0: FATAL: /mkiocccentry non-zero exit code: $status" 1>&2
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -232,7 +232,6 @@ struct info {
     bool wordbuf_warning;	/* true ==> word buffer overflow detected */
     bool ungetc_warning;	/* true ==> ungetc warning detected */
     bool Makefile_override;	/* true ==> Makefile rule override requested */
-    bool warnings_ignored;	/* true ==> warnings were ignored */
     int answers_errors;		/* > 0 ==> flushing or closing answers file failed */
     /*
      * filenames
@@ -2979,7 +2978,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	}
 	infop->empty_override = true;
 	infop->rule_2a_override = false;
-	infop->warnings_ignored = true;
 
     /*
      * warn if prog.c is too large under Rule 2a
@@ -3010,7 +3008,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	}
 	infop->empty_override = false;
 	infop->rule_2a_override = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->empty_override = false;
 	infop->rule_2a_override = false;
@@ -3061,7 +3058,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	        (long)infop->rule_2a_size, (long)size.rule_2a_size);
 	}
 	infop->rule_2a_mismatch = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->rule_2a_mismatch = false;
     }
@@ -3085,7 +3081,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	    dbg(DBG_LOW, "user says that prog.c %s having character(s) with high bit is OK", prog_c);
 	}
 	infop->highbit_warning = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->highbit_warning = false;
     }
@@ -3109,7 +3104,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	    dbg(DBG_LOW, "user says that prog.c %s having NUL character(s) is OK", prog_c);
 	}
 	infop->nul_warning = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->nul_warning = false;
     }
@@ -3133,7 +3127,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	    dbg(DBG_LOW, "user says that prog.c %s having unknown or invalid trigraph(s) is OK", prog_c);
 	}
 	infop->trigraph_warning = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->trigraph_warning = false;
     }
@@ -3158,7 +3151,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	    dbg(DBG_LOW, "user says that prog.c %s triggered a word buffer overflow is OK", prog_c);
 	}
 	infop->wordbuf_warning = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->wordbuf_warning = false;
     }
@@ -3183,7 +3175,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	    dbg(DBG_LOW, "user says that prog.c %s triggered an ungetc warning OK", prog_c);
 	}
 	infop->ungetc_warning = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->ungetc_warning = false;
     }
@@ -3210,7 +3201,6 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	        (unsigned long)infop->rule_2b_size, (unsigned long)RULE_2B_SIZE);
 	}
 	infop->rule_2b_override = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->rule_2b_override = false;
     }
@@ -3646,7 +3636,6 @@ check_Makefile(struct info *infop, char const *entry_dir, char const *cp, char c
 	    }
 	}
 	infop->Makefile_override = true;
-	infop->warnings_ignored = true;
     } else {
 	infop->Makefile_override = false;
     }
@@ -6769,9 +6758,66 @@ remind_user(char const *work_dir, char const *entry_dir, char const *tar, char c
 	    }
 	}
 
-	if (infop->warnings_ignored == true) {
-	    errno = 0;	/* pre-clear errno for errp() */
-	    ret = printf("\nYou've ignored one or more warnings!\n");
+	if (infop->empty_override == true ||
+	    infop->rule_2a_override == true ||
+	    infop->rule_2a_mismatch == true ||
+	    infop->rule_2b_override == true ||
+	    infop->highbit_warning == true ||
+	    infop->nul_warning == true ||
+	    infop->trigraph_warning == true ||
+	    infop->wordbuf_warning == true ||
+	    infop->ungetc_warning == true ||
+	    infop->Makefile_override == true) {
+	    /* short summary on warnings */
+	    do {
+	        errno = 0;	/* pre-clear errno for errp() */
+		ret = printf("\nYou've ignored one or more warnings:\n");	    
+		if (ret <= 0) break;
+		if (infop->empty_override) {
+		    ret = printf("\nWARNING: prog.c is empty\n");
+		    if (ret <= 0) break;
+		}
+		if (infop->rule_2a_override) {
+		    ret = printf("\nWARNING: prog.c size: %ld > Rule 2a maximum: %ld\n",
+			(long)infop->rule_2a_size, (long)RULE_2A_SIZE);
+		    if (ret <= 0) break;
+		}
+		if (infop->rule_2a_mismatch) {
+		    ret = printf("\nWARNING: prog.c file size: %ld != rule_count function size\n",
+			(long)infop->rule_2a_size);
+		    if (ret <= 0) break;
+		}
+		if (infop->rule_2b_override) {
+		    ret = printf("\nWARNING: prog.c size: %ld > Rule 2b maximum: %ld\n",
+			(long)infop->rule_2b_size, (long)RULE_2B_SIZE);
+		    if (ret <= 0) break;
+		}
+		if (infop->highbit_warning) {
+		    ret = printf("\nWARNING: prog.c has character(s) with high bit set\n");
+		    if (ret <= 0) break;
+		}
+		if (infop->nul_warning) {
+		    ret = printf("\nWARNING: prog.c has NUL character(s)\n");
+		    if (ret <= 0) break;
+		}
+		if (infop->trigraph_warning) {
+		    ret = printf("\nWARNING: prog.c has unknown or invalid trigraph(s) found\n");
+		    if (ret <= 0) break;
+		}
+		if (infop->wordbuf_warning) {
+		    ret = printf("\nWARNING: prog.c triggered a word buffer overflow\n");
+		    if (ret <= 0) break;
+		}
+		if (infop->ungetc_warning) {
+		    ret = printf("\nWARNING: prog.c triggered an ungetc error\n");
+		    if (ret <= 0) break;
+		}
+		if (infop->Makefile_override) {
+		    ret = printf("\nWARNING: Makefile don't have all the required targets\n");
+		    if (ret <= 0) break;
+		}
+	    } while (0);
+
 	    if (ret <= 0) {
 		warnp(__func__, "unable to warn user about ignoring warnings when writing to the answer file");
 	    }
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -681,7 +681,7 @@ static void sanity_chk(struct info *infop, char const *work_dir, char const *tar
 static void para(char const *line, ...);
 static void fpara(FILE * stream, char const *line, ...);
 static char *prompt(char const *str, size_t *lenp);
-static char *get_contest_id(struct info *infop, bool *testp);
+static char *get_contest_id(struct info *infop, bool *testp, bool *i_flag_used);
 static int get_entry_num(struct info *infop);
 static char *mk_entry_dir(char const *work_dir, char const *ioccc_id, int entry_num, char **tarball_path, time_t tstamp);
 static bool inspect_Makefile(char const *Makefile);
@@ -916,8 +916,16 @@ main(int argc, char *argv[])
     /*
      * obtain the IOCCC contest ID
      */
-    info.ioccc_id = get_contest_id(&info, &test_mode);
+    info.ioccc_id = get_contest_id(&info, &test_mode, &i_flag_used);
     dbg(DBG_MED, "IOCCC contest ID: %s", info.ioccc_id);
+
+    /*
+     * found the answer file header in stdin
+     */
+    if (i_flag_used == true && answers == NULL) {
+	answerp = stdin;
+    }
+
     /*
      * obtain entry number
      */
@@ -942,6 +950,10 @@ main(int argc, char *argv[])
 	if (answerp == NULL) {
 	    errp(8, __func__, "cannot create answers file: %s", answers);
 	    not_reached();
+	}
+        ret = fprintf(answerp, "MKIOCCCENTRY_ANSWERS_V1\n");
+	if (ret <= 0) {
+	    warnp(__func__, "fprintf error printing header to the answers file");
 	}
     }
 
@@ -1093,10 +1105,12 @@ main(int argc, char *argv[])
 		info.answers_errors++;
 	    }
 	}
-	ret = fclose(answerp);
-	if (ret != 0) {
-	    warnp(__func__, "error in fclose to the answers file");
-	    info.answers_errors++;
+	if (answers != NULL) {
+	    ret = fclose(answerp);
+	    if (ret != 0) {
+	        warnp(__func__, "error in fclose to the answers file");
+	        info.answers_errors++;
+	    }
 	}
 	answerp = NULL;
     }
@@ -1105,8 +1119,6 @@ main(int argc, char *argv[])
      * remind the user about their answers file
      */
     if (answers != NULL) {
-	int yorn;
-
 	if (need_hints == true) {
 	    errno = 0;			/* pre-clear errno for errp() */
 	    ret = printf("\nTo more easily update this entry you can run:\n\n    ./mkiocccentry -i %s ...\n", answers);
@@ -1114,6 +1126,10 @@ main(int argc, char *argv[])
 		warnp(__func__, "unable to tell user how to more easily update entry");
 	    }
 	}
+    }
+
+    if (i_flag_used == true) {
+	int yorn;
 
 	if (info.answers_errors > 0) {
 	    errno = 0;	/* pre-clear errno for errp() */
@@ -2606,7 +2622,7 @@ prompt(char const *str, size_t *lenp)
  * This function does not return on error or if the contest ID is malformed.
  */
 static char *
-get_contest_id(struct info *infop, bool *testp)
+get_contest_id(struct info *infop, bool *testp, bool *i_flag_used)
 {
     char *malloc_ret;		/* malloced return string */
     size_t len;			/* input string length */
@@ -2616,6 +2632,7 @@ get_contest_id(struct info *infop, bool *testp)
     unsigned int variant = 0;	/* UUID variant hex character */
     char guard;			/* scanf guard to catch excess amount of input */
     size_t i;
+    bool seen_answers_header = false;
 
     /*
      * firewall
@@ -2659,6 +2676,24 @@ get_contest_id(struct info *infop, bool *testp)
 	 * prompt for the contest ID
 	 */
 	malloc_ret = prompt("Enter IOCCC contest ID or test", &len);
+	if (seen_answers_header == false && !strcmp(malloc_ret, "MKIOCCCENTRY_ANSWERS_V1")) {
+	    dbg(DBG_HIGH, "found answers header");
+	    seen_answers_header = true;
+	    *i_flag_used = true;
+	    need_retry = false;
+	    need_confirm = false;
+	    need_hints = false;
+	    prompt_fix_echo = true;
+
+	    free(malloc_ret);
+	    malloc_ret = prompt("", &len);
+	}
+	if (*i_flag_used == true && seen_answers_header == false) {
+	    dbg(DBG_HIGH, "the IOCCC contest ID as entered is: %s", malloc_ret);
+	    err(75, __func__, "didn't find the correct answers file header");
+	    not_reached();
+	}
+
 	dbg(DBG_HIGH, "the IOCCC contest ID as entered is: %s", malloc_ret);
 	ret = 0;		/* initialize paranoia */
 
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -1,3 +1,4 @@
+/* vim: set tabstop=8 softtabstop=4 shiftwidth=4 noexpandtab : */
 /*
  * mkiocccentry - make an ioccc entry
  *
@@ -232,6 +233,7 @@ struct info {
     bool ungetc_warning;	/* true ==> ungetc warning detected */
     bool Makefile_override;	/* true ==> Makefile rule override requested */
     bool warnings_ignored;	/* true ==> warnings were ignored */
+    int answers_errors;		/* > 0 ==> flushing or closing answers file failed */
     /*
      * filenames
      */
@@ -652,6 +654,10 @@ static struct location {
  * globals
  */
 int verbosity_level = DBG_DEFAULT;	/* debug level set by -v */
+static bool need_confirm = true;	/* true ==> ask for confirmations */
+static bool need_hints = true;		/* true ==> show hints */
+static bool prompt_fix_echo = false;    /* true ==> print the line read from the stream (fixes the log when using the -i option) */
+static bool need_retry = true;
 
 /*
  * forward declarations
@@ -667,8 +673,7 @@ static bool is_dir(char const *path);
 static bool is_read(char const *path);
 static bool is_write(char const *path);
 static ssize_t file_size(char const *path);
-static void check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char const *prog_c, FILE *answerp,
-			 bool i_flag_used, unsigned *answers_errors);
+static void check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char const *prog_c);
 static ssize_t readline(char **linep, FILE * stream);
 static char *readline_dup(char **linep, bool strip, size_t *lenp, FILE * stream);
 static void sanity_chk(struct info *infop, char const *work_dir, char const *tar,
@@ -680,8 +685,7 @@ static char *get_contest_id(struct info *infop, bool *testp);
 static int get_entry_num(struct info *infop);
 static char *mk_entry_dir(char const *work_dir, char const *ioccc_id, int entry_num, char **tarball_path, time_t tstamp);
 static bool inspect_Makefile(char const *Makefile);
-static void check_Makefile(struct info *infop, char const *entry_dir, char const *cp, char const *Makefile, FILE *answerp,
-			   bool i_flag_used, unsigned *answers_errors);
+static void check_Makefile(struct info *infop, char const *entry_dir, char const *cp, char const *Makefile);
 static void check_remarks_md(struct info *infop, char const *entry_dir, char const *cp, char const *remarks_md);
 static char *base_name(char const *path);
 static void check_extra_data_files(struct info *infop, char const *entry_dir, char const *cp, int count, char **args);
@@ -689,8 +693,7 @@ static char const *lookup_location_name(char const *upper_code);
 static bool yes_or_no(char const *question);
 static char *get_title(struct info *infop);
 static char *get_abstract(struct info *infop);
-static int get_author_info(struct info *infop, char *ioccc_id, struct author **author_set, FILE *answerp, bool i_flag_used, 
-	unsigned *answers_errors);
+static int get_author_info(struct info *infop, char *ioccc_id, struct author **author_set);
 static void verify_entry_dir(char const *entry_dir, char const *ls);
 static bool json_putc(int const c, FILE *stream);
 static bool json_fprintf_str(FILE *stream, char const *str);
@@ -705,7 +708,7 @@ static void write_info(struct info *infop, char const *entry_dir, bool test_mode
 static void write_author(struct info *infop, int author_count, struct author *authorp, char const *entry_dir);
 static void form_tarball(char const *work_dir, char const *entry_dir, char const *tarball_path, char const *tar, char const *ls);
 static void remind_user(char const *work_dir, char const *entry_dir, char const *tar, char const *tarball_path, bool test_mode,
-			char const *answers, struct info *infop, unsigned answers_errors);
+			char const *answers, struct info *infop);
 static char *cmdprintf(char const *format, ...);
 
 
@@ -738,7 +741,6 @@ main(int argc, char *argv[])
     bool l_flag_used = false;	/* true ==> -l /path/to/ls was given */
     bool a_flag_used = false;	/* true ==> -a write answers to answers file */
     bool i_flag_used = false;	/* true ==> -i read answers from answers file */
-    unsigned answers_errors = 0; /* > 0 ==> flushing or closing answers file failed */
     int ret;			/* libc return code */
     int i;
 
@@ -791,6 +793,9 @@ main(int argc, char *argv[])
 	case 'i':		/* -i input_recorded_answers */
 	    answers = optarg;
 	    i_flag_used = true;
+	    need_confirm = false;
+	    need_hints = false;
+	    prompt_fix_echo = true;
 	    break;
 	default:
 	    usage(1, __func__, "invalid -flag", program, tar, cp, ls); /*ooo*/
@@ -941,7 +946,7 @@ main(int argc, char *argv[])
     /*
      * write the IOCCC id and entry number to the answers file
      */
-    if (answerp != NULL) {
+    if (answerp != NULL && a_flag_used == true) {
 	errno = 0;			/* pre-clear errno for errp() */
         ret = fprintf(answerp, "%s\n", info.ioccc_id);
 	if (ret <= 0) {
@@ -958,14 +963,14 @@ main(int argc, char *argv[])
      * check prog.c
      */
     para("", "Checking prog.c ...", NULL);
-    check_prog_c(&info, entry_dir, cp, prog_c, answerp, i_flag_used, &answers_errors);
+    check_prog_c(&info, entry_dir, cp, prog_c);
     para("... completed prog.c check.", "", NULL);
 
     /*
      * check Makefile
      */
     para("Checking Makefile ...", NULL);
-    check_Makefile(&info, entry_dir, cp, Makefile, answerp, i_flag_used, &answers_errors);
+    check_Makefile(&info, entry_dir, cp, Makefile);
     para("... completed Makefile check.", "", NULL);
 
     /*
@@ -987,7 +992,7 @@ main(int argc, char *argv[])
      */
     info.title = get_title(&info);
     dbg(DBG_LOW, "entry title: %s", info.title);
-    if (answerp != NULL) {
+    if (answerp != NULL && a_flag_used == true) {
 	errno = 0;			/* pre-clear errno for errp() */
 	ret = fprintf(answerp, "%s\n", info.title);
 	if (ret <= 0) {
@@ -1001,7 +1006,7 @@ main(int argc, char *argv[])
      */
     info.abstract = get_abstract(&info);
     dbg(DBG_LOW, "entry abstract: %s", info.abstract);
-    if (answerp != NULL) {
+    if (answerp != NULL && a_flag_used == true) {
 	errno = 0;			/* pre-clear errno for errp() */
 	ret = fprintf(answerp, "%s\n", info.abstract);
 	if (ret <= 0) {
@@ -1013,9 +1018,37 @@ main(int argc, char *argv[])
     /*
      * obtain author information
      */
-    author_count = get_author_info(&info, info.ioccc_id, &author_set, answerp, i_flag_used, &answers_errors);
+    author_count = get_author_info(&info, info.ioccc_id, &author_set);
     dbg(DBG_LOW, "collected information on %d authors", author_count);
 
+    /*
+    * if we have an answers file, record the verified author information
+    */
+    if (answerp != NULL && i_flag_used == false) {
+	errno = 0;			/* pre-clear errno for errp() */
+        ret = fprintf(answerp, "%d\n", author_count);
+	if (ret <= 0) {
+	    warnp(__func__, "fprintf error printing IOCCC author count to the answers file");
+	    info.answers_errors++;
+	}
+
+	for (i = 0; i < author_count; i++) {
+	    errno = 0;			/* pre-clear errno for errp() */
+	    ret = fprintf(answerp,
+		"%s\n" "%s\n" "%s\n" "%s\n" "%s\n" "%s\n" "%s\n",
+		author_set[i].name,
+		author_set[i].location_code,
+		author_set[i].email,
+		author_set[i].url,
+		author_set[i].twitter,
+		author_set[i].github,
+		author_set[i].affiliation);
+	    if (ret <= 0) {
+		warnp(__func__, "fprintf error printing author info the answers file");
+		info.answers_errors++;
+	    }
+	}
+    }
 
     /*
      * write the .info.json file
@@ -1032,42 +1065,48 @@ main(int argc, char *argv[])
     para("... completed .author.json file.", "", NULL);
 
     /*
-     * form the .txz file
-     */
-    form_tarball(work_dir, entry_dir, tarball_path, tar, ls);
-
-    /*
      * finalize the answers file, writing final answers (if writing answers) and
      * then closing the stream.
      */
     if (answerp != NULL) {
-	if (a_flag_used == true) {
-	    errno = 0;			/* pre-clear errno for errp() */
-	    ret = fprintf(answerp, "y\n");
-	    if (ret <= 0) {
-		++answers_errors;
-		warnp(__func__, "unable to write confirming y to the answers file");
+	if (i_flag_used == true) {
+	    char *linep = NULL;
+	    char *line;
+	    bool error = true;
+
+	    line = readline_dup(&linep, true, NULL, answerp);
+	    if (linep != NULL) {
+		error = strcmp(line, "ANSWERS_EOF") != 0;
+		free(linep);
 	    }
-	    errno = 0;			/* pre-clear errno for errp() */
-	    ret = fflush(answerp);
-	    if (ret != 0) {
-		warnp(__func__, "unable to flush data to the answers file");
-		++answers_errors;
+	    if (error == true) {
+	        errp(1, __func__, "expected ANSWERS_EOF marker at the end of the answers file");
+	        not_reached();
+	    }
+	} else {
+	    ret = fprintf(answerp, "ANSWERS_EOF\n");
+	    if (ret <= 0) {
+	        warnp(__func__, "fprintf error printing ANSWERS_EOF marker to the answers file");
+		info.answers_errors++;
 	    }
 	}
 	ret = fclose(answerp);
 	if (ret != 0) {
 	    warnp(__func__, "error in fclose to the answers file");
-	    ++answers_errors;
+	    info.answers_errors++;
 	}
 	answerp = NULL;
     }
 
+    /*
+     * form the .txz file
+     */
+    form_tarball(work_dir, entry_dir, tarball_path, tar, ls);
 
     /*
      * remind user various things e.g., to upload (unless in test mode)
      */
-    remind_user(work_dir, entry_dir, tar, tarball_path, test_mode, answers, &info, answers_errors);
+    remind_user(work_dir, entry_dir, tar, tarball_path, test_mode, answers, &info);
 
     /*
      * free storage
@@ -2435,6 +2474,10 @@ prompt(char const *str, size_t *lenp)
 	linep = NULL;
     }
 
+    if (prompt_fix_echo == true) {
+	fprintf(stdout, "%s\n", buf);
+    }
+
     /*
      * return malloced input buffer
      */
@@ -2483,25 +2526,27 @@ get_contest_id(struct info *infop, bool *testp)
     /*
      * explain contest ID
      */
-    para("To submit entries to the IOCCC, you must a registered contestant and have received a",
-	 "IOCCC contest ID (via email) shortly after you have been successfully registered.",
-	 "If the IOCCC is open, you may register as a contestant. See:",
-	 "",
-	 NULL);
-    errno = 0;		/* pre-clear errno for errp() */
-    ret = fprintf(stderr, "    %s\n", IOCCC_REGISTER_URL);
-    if (ret <= 0) {
-	warnp(__func__, "fprintf error printing IOCCC_REGISTER_URL");
+    if (need_hints == true) {
+	para("To submit entries to the IOCCC, you must a registered contestant and have received a",
+	     "IOCCC contest ID (via email) shortly after you have been successfully registered.",
+	     "If the IOCCC is open, you may register as a contestant. See:",
+	     "",
+	     NULL);
+	errno = 0;		/* pre-clear errno for errp() */
+	ret = fprintf(stderr, "    %s\n", IOCCC_REGISTER_URL);
+	if (ret <= 0) {
+	    warnp(__func__, "fprintf error printing IOCCC_REGISTER_URL");
+	}
+	para("",
+	     "If you do not have an IOCCC contest ID and you with to test this program,",
+	     "you may use the special contest ID:",
+	     "",
+	     "    test",
+	     "",
+	     "Note you will not be able to submit the resulting compressed tarball when using test.",
+	     "",
+	     NULL);
     }
-    para("",
-	 "If you do not have an IOCCC contest ID and you with to test this program,",
-	 "you may use the special contest ID:",
-	 "",
-	 "    test",
-	 "",
-	 "Note you will not be able to submit the resulting compressed tarball when using test.",
-	 "",
-	 NULL);
 
     /*
      * keep asking for an entry number until we get a valid reply
@@ -2596,7 +2641,12 @@ get_contest_id(struct info *infop, bool *testp)
 		  "",
 		  NULL);
 	}
-    } while (ret != 8);
+    } while (ret != 8 && need_retry == true);
+
+    if (ret != 8) {
+	errp(100, __func__, "retry prompt is disabled");
+    }
+
     dbg(DBG_MED, "IOCCC contest ID is a UUID: %s", malloc_ret);
 
     /*
@@ -2647,18 +2697,20 @@ get_entry_num(struct info *infop)
 	/*
 	 * explain entry numbers
 	 */
-	errno = 0;		/* pre-clear errno for errp() */
-	ret = printf("\nYou are allowed to submit up to %d entries to a given IOCCC.\n", MAX_ENTRY_NUM + 1);
-	if (ret <= 0) {
-	    errp(77, __func__, "printf error printing number of entries allowed");
-	    not_reached();
+	if (need_hints == true) {
+	    errno = 0;		/* pre-clear errno for errp() */
+	    ret = printf("\nYou are allowed to submit up to %d entries to a given IOCCC.\n", MAX_ENTRY_NUM + 1);
+	    if (ret <= 0) {
+		errp(77, __func__, "printf error printing number of entries allowed");
+		not_reached();
+	    }
+	    para("",
+		 "As in C, Entry numbers start with 0.  If you are updating a previous entry, PLEASE",
+		 "use the same entry number that you previously uploaded so we know which entry we",
+		 "should replace. If this is your 1st entry to this given IOCCC, enter 0.",
+		 "",
+		 NULL);
 	}
-	para("",
-	     "As in C, Entry numbers start with 0.  If you are updating a previous entry, PLEASE",
-	     "use the same entry number that you previously uploaded so we know which entry we",
-	     "should replace. If this is your 1st entry to this given IOCCC, enter 0.",
-	     "",
-	     NULL);
 
 	/*
 	 * ask for the entry number
@@ -2687,7 +2739,11 @@ get_entry_num(struct info *infop)
 	    entry_str = NULL;
 	}
 
-    } while (entry_num < 0 || entry_num > MAX_ENTRY_NUM);
+    } while ((entry_num < 0 || entry_num > MAX_ENTRY_NUM) && need_retry == true);
+
+    if (entry_num < 0 || entry_num > MAX_ENTRY_NUM) {
+	errp(100, __func__, "retry prompt is disabled");
+    }
 
     /*
      * return the entry number
@@ -2824,15 +2880,11 @@ mk_entry_dir(char const *work_dir, char const *ioccc_id, int entry_num, char **t
  *      entry_dir       - newly created entry directory (by mk_entry_dir()) under work_dir
  *      cp              - cp utility path
  *      prog_c          - prog_c arg: given path to prog.c
- *      answerp		- answer file
- *      i_flag_used	- if -i was specified
- *      answers_errors  - > 0 I/O errors on answers file
  *
  * This function does not return on error.
  */
 static void
-check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char const *prog_c, FILE *answerp, bool i_flag_used,
-	unsigned *answers_errors)
+check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char const *prog_c)
 {
     FILE *prog_stream;		/* prog.c open file stream */
     struct iocccsize size;	/* rule_count() processing results */
@@ -2917,24 +2969,17 @@ check_prog_c(struct info *infop, char const *entry_dir, char const *cp, char con
 	      "entry prog.c is not another smallest self-replicating program.",
 	      "",
 	      NULL);
-	yorn = yes_or_no("Are you sure you want to submit an empty prog.c file? [yn]");
-	if (yorn == false) {
-	    err(93, __func__, "please fix your prog.c file: %s", prog_c);
-	    not_reached();
+	if (need_confirm == true) {
+	    yorn = yes_or_no("Are you sure you want to submit an empty prog.c file? [yn]");
+	    if (yorn == false) {
+		err(93, __func__, "please fix your prog.c file: %s", prog_c);
+		not_reached();
+	    }
+	    dbg(DBG_LOW, "user says that their empty prog.c: %s is OK", prog_c);
 	}
-	dbg(DBG_LOW, "user says that their empty prog.c: %s is OK", prog_c);
 	infop->empty_override = true;
 	infop->rule_2a_override = false;
 	infop->warnings_ignored = true;
-	if (answerp != NULL && i_flag_used == false) {
-	    errno = 0;			/* pre-clear errno for errp() */
-	    ret = fprintf(answerp, "y\n");
-	    if (ret <= 0) {
-		warnp(__func__, "fprintf error writing user confirmation that empty prog.c is OK to answers file");
-		(*answers_errors)++;
-	    }
-	}
-
 
     /*
      * warn if prog.c is too large under Rule 2a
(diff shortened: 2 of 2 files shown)


ilyakurdyukov:
There are some broken error codes to fix and error messages to improve.

xexyl:
Disclaimer: I didn't really look at the code (though I see that you wrote to the answers file after everything was done and that could work) and there could be a language barrier between us.

> * The answers file contains no more (y/n) confirmations.

The problem here is if it doesn't have the confirmations then the answers file will not work because the questions have to be answered to proceed; if the answers file no longer has them then how can the questions be answered?

> * Reading the answers file doesn't ask for confirmation.

I don't know what you mean by this: the only code that asked for confirmation was the check functions that I simply added answers to. You can add the answers later of course but the answers file didn't ask for confirmation; the tool did.

> * Reading the answers file doesn't repeat questions again, exits with an error if there were problems.

Same as above.

> * Reading the answers file doesn't print hints.

Then if they change their answers (or files that no longer have these problems) they won't be alerted to a possible problem. When something is automated at the same time as warnings being ignored the user should be alerted to the possible problem. I would argue that whenever warnings are ignored for this contest they should be alerted to a possible problem. People are used to automating things without thinking and this could result in problems.


> * Fixed the echo of the lines read from the answers file.

If this is what I think it is I like that.

xexyl:
Oh and answers errors should be there because if there are I/O errors on the answers file then they obviously cannot use the answers file and expect valid results. Removing that check means that the user will not be alerted to possible problems. But maybe you added that later on and I didn't see (I just took a quick glance).

lcn2:
We think @xexyl has made some valid points above.

Moreover we are not sure that avoiding the questions is a good thing.  It certainly complicates the code to try and avoid them.

When we just had the "-a answers" feature, one could do this:

./mkiocccentry .. < answers

We suspect that any "-i answers" mode should simply be another way to accomplish feeding the answers file into stdin.

Yes, if someone were to edit the answers file, such as supplying a sting that is too long, or too short, etc. one would get the desired result when getting that modified answers file back on stdin (or via "-i answers").  But we don't think the code should be complicated to try and handle such a case.

Please recall that this mkiocccentry code is something an IOCCC contestant would run a few times per year.  We, the IOCCC judges, have to do a lot of work for the number of entries we do receive for a given IOCCC.  That is why we limit try and limit the number of entries someone can submit for a given contest.  And of course the submitter has to do a huge amount of work to just write also those entries to feed into mkiocccentry!

While it is fun to polish the code, using mkiocccentry efficiently might not be a significant goal to focus on.

[[see our next comment in this thread for what we are focused on now]]

While we are still pondering this issue, we suggest that it might be better if:

./mkiocccentry -i answers work_dir ...

Behave exactly like:

./mkioccentry work_dir ... < answers

And NOT try to avoid the questions and prompts and hints, maybe?

Something to ponder.  Comments welcome.

See our next comment for our current focus.

ilyakurdyukov:
> I don't know what you mean by this: the only code that asked for confirmation was the check functions that I simply added answers to. You can add the answers later of course but the answers file didn't ask for confirmation; the tool did.

By confirmation, I mean asking for yes/no.

> Then if they change their answers (or files that no longer have these problems) they won't be alerted to a possible problem. 

No. The user receives a warning with a list of problems at the end.

>> Fixed the echo of the lines read from the answers file.
>
> If this is what I think it is I like that.

The log becomes a complete unreadable mess without this fix.

> ./mkiocccentry .. < answers

If someone wants to use this they can use `tee answers | ./mkiocccentry ..` to get answers in this format.

And if there is exactly the same as stdin copy, then I don't see the point of mkioccccentry having the `-a` option if it just copies the existing tool.

> Added a marker for the end of answers file and checking for its presence.

This is used as a form of confirmation to ensure that the answer file was read correctly.

ilyakurdyukov:
The reason I hide "hints" is because these hints are too over the top and annoying. When using automation like this, I don't want to scroll through multiple screens of things I already know looking for warnings. So I printed out a list of warnings at the end. So I can just look at the last screen in the console without scrolling to see if something bad happen or all is OK.

If someone is using saved answers, that means that person has already read all of these explanations.

xexyl:
> 
> > Then if they change their answers (or files that no longer have these problems) they won't be alerted to a possible problem.
> 
> No. The user receives a warning with a list of problems at the end.

I see that now as I tried it. But there was no need to make this change. Functionally you removed the warnings (and warnings are good to have) and didn't do anything I didn't do (well you added the EOF marker but I mean the results are the same); you just went about it a different way. You give warnings at the end of the output rather than in the middle of it. I had warnings too and warnings are important.


> The log becomes a complete unreadable mess without this fix.
> 
> > ./mkiocccentry .. < answers
> 
> If someone wants to use this they can use `tee answers | ./mkiocccentry ..` to get answers in this format.

This doesn't mean it's not good to have it built-in. I already gave reasons why it's good to have it built-in. Just because someone could use tee doesn't mean that it's not good to have it built-in.

> 
> And if there is exactly the same as stdin copy, then I don't see the point of mkioccccentry having the `-a` option if it just copies the existing tool.

Then why even work your changes if you don't see the point? I already gave reasons it's good to have it built-in. It takes the burden from the user: you can rely on the tool itself rather than a pipeline. Also tee will happily overwrite the file and that's easy enough to do if you do it quickly or without thinking. By having the options it ensures that this won't happen.

xexyl:
> The reason I hide "hints" is because these hints are too over the top and annoying. When using automation like this, I don't want to scroll through multiple screens of things I already know looking for warnings. So I printed out a list of warnings at the end. So I can just look at the last screen in the console without scrolling to see if something bad happen or all is OK.

I think this is a problem with you not being able to focus on the important bits - the warnings are important for users who ignore warnings. How far do you go? One could use your logic (below) that if they have seen it once then you shouldn't have the warnings of possibly invalid files at the end when using the answers file.

> 
> If someone is using saved answers, that means that person has already read all of these explanations.

You have far too much faith in users if you think like that! Besides that it has to be done at first for them to know it. But if you think people will remember it or they won't make a mistake when ignoring warnings you don't know users well enough. As a game developer the things I saw...

You're not likely to have to refer to the warnings because you're not likely to submit those files that trigger warnings anyway. And having warnings is always good. Always.

xexyl:
> While we are still pondering this issue, we suggest that it might be better if:
> 
> ./mkiocccentry -i answers work_dir ...
> 
> Behave exactly like:
> 
> ./mkioccentry work_dir ... < answers
> 
> And NOT try to avoid the questions and prompts and hints, maybe?

It already does. Before I added the second option that's what you could do (you could also use cat answers | ... but that's unnecessary use of cat).

xexyl:
> > While we are still pondering this issue, we suggest that it might be better if:
> > ./mkiocccentry -i answers work_dir ...
> > Behave exactly like:
> > ./mkioccentry work_dir ... < answers
> > And NOT try to avoid the questions and prompts and hints, maybe?
> 
> It already does. Before I added the second option that's what you could do (you could also use cat answers | ... but that's unnecessary use of cat).

However trying it with Ilya's code using the option works okay; however with the < I get an error:

Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
Are you sure you want to submit an empty prog.c file? [yn]: FATAL[74]: prompt: EOF while reading prompt input

...so it doesn't even form the tarball.

xexyl:
One more thing I noticed is it seems that once again fclose() on answer file is done before forming the tarball. 

Yesterday I moved the fclose() call to be after because freopen() makes stdin refer to the answers file and you can get bad file descriptor by closing the answers file as the function verify_entry_dir prompts for input. I did not notice this happening every time though.

Unfortunately the git log didn't show up so all you see is that I had some minor fixes (I added the bad file descriptor fix after starting the other change). I specifically stated the above part though in the git log on my branch.

lcn2:
# We need help with more tools in this repo

You can help.  See below.

# We are working hard to be able to run a MOCK IOCCC in 2022.

> By MOCK IOCCC we mean a beta test run of an IOCCC where entries submitted would NOT be judged.  Instead the IOCCC judges would go though the motions of running mock contest in order to beta test the process.  Nobody would win the MOCK IOCCC, although we would certainly thank those who helped test the process with special thanks to those who found something in the MOCK process that we need to fix.

> Any important lessons learned during the MOCK IOCCC would result in changes to the real IOCCC28.

> Our goal is to hold the MOCK IOCCC in 2022 and then start the reason IOCCC28 in 2022.

There is A LOT of work that goes on behind the scenes by the IOCCC judges.  You can help the IOCCC judges by helping us with the following tools related to mkiocccentry.  See below.

# JSON sanity check tools

- Write a tool that attempts validate (performs sanity checks) on the .info.json produced by mkiocccentry.
- Write a tool that attempts validate (performs sanity checks) on the .author.json produced by mkiocccentry

Those two tools could be written as pull requests to add them as separate programs to this repo.  Moreover mkiocccentry could be modified to perform a function call into that code (in the same way the code iocccsize function is executed directly from mkiocccentry while iocccsize is also compiled as a separate tool).

Even though mkiocccenty would run the core of those tools as function calls, we still need the tools as standalone programs for use elsewhere.  So both mkiocccentry callable functions AND standalone tools are needed.

The standalone tools should take a file as an argument. If sane, these tools should exit 0, otherwise exit non-zero.

# compressed tarball filename sanity check tool

- Write a tool that validates the filename of a compressed tarball file.

The mkiocccenty tool forms a compressed tarball file with a very specific form of a filename (for internal IOCCC judging process reasons).  We need a tool that performs a sanity check on the filename of a compressed tarball file.

That tool could be added to this repo as a pull request.  Moreover mkiocccentry could be modified to perform a function call into that code (in the same way the code iocccsize function is executed directly from mkiocccentry while iocccsize is also compiled as a separate tool) as a sanity check of what mkiocccentry produced.

Even though mkiocccenty would run the core of those tools as function calls, we still need the tool as standalone programs for use elsewhere.  So both mkiocccentry callable functions AND standalone tool are needed.

The standalone tool should take a file as an argument. If sane, these tool should exit 0, otherwise exit non-zero.

# compressed tarball sanity check tool

- Write a tool to validate a compressed tarball produced by mkiocccentry

Such a compressed tarball tool would need to perform safety and sanity checks on the compressed tarball WITHOUT uncompressing it on disk. It would need to be paranoid, expecting the worst while attempting to validate for the best.

For example, it could run "tar -t" and look at the output to verify that the files un-tar into a sub-directory (i.e., not wander off with ../../../foo path, not attempting to write something under /foo).  

It would also call the code of the "validate the filename of the compressed tarball file" (in the same way the code iocccsize function is executed directly from mkiocccentry while iocccsize is also compiled as a separate tool).

This tool would need to verify that the sub-directory that the compressed tarball would produce is related to the previously validated compressed tarball filename.

This compressed tarball tool would need to perform the two checks on the tarball size both checking the size of the compressed tarball and tar reports as the overall size of the elements of the tarball (see limit_ioccc.h). 

This sanity check tool should verify that the compressed tarball has the required files (prog.c, Makefile, remarks.md, .info.json, .author.json).

The sanity check tool should be sure that no other files beginning with "." would be created.  It should check that no sub-sub-directories would be created, just the entry directory from which the tarball was created in the first place.  It should check that all files created are UNDER the entry directory only.

That tool could be added to this repo as a pull request.  Moreover mkiocccentry could be modified to perform a function call into that code (in the same way the code iocccsize function is executed directly from mkiocccentry while iocccsize is also compiled as a separate tool) as a sanity check of what mkiocccentry produced.

Even though mkiocccenty would run the core of those tools as function calls, we still need the tool as standalone programs for use elsewhere.  So both mkiocccentry callable functions AND standalone tool are needed.

BTW: The reason why mkiocccentry should be able to execute the sanity check code of the above mention tools is that we (the IOCCC judges) plan to use those standalone tools in our judging procedures.  Nevertheless mkiocccentry should perform the same sanity checks to be help be sure what is generates will later will be checked by the IOCCC judges via the standalone tools.  

We don't want a mismatch of what mkiocccentry produces to conflict with the result of the sanity check standalone tools.  We don't want someone's IOCCC entry they worked hard on, invalidated because mkiocccenty produced something that the standalone tool later rejected.  That is why mkiocccentry should call on the sanity check code to validate what it is doing / creating.

Instead of loading up mkiocccentry with fork/exec of calling standalone tools and parsing the results, also allow function calls from mkiocccentry to run the same sanity checking code.

The standalone compress tarball sanity check tool should take a file as an argument. If sane, these tool should exit 0, otherwise exit non-zero.

# In general

These sanity check tools need to be written in a clean, highly readable, trivial to understand style.  Ironic with such a request given that this is in support of Obfuscated C, yes, but IOCCC is about irony in a way.

We want it to be easy for us and others to inspect what this code is doing.  We want it easy for this code to be modified and improved on in the future.

We want this code to stand as a good example of how to write well written, easy to understand, easy to improve C code.

Use of the dbg facility, and the -v level debugging style used by mkiocccentry will be important.  And if a similar coding style with ALL libc function calls checked for errors is also important.

And as a C contest, we would highly prefer these tools be written in portable C and compiled with gcc or clang without warnings.

# In summary

May we suggest that instead of focusing in issues related to "-i answers", that one consider writing the above tools AND modifying mkiocccentry to use that functionality?

Your help wanted AND is appreciated!

lcn2:
> > > While we are still pondering this issue, we suggest that it might be better if:
> > > ./mkiocccentry -i answers work_dir ...
> > > Behave exactly like:
> > > ./mkioccentry work_dir ... < answers
> > > And NOT try to avoid the questions and prompts and hints, maybe?
> > 
> > 
> > It already does. Before I added the second option that's what you could do (you could also use cat answers | ... but that's unnecessary use of cat).
> 
> But trying it with Ilya's code using the option works okay; however with the < I get an error:
> 
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > Are you sure you want to submit an empty prog.c file? [yn]: FATAL[74]: prompt: EOF while reading prompt input
> 
> ...so it doesn't even form the tarball.

By "Ilya's code" do you mean the code at the top of the master branch now, or do you mean as modified with the proposed #37 pull request as it is now?

lcn2:
> Yesterday I moved the fclose() call to be after because freopen() makes stdin refer to the answers file and you can get bad file descriptor by closing the answers file as the function verify_entry_dir prompts for input. I did not notice this happening every time though.
> 
> Unfortunately the git log didn't show up so all you see is that I had some minor fixes (I added the bad file descriptor fix after starting the other change). I specifically stated the above part though in the git log on my branch.

That certainly need to be addressed ...

(along with https://github.com/ioccc-src/mkiocccentry/pull/37#issuecomment-1024915633) :-)

ilyakurdyukov:
> You give warnings at the end of the output rather than in the middle of it.

No, warnings in the middle are shown just as before.

ilyakurdyukov:
> But trying it with Ilya's code using the option works okay; however with the < I get an error:

With redirection to standard input without the -i option? You should not do this and should not expect it to work.

ilyakurdyukov:
> Then why even work your changes if you don't see the point?

My changes make sense because it's more than redirecting the answers to or from a file. It also makes it easier and prevents the answers from being misread if there a warnings happen/not happen.

xexyl:
> > > > While we are still pondering this issue, we suggest that it might be better if:
> > > > ./mkiocccentry -i answers work_dir ...
> > > > Behave exactly like:
> > > > ./mkioccentry work_dir ... < answers
> > > > And NOT try to avoid the questions and prompts and hints, maybe?
> > > 
> > > 
> > > It already does. Before I added the second option that's what you could do (you could also use cat answers | ... but that's unnecessary use of cat).
> > 
> > 
> > But trying it with Ilya's code using the option works okay; however with the < I get an error:
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: Please enter either y (yes) or n (no)
> > > Are you sure you want to submit an empty prog.c file? [yn]: FATAL[74]: prompt: EOF while reading prompt input
> > 
> > 
> > ...so it doesn't even form the tarball.
> 
> By "But trying it with Ilya's code" do you mean the code at the top of the master branch now, or do you mean as modified with the proposed #37 pull request as it is now?

I meant his patch - #37.

xexyl:
> > Yesterday I moved the fclose() call to be after because freopen() makes stdin refer to the answers file and you can get bad file descriptor by closing the answers file as the function verify_entry_dir prompts for input. I did not notice this happening every time though.
> > Unfortunately the git log didn't show up so all you see is that I had some minor fixes (I added the bad file descriptor fix after starting the other change). I specifically stated the above part though in the git log on my branch.
> 
> That certainly needs to be addressed ...
> 
> (along with [#37 (comment)](https://github.com/ioccc-src/mkiocccentry/pull/37#issuecomment-1024915633)) :-)

How do you mean? I already moved fclose() to after the form tarball function but in Ilya's patch it was moved back to before forming the tarball. Or do you mean I should make a commit so that there is a git log of it?

xexyl:
> > But trying it with Ilya's code using the option works okay; however with the < I get an error:
> 
> With redirection to standard input without the -i option? You should not do this and should not expect it to work.

Right but Landon suggested it has the same behaviour and yours doesn't. That's what I was saying; yours errors out because you removed the confirmations from the answers file.

xexyl:
> > Then why even work your changes if you don't see the point?
> 
> My changes make sense because it's more than redirecting the answers to or from a file. It also makes it easier and prevents the answers from being misread if there a warnings happen/not happen. That's why it's in a slightly different format than just a copy of stdin.

Mine makes just as much sense but has more warnings which is a good thing when ignoring warnings. The copy of stdin means one doesn't have to rely on the program but they can. It has the same behaviour. Your change does not make it easier: you still use the options along with the answers file. The difference is you omit confirming the warnings by the tool and that's not a good idea. 

But you still have to initially answer that you want to continue anyway - after that (using the answers file) you don't have to (just as with mine).

The tarball is formed fine under most circumstances in your code but I gave one where it isn't because you changed the format (namely removing the 'y' lines).

ilyakurdyukov:
> Right but Landon suggested it has the same behaviour and yours doesn't. That's what I was saying; yours errors out because you removed the confirmations from the answers file.

Since this is a fix, confirmations may appear in different numbers depending on warnings, but answers file will work in any cases now (with the -i option, using stdin is deprecated) and not depend on the warnings anymore.

ilyakurdyukov:
> How do you mean? I already moved fclose() to after the form tarball function but in Ilya's patch it was moved back to before forming the tarball. Or do you mean I should make a commit so that there is a git log of it?

It doesn't need to be after tarball function now.

ilyakurdyukov:
I can make < work with the new format if you insist. By simply adding a header to the answers file. So if mkioccccentry sees this header in the first prompt(), then interpret it as an answers file from the "-i" option.

lcn2:
You (@ilyakurdyukov and @xexyl) are having a good discussion.

Based on some of your replies, we seem to have been mistaken when we wrote some of our previous comments.

May we suggest that you try to arrive at some common recommendation if possible?  For example does the pull request meet your common ideas or does it need to be further updated?

lcn2:
> I can make < work with the new format if you insist. By simply adding a header to the answers file. So if mkioccccentry sees this header in the first prompt(), then interpret it as an answers file from the "-i" option.

That seems like an interest possibility.  What are the downsides to doing that, if any?

ilyakurdyukov:
> That seems like an interest possibility. What are the downsides to doing that, if any?

I want to include the version in this header, this might be a nice feature if something changes later in the tool. So the version in that header will also change and the newer tool will say that the answer file is in an older format and should be re-entered.

But change the version in the header only if there are incompatible changes, keep the same if changes in the tool do not change the field order or introduce a new fields/questions.

ilyakurdyukov:
> You (@ilyakurdyukov and @xexyl) are having a good discussion.

Cody and I independently and at the same time came up with the idea that we needed automation to avoid re-entering questions to update the entry. Because entering the same thing again is annoying and also can lead to errors when you repeat the same thing over and over again.

xexyl:
> > That seems like an interest possibility. What are the downsides to doing that, if any?
> 
> I want to include the version in this header, this might be a nice feature if something changes later in the tool. So the version in that header will also change and the newer tool will say that the answer file is in an older format and should be re-entered.
> 
> But change the version in the header only if there are incompatible changes, keep the same if changes in the tool do not change the field order or introduce a new fields/questions.

By all means add a version. But is it necessary to rewrite everything I did? As Landon said there are reasons to not ignore the confirmation of the answers. I have given you reasons too.

There's also another issue with your patch that I mentioned above (one I fixed in my last commit but which seems to be reverted).

It'd be better, I think, to move on and work on the parts that need to be done.

It seems that part of the problem you're having with my change is the warnings. If this is the case why don't you just change it (or I can) to write to stderr so you can in your scripts (I'm sure you'll be writing scripts to automate the process of creating the initial tarball and updating it - I definitely will be and maybe you'd like to even add a script template that helps people do it?) redirect it to /dev/null.

That would solve your worries and it would keep the rest as it is. You're right that if a new field is added it could break the file but keep in mind that this will happen whether you have a version in it or not: the user will still have to re-input information if a new field is added. You can't get past that. That's the nature of recording information in a file. If the input expected is no longer the same then obviously the new information has to be input.

xexyl:
> > I can make < work with the new format if you insist. By simply adding a header to the answers file. So if mkioccccentry sees this header in the first prompt(), then interpret it as an answers file from the "-i" option.
> 
> That seems like an interest possibility. What are the downsides to doing that, if any?

There's still another problem in the patch - one I fixed yesterday. ``fclose()`` MUST be called _after_ forming the tarball!

It seems like this is only complicating a process that already works. Why is there even a need to do that? It'd be better to work on things that need to be added. 

I'll reiterate: no matter what is done if the required input changes the user will have to re-enter their input. It doesn't matter what the input file looks like. There's no getting past this. You can in some cases write a conversion tool but that doesn't change the fact that new input has to be put in. In the case of the contest the user has to input it. How can the answers file magically have the answer to a new question in any format? It can't.

Maybe Ilya could make a tool that updates the answer file instead of trying to reinvent the answers file? A conversion tool or something? That seems more useful to me.

ilyakurdyukov:
> But is it necessary to rewrite everything I did?

Half, not everything, it's done on top of your pull requests. 

> If the input expected is no longer the same then obviously the new information has to be input.

Not certainly in that way. If you have two entries, you can re-enter just one and see what has changed in the answer file and edit the answers for the second entry accordingly. This is what I would do in such a case.

xexyl:
> > You (@ilyakurdyukov and @xexyl) are having a good discussion.
> 
> Cody and I independently and at the same time came up with the idea that we needed automation to avoid re-entering questions to update the entry. Because entering the same thing again is annoying and also can lead to errors when you repeat the same thing over and over again.

Right. It was the very first thought I had. You came up with a script and I thought of a file. I did think of another way too but the file seemed the better approach.

ilyakurdyukov:
> There's still another problem in the patch - one I fixed yesterday. fclose() MUST be called after forming the tarball!

What is the problem that this fixes?

> Maybe Ilya could make a tool that updates the answer file instead of trying to reinvent the answers file? A conversion tool or something? That seems more useful to me.

Warning confirmations cannot be predicted, so the format must be clear.

xexyl:
> > But is it necessary to rewrite everything I did?
> 
> Half, not everything, it's done on top of your pull requests.
> 

I see. But still it seems like you went through a lot of work that's unnecessary. I know you really focused on the warnings and I gave a solution to that - write it to stderr so you can redirect it to /dev/null (as I expect you won't need to know the information that goes there).

> > If the input expected is no longer the same then obviously the new information has to be input.
> 
> Not certainly in that way. If you have two entries, you can re-enter just one and see what has changed in the answer file and edit the answers for the second entry accordingly. This is what I would do in such a case.

You could manually edit the answers file. This is true. But you still have to input the new answers. Until you do the answers file is outdated. Having a version can tell you this, it is true, but it won't prevent you from having to enter the new information. Maybe having the version is a good idea for this reason - you could warn them. Why don't you do that? But it's wise to write the confirmation of the ignoring warnings. 

At the same time there's no way you can use the same answers file for more than one entry - unless you write to more than one answer file. For example one thought I had is that instead of having the author information tied to the .author.json file it could be tied to the contestant ID. I almost sent that suggestion to the judges. If this was done it would simplify adding/updating entries quite a lot.

lcn2:
We tested the patch we pulled it just a few motets ago.  Things look OK so far.

> There's still another problem in the patch - one I fixed yesterday. fclose() MUST be called after forming the tarball!

That seems correct.

> I'll reiterate: no matter what is done if the required input changes the user will have to re-enter their input.

That also seems correct.

In particular if there is a change needed to the input, the user is MUCH better rerunning the tool again.

It seems the use case for using **-a answers** and then **-i answers** is when the user is doing things such as updating their prog.c, or Makefile, or changing the remarks.md, or changing extra files.

xexyl:
> > There's still another problem in the patch - one I fixed yesterday. fclose() MUST be called after forming the tarball!
> 
> What is the problem that this fixes?

I use ``freopen()`` which makes stdin refer to the answers file. But when forming the tarball the called function also asks for input which means you can get a bad file descriptor. This breaks the tool.
> 
> > Maybe Ilya could make a tool that updates the answer file instead of trying to reinvent the answers file? A conversion tool or something? That seems more useful to me.
> 
> Warning confirmations cannot be predicted, so the format must be clear.

That's the point though. It's not predictable which is why they should be warned.

lcn2:
> At the same time there's no way you can use the same answers file for more than one entry ..

Correct, each entry will have its own answers file.  You shouldn't mix and march them for different entries it seems.

ilyakurdyukov:
> At the same time there's no way you can use the same answers file for more than one entry - unless you write to more than one answer file.

It's basically the same, the author information is the same (which is the most of the answers file), just edit the entry name and title, and you don't have to re-enter anything.

xexyl:
> We tested the patch we pulled it just a few motets ago. Things look OK so far.
> 
> > There's still another problem in the patch - one I fixed yesterday. fclose() MUST be called after forming the tarball!
> 
> That seems correct.
> 

Yes. I ran across another source of this problem which I also fixed before committing that time but this one I did not fix until last commit because I did not see it. The reason is simple: Landon added the finalising information before I added the option to read in the answers file which uses ``freopen()`` - and I did not think of it until then.

> > I'll reiterate: no matter what is done if the required input changes the user will have to re-enter their input.
> 
> That also seems correct.

Of course.

> 
> In particular if there is a change needed to the input, the user is MUCH better rerunning the tool again.

Exactly my point.

> 
> It seems the use case for using **-a answers** and then **-i answers** is when the user is doing things such as updating their prog.c, or Makefile, or changing the remarks.md, or changing extra files.

That's exactly what I made it for.

lcn2:
> It's basically the same, the author information is the same (which is the most of the answers file), just edit the entry name and title, and you don't have to re-enter anything.

Correct.  If someone wants to take that chance, there is no stopping them.

If we were doing it, we'd only reuse an answers file for the entry number it was created for.  But that is just us, maybe :-).

xexyl:
> > At the same time there's no way you can use the same answers file for more than one entry ..
> 
> Correct, each entry will have its own answers file. You shouldn't mix and march them for different entries it seems.

Of course that's the case. Because it also has title, abstract etc. I still think it would make it easier if the contestant ID/account (or whatever you want to call it) had the authors information tied to it so that you don't need the .authors.json file but until that's the case the answers file has to include it.

xexyl:
> > It's basically the same, the author information is the same (which is the most of the answers file), just edit the entry name and title, and you don't have to re-enter anything.
> 
> Correct. If someone wants to take that chance, there is no stopping them.
> 
> If we were doing it, we'd only reuse an answers file for the entry number it was created for. But that is just us, maybe :-).

Indeed it would be risky to use the file for more than one entry. I thought that was a given. Nobody can stop someone from trying to use the same file but it seems silly to do it anyway. Why not just make it as easy as possible by letting the tool write the answers file and then using the tool to read the answers?

ilyakurdyukov:
> I use `freopen()` which makes stdin refer to the answers file. But when forming the tarball the called function also asks for input which means you can get a bad file descriptor. This breaks the tool.

It's a call to yes_or_no()? Ok, I'll move it back.

lcn2:
So we recommend that the fclose() issue be address by modifying the patch.

We recommend that the intent of the answers file be for re-forming the same entry (say because the prog.c is being updated, remark's improved, new extra file added, etc.).

lcn2:
BTW: We will use '-i answers' for valgrind / ASAN testing.  It would have been nice when were were re-entering input date over and over again during the early days of testing.

lcn2:
BTW: We will use '-i answers' for valgrind / ASAN testing.  It would have been nice when were were re-entering input date over and over again during the early days of testing.

ilyakurdyukov:
> Why not just make it as easy as possible by letting the tool write the answers file and then using the tool to read the answers?

Because changing the source (prog.c) can make old answers inconsistent and inconsistent if changes to the source result in new warnings or fixing previous warnings.

Also there is "prog_c" in the messages, it should be "prog.c" which I noticed and haven't fixed yet.

xexyl:
> BTW: We will use '-i answers' for valgrind / ASAN testing. It would have been nice when were were re-entering input date over and over again during the early days of testing.

I'm glad I added it then! 

What about tying the author information to the contestant account? That would simplify the process of forming the tarball as there wouldn't be any need to enter it for every entry.

xexyl:
> > Why not just make it as easy as possible by letting the tool write the answers file and then using the tool to read the answers?
> 
> Because changing the source (prog.c) can make old answers inconsistent and inconsistent if changes to the source result in new warnings or fixing previous warnings.

Exactly. That's why I added the warning that you removed. There's no way past this. No matter what you do to the answers file this will _always_ be a problem. If you have a problem with that you shouldn't use the answers file. Well it is an option not a requirement.

> 
> Also there is "prog_c" in the messages, it should be "prog.c" which I noticed and haven't fixed yet.

Yes. I fixed some like that too though it was remarks_md I think (and it was that name in the tarball even).

lcn2:
> > Why not just make it as easy as possible by letting the tool write the answers file and then using the tool to read the answers?
> 
> Because changing the source (prog.c) can make old answers inconsistent and inconsistent if changes to the source result in new warnings or fixing previous warnings.

To test this, we replaced the small prog.c used by iocccsize-test.sh with mkiocccentry.c itself.  Then we ran:

```sh
./mkiocccentry -i test_work/answers.txt ./test_work test_src/{prog.c,Makefile,remarks.md} LICENSE README.md
```

And we saw mkiocccentry report:

```
You've ignored one or more warnings:

WARNING: prog.c size: 201955 > Rule 2a maximum: 4096

WARNING: prog.c size: 140507 > Rule 2b maximum: 2053

WARNING: prog.c has character(s) with high bit set
```

That seems OK.  Or are we missing something?

ilyakurdyukov:
> prog.c has character(s) with high bit set

Better fix it because some compilers don't like it.

An additional note should be added for the user to check the log for a more detailed explanation of the warnings. This is a short list of warnings. But this list is clearly visible, because it is at the same end, and not somewhere in the middle.

lcn2:
> WARNING: prog.c has character(s) with high bit set

The above was a result of using mkiocccentry.c as our prog.c.

However when we run iocccsize on it directly:

```sh
./iocccsize mkiocccentry.c
```

The iocccsize only reports:

```
140507
Warning: character(s) with high bit set found! Be careful you don't violate rule 13!
Warning: your source under Rule 2a: 201955 exceeds Rule 2a limit: 4096: Rule 2a violation!
```

We searched mkiocccentry.c and didn't find a character with the high bit set.  Did we miss something or is this also a bug?

ilyakurdyukov:
> We searched mkiocccentry.c and didn't find a character with the high bit set. Did we miss something or is this also a bug?

I don't know, but this clearly means that `infop->highbit_warning` is set to `true`.

xexyl:
> > > Why not just make it as easy as possible by letting the tool write the answers file and then using the tool to read the answers?
> > 
> > 
> > Because changing the source (prog.c) can make old answers inconsistent and inconsistent if changes to the source result in new warnings or fixing previous warnings.
> 
> To test this, we replaced the small prog.c used by iocccsize-test.sh with mkiocccentry.c itself. Then we ran:
> 
> ```shell
> ./mkiocccentry -i test_work/answers.txt ./test_work test_src/{prog.c,Makefile,remarks.md} LICENSE README.md
> ```
> 
> And we saw mkiocccentry report:
> 
> ```
> You've ignored one or more warnings:
> 
> WARNING: prog.c size: 201955 > Rule 2a maximum: 4096
> 
> WARNING: prog.c size: 140507 > Rule 2b maximum: 2053
> 
> WARNING: prog.c has character(s) with high bit set
> ```
> 
> That seems OK. Or are we missing something?

Well it doesn't warn the user about how IF they make a change that triggers a warning it could result in invalid input. I think it's wise to include a warning about this when automating it. 

This patch also bypasses the warning of the user of e.g. an empty prog.c until after the tarball is formed. In other words it doesn't let the user confirm that they want to fix the prog.c. It doesn't let them decide. That seems like a bad idea to me. It does show at the end that prog.c is empty but it doesn't give the choice to the user: it creates it no matter what.

For example let's say that you have a valid prog.c and then you update it. Now let's say your update triggers a warning. This patch doesn't let you decide whether or not you want to fix it; it just assumes yes and the tarball is created. Then it gives a warning.

lcn2:
> > We searched mkiocccentry.c and didn't find a character with the high bit set. Did we miss something or is this also a bug?
> 
> I don't know, but this clearly means that `infop->highbit_warning` is set to `true`.

FYI: This is correct:

```sh
./iocccsize mkiocccentry
```

as the binary surely has high bits :-)

```
97551
Warning: character(s) with high bit set found! Be careful you don't violate rule 13!
Warning: NUL character(s) found! Be careful you don't violate rule 13!
Warning: your source under Rule 2a: 105612 exceeds Rule 2a limit: 4096: Rule 2a violation!
```

Replacing test_src/prog.c with the mkiocccentry binary:

```sh
rm -rf ./test_work/test-0
cp mkiocccentry test_src/prog.c
./mkiocccentry -i test_work/answers.txt ./test_work test_src/{prog.c,Makefile,remarks.md} LICENSE README.md
```

yields:

```
ou've ignored one or more warnings:

WARNING: prog.c size: 105721 > Rule 2a maximum: 4096

WARNING: prog.c file size: 105721 != rule_count function size

WARNING: prog.c size: 97551 > Rule 2b maximum: 2053

WARNING: prog.c has character(s) with high bit set

WARNING: prog.c has NUL character(s)
```

Hummm ...

xexyl:
> > > We searched mkiocccentry.c and didn't find a character with the high bit set. Did we miss something or is this also a bug?
> > 
> > 
> > I don't know, but this clearly means that `infop->highbit_warning` is set to `true`.
> 
> FYI: This is correct:
> 
> ```shell
> ./iocccsize mkiocccentry
> ```
> 
> as the binary surely has high bits :-)
> 
> ```
> 97551
> Warning: character(s) with high bit set found! Be careful you don't violate rule 13!
> Warning: NUL character(s) found! Be careful you don't violate rule 13!
> Warning: your source under Rule 2a: 105612 exceeds Rule 2a limit: 4096: Rule 2a violation!
> ```
> 
> Replacing test_src/prog.c with the mkiocccentry binary:
> 
> ```shell
> rm -rf ./test_work/test-0
> cp mkiocccentry test_src/prog.c
> ./mkiocccentry -i test_work/answers.txt ./test_work test_src/{prog.c,Makefile,remarks.md} LICENSE README.md
> ```
> 
> yields:
> 
> ```
> ou've ignored one or more warnings:
> 
> WARNING: prog.c size: 105721 > Rule 2a maximum: 4096
> 
> WARNING: prog.c file size: 105721 != rule_count function size
> 
> WARNING: prog.c size: 97551 > Rule 2b maximum: 2053
> 
> WARNING: prog.c has character(s) with high bit set
> 
> WARNING: prog.c has NUL character(s)
> ```
> 
> Hummm ...

This might be because of what I noted in the my last comment. The patch doesn't let you answer whether you want to continue or not; it just assumes yes and only warns you at the end.

lcn2:
> Well it doesn't warn the user about how IF they make a change that triggers a warning it could result in invalid input. I think it's wise to include a warning about this when automating it.
> 
> This patch also bypasses the warning of the user of e.g. an empty prog.c until after the tarball is formed. In other words it doesn't let the user confirm that they want to fix the prog.c. It doesn't let them decide. That seems like a bad idea to me. It does show at the end that prog.c is empty but it doesn't give the choice to the user: it creates it no matter what.
> 
> For example let's say that you have a valid prog.c and then you update it. Now let's say your update triggers a warning. This patch doesn't let you decide whether or not you want to fix it; it just assumes yes and the tarball is created. Then it gives a warning.

Perhaps a waning boolean flag could be set to true of something is ignored and a cautionary warning at the end could be printed?  Or are we confused?

xexyl:
> > Well it doesn't warn the user about how IF they make a change that triggers a warning it could result in invalid input. I think it's wise to include a warning about this when automating it.
> > This patch also bypasses the warning of the user of e.g. an empty prog.c until after the tarball is formed. In other words it doesn't let the user confirm that they want to fix the prog.c. It doesn't let them decide. That seems like a bad idea to me. It does show at the end that prog.c is empty but it doesn't give the choice to the user: it creates it no matter what.
> > For example let's say that you have a valid prog.c and then you update it. Now let's say your update triggers a warning. This patch doesn't let you decide whether or not you want to fix it; it just assumes yes and the tarball is created. Then it gives a warning.
> 
> Perhaps a waning boolean flag could be set to true of something is ignored and a cautionary warning at the end could be printed? Or are we confused?

I actually had those. They were removed by this patch.

xexyl:
I think Ilya doesn't like the extra warnings. I gave a possible workaround: write my warnings to stderr and then he could in his scripts redirect stderr to /dev/null.

I can see the use of a version tag in the file so that IF the user is using an outdated version they are warned. But I think it's unwise to remove warnings.

I also think it's unwise to not let the user confirm whether they want to ignore the warning or not. It's quite easy to miss that whereas if it fails because of different input required the user can look at what happened and then fix it. This patch doesn't allow for that. It's so easy to ignore the final warning.

ilyakurdyukov:
> This patch also bypasses the warning of the user of e.g. an empty prog.c until after the tarball is formed. In other words it doesn't let the user confirm that they want to fix the prog.c. It doesn't let them decide. That seems like a bad idea to me. It does show at the end that prog.c is empty but it doesn't give the choice to the user: it creates it no matter what.

So which choice do you want, to force the user to re-enter everything to prove that this change is intentional?

What if I write the response file before I reach the size of Rule 2 and plan to use that answers after?

Also, in a situation where you have one warning of one type, and it will be replaced by a warning of another type - no one will stop you, because all warning confirmations are the same.

xexyl:
> > > We searched mkiocccentry.c and didn't find a character with the high bit set. Did we miss something or is this also a bug?
> > 
> > 
> > I don't know, but this clearly means that `infop->highbit_warning` is set to `true`.
> 
> FYI: This is correct:
> 
> ```shell
> ./iocccsize mkiocccentry
> ```
> 
> as the binary surely has high bits :-)
> 
> ```
> 97551
> Warning: character(s) with high bit set found! Be careful you don't violate rule 13!
> Warning: NUL character(s) found! Be careful you don't violate rule 13!
> Warning: your source under Rule 2a: 105612 exceeds Rule 2a limit: 4096: Rule 2a violation!
> ```
> 
> Replacing test_src/prog.c with the mkiocccentry binary:
> 
> ```shell
> rm -rf ./test_work/test-0
> cp mkiocccentry test_src/prog.c
> ./mkiocccentry -i test_work/answers.txt ./test_work test_src/{prog.c,Makefile,remarks.md} LICENSE README.md
> ```
> 
> yields:
> 
> ```
> ou've ignored one or more warnings:
> 
> WARNING: prog.c size: 105721 > Rule 2a maximum: 4096
> 
> WARNING: prog.c file size: 105721 != rule_count function size
> 
> WARNING: prog.c size: 97551 > Rule 2b maximum: 2053
> 
> WARNING: prog.c has character(s) with high bit set
> 
> WARNING: prog.c has NUL character(s)
> ```
> 
> Hummm ...

This might be because of what I noted in the my last comment. The patch doesn't let you answer whether you want to continue or not; it just assumes yes and only warns you at the end.

> > WARNING: prog.c has character(s) with high bit set
> 
> The above was a result of using mkiocccentry.c as our prog.c.
> 
> However when we run iocccsize on it directly:
> 
> ```shell
> ./iocccsize mkiocccentry.c
> ```
> 
> The iocccsize only reports:
> 
> ```
> 140507
> Warning: character(s) with high bit set found! Be careful you don't violate rule 13!
> Warning: your source under Rule 2a: 201955 exceeds Rule 2a limit: 4096: Rule 2a violation!
> ```
> 
> We searched mkiocccentry.c and didn't find a character with the high bit set. Did we miss something or is this also a bug?

I got the same results in the main repo so I guess it's a bug in iocccsize.c unless there really is a high bit set in mkioccccentry.

xexyl:
> > This patch also bypasses the warning of the user of e.g. an empty prog.c until after the tarball is formed. In other words it doesn't let the user confirm that they want to fix the prog.c. It doesn't let them decide. That seems like a bad idea to me. It does show at the end that prog.c is empty but it doesn't give the choice to the user: it creates it no matter what.
> 
> So which choice do you want, to force the user to re-enter everything to prove that this change is intentional?

Yes. Better safe than sorry. We humans are too used to automating things and when this happens we stop thinking. If something has changed then they should be alerted to this possibility and they can then confirm whether or not they intend to keep it that way. Better to let the user decide if they intentionally want to ignore the warning.

> 
> What if I write the response file before I reach the size of Rule 2 and plan to use that answers after?

Then you should be alerted to the problem!

> 
> Also, in a situation where you have one warning of one type, and it will be replaced by a warning of another type - no one will stop you, because all warning confirmations are the same.

That's a good catch but what can be done? As long as you only have to confirm to ignore a warning (instead of answering everything) this can happen. Which again is exactly why I had warnings about ignoring warnings.

ilyakurdyukov:
> I think Ilya doesn't like the extra warnings. I gave a possible workaround: write my warnings to stderr and then he could in his scripts redirect stderr to /dev/null.

Not really. I don't like to read extra hints over and over again when reading the log.

I didn't touch the warnings at all. Other than the replaced warning and the end saying that the answers will be incompatible if anything changes, because now the answers file are always compatible.

So, you want an error at the end if there is any warnings, and the final yes/no confirmation, even if using an answers file, will you be happy if I implement that? Or are you totally against my patch?

lcn2:
> So which choice do you want, to force the user to re-enter everything to prove that this change is intentional?
>
> What if I write the response file before I reach the size of Rule 2 and plan to use that answers after?
>
> Also, in a situation where you have one warning of one type, and it will be replaced by a warning of another type - no one will stop you, because all warning confirmations are the same.

Perhaps the safer thing to do is to abort on any problem when using **-i answers** and to issue a message that the user fix the issue or re-run without the answers file?

ilyakurdyukov:
I think that when the file is opened with the -i option, yes_no() can be called at the end, reading from stdin, to confirm if there are any warnings.

lcn2:
> I think that when the file is opened with the -i option, yes_no() can be called at the end, reading from stdin, to confirm if there are any warnings.

We can see your point as well.

How about, if there are warnings, issue them at the end AND exit non-zero perhaps?

lcn2:
> I think that when the file is opened with the -i option, yes_no() can be called at the end, reading from stdin, to confirm if there are any warnings.

That seems like a good thing to do.

xexyl:
> > I think Ilya doesn't like the extra warnings. I gave a possible workaround: write my warnings to stderr and then he could in his scripts redirect stderr to /dev/null.
> 
> Not really. I don't like to read extra hints over and over again when reading the log.
> 

That's exactly what I said.

> I didn't touch the warnings at all. Other than the replaced warning and the end saying that the answers will be incompatible if anything changes, because now the answers file are always compatible.

Yes you did. I had:

``
            ret = printf("\nYou've ignored one or more warnings. If you update your files so that these warnings\n"
                         "aren't triggered or if different warnings are triggered, it is very likely that the\n"
                         "answers file will cause invalid results and we STRONGLY recommend that you re-enter\n"
                         "your answers!\n");
``

And also

``
printf("\nAlthough you've not ignored any warnings be aware that if you add or change a file that\n"
                    "triggers a warning the next time, then this can result in improper answers being provided.\n");
``

And there might have been others.

> 
> So, you want an error at the end if there is any warnings, and the final yes/no confirmation, even if using an answers file, will you be happy if I implement that? Or are you totally against my patch?

Well I think you have some nice ideas like adding a header with the version. But I think it's unwise to remove the warnings and also to not have the confirmations of ignoring the warnings. That includes not letting the user decide. I don't think that's a good idea at all - as I said many will not notice this at all.

xexyl:
> > I think that when the file is opened with the -i option, yes_no() can be called at the end, reading from stdin, to confirm if there are any warnings.
> 
> That seems like a good thing to do.

This would however also mean that you have to interact with the program. Yes it would be mostly automatic but it wouldn't be entirely. Is there a way to restore stdin after using ``freopen()``? If there is I'm not sure. If there isn't then that would not be feasible since stdin becomes the answers file.

lcn2:
We have to go to a meeting on a different topic.  Good discussion.

Please do consider updating the patch and let us know if you are able to reach some sort of consensus.

lcn2:
> ret = printf("\nYou've ignored one or more warnings. If you update your files so that these warnings\n" "aren't triggered or if different warnings are triggered, it is very likely that the\n" "answers file will cause invalid results and we STRONGLY recommend that you re-enter\n" "your answers!\n");
>
> And also
>
> printf("\nAlthough you've not ignored any warnings be aware that if you add or change a file that\n" "triggers a warning the next time, then this can result in improper answers being provided.\n");

We liked those warnings, BTW.

xexyl:
> We have to go to a meeting on a different topic. Good discussion.
> 
> Please do consider updating the patch and let us know if you are able to reach some sort of consensus.

I personally think it's better to work on the things that still haven't been done. Ilya seems to be experienced in JSON so he could go for that part. He could also change the warnings to go to stderr if he doesn't like them so he could redirect it to /dev/null (as I suggested several times). That sounds like a good compromise to me.

What seems very important to me is to make sure to have those warnings though and also let the user decide whether or not they want to proceed. Ilya makes a point about how a different warning might be ignored because of the 'y' in the file but his patch doesn't let you decide at all - it just assumes yes for all warnings. But what can be done with the y ignoring a different warning? There IS a certain risk with ignoring warnings and there's no way past that. But it's better to give the user the chance to abort and tell warn them about this.

xexyl:
> > ret = printf("\nYou've ignored one or more warnings. If you update your files so that these warnings\n" "aren't triggered or if different warnings are triggered, it is very likely that the\n" "answers file will cause invalid results and we STRONGLY recommend that you re-enter\n" "your answers!\n");
> > And also
> > printf("\nAlthough you've not ignored any warnings be aware that if you add or change a file that\n" "triggers a warning the next time, then this can result in improper answers being provided.\n");
> 
> We liked those warnings, BTW.

I have a lot of experience with users. If you don't have warnings like this there will be people who complain! It's true that warnings might annoy some users but those users can simply ignore them; warnings like these are for the people who are actually doing the risky things.

ilyakurdyukov:
> This would however also mean that you have to interact with the program. Yes it would be mostly automatic but it wouldn't be entirely. Is there a way to restore stdin after using freopen()? If there is I'm not sure. If there isn't then that would not be feasible since stdin becomes the answers file.

I'll try to do this (so you have the feature you want and I have the feature I want), deal?

But it's in the morning, because it's already very late in my time zone.

xexyl:
> > This would however also mean that you have to interact with the program. Yes it would be mostly automatic but it wouldn't be entirely. Is there a way to restore stdin after using freopen()? If there is I'm not sure. If there isn't then that would not be feasible since stdin becomes the answers file.
> 
> I'll try to do this (so you have the feature you want and I have the feature I want), deal?

Well... as the judges liked my warnings I think they should be put back. I still also think it's imperative that the user gets to decide whether to ignore the warnings or not.

I also think using shell redirection should have the same behaviour. Maybe we should have been a bit clearer with each other. I'll start:

What _exactly_ do you want? I can then reply to that though it won't be until my tomorrow as I go to bed early. Maybe we can come to some consensus.


> 
> But it's in the morning, because it's already very late in my time zone.

Yeah I was thinking that. Sleep well!

xexyl:
A possible way to restore stdin occurred to me. I've not tested this but doing something else entirely the dup(2) syscall popped into my head. I have no idea why but this happens with me. It does seem like it would complicate it unnecessarily though (though see below). Also why have change stdin to be a file only for it to go back to input at the keyboard? 

Is the purpose to make the user confirm everything is okay after the tarball has been formed? I can maybe see that as an idea - since they are relying on automation. But at the same time they can look at the output and see if it's correct. I don't know. I'm kind of mixed here. As I'm thinking on it this might actually be a fair idea. I don't know if it's necessary though and it would complicate things quite a bit.

But Ilya it seems you're after certain things. I want to reiterate my question of what are those things you're after. I think that's the fair approach.

As for what I want it's basically what I did. I will not ever ignore the warnings but I think having the warnings there is a good idea (as do the judges as they have said). If you want a header/version tag I think that that might be an idea as long as it behaves the same as ``< answers``. Even if it doesn't though I suppose it's not a huge problem (I would use the option in the tool anyway). Having an EOF marker might also be an idea.  I'm not sure.

And you're right that if the y is there it might inadvertently say yes to a different warning but as I noted ignoring warnings has a certain risk to begin with and there's no perfect solution (but it's better to not assume yes for all the warnings).

What I am against is taking the choice of whether to ignore the warnings or not away from the user and I think that in addition to keeping that the user should be alerted to the fact that if they change the files (in a way that triggers - or no longer triggers - a warning) it might result in invalid input.

My original idea was that I could simply input the answers once and then not have to repeatedly input the same thing. I would never ignore the warnings about e.g. empty prog.c (though interestingly I did have a different take on an empty prog.c in 2018 and I thought it was fun and funny but I knew it was highly unlikely that it would win ... I knew also that if any of my entries that year would win it would be the Weasel entry) but I added this to the code though because I felt it should be complete - without it the answers file would not work (and you saw what happens if you change it from valid to not valid in your patch).

So what are your wishes? I again stress that if you're not liking the extra warnings just make them go to stderr so you can redirect it to /dev/null. That would solve that part of the problem for you but it would still be there for people who are actually going to ignore warnings. And as for me (besides above) the more I think on it the more I wonder if it's a good idea to have the user have to actually type 'y' that confirms that the tarball is correct. It might be a bit annoying for me but it might be the better way to go about it for the contest as a whole. The dup() syscall is the only way I can think of doing this but again I've not tested this (maybe there's another way too).

I think that's all I have for the day. Hope you're having a good sleep and a good day when you wake up!

xexyl:
> Perhaps the safer thing to do is to abort on any problem when using **-i answers** and to issue a message that the user fix the issue or re-run without the answers file?

Would you mind defining 'problem'? Because it might be something that's implicitly there. If the input is invalid then the process will fail. But define invalid input. If it's an invalid sequence it will fail but how to tell if it works but is actually not correct. Ilya pointed this out with 'y' being there could make it ignore a different warning though they already are ignoring warnings which is risky but is there another thing you can think of?

I think the idea of abort on any problems and issue a message that the user fix or re-run without the answers file is a sound idea but the question is (as above) how to define problem.

xexyl:
> > WARNING: prog.c has character(s) with high bit set
> 
> The above was a result of using mkiocccentry.c as our prog.c.
> 
> However when we run iocccsize on it directly:
> 
> ```shell
> ./iocccsize mkiocccentry.c
> ```
> 
> The iocccsize only reports:
> 
> ```
> 140507
> Warning: character(s) with high bit set found! Be careful you don't violate rule 13!
> Warning: your source under Rule 2a: 201955 exceeds Rule 2a limit: 4096: Rule 2a violation!
> ```
> 
> We searched mkiocccentry.c and didn't find a character with the high bit set. Did we miss something or is this also a bug?

FWIW I added in a copy of iocccsize.c in read_ch() an fprintf() call to print the character and this is what I got running it on mkiocccentry.c:

>high bit: 226
>high bit: 226
>high bit: 128
>high bit: 128
>high bit: 147
>high bit: 147
>high bit: 226
>high bit: 226
>high bit: 128
>high bit: 128
>high bit: 147
>high bit: 147
>high bit: 226
>high bit: 226
>high bit: 128
>high bit: 128
>high bit: 147
>high bit: 147

Running the 2020 iocccsize like:

``
./iocccsize -i < ../../mkiocccentry/mkiocccentry.c >/dev/null
``

Gave this:

> iocccsize: NUL or non-ASCII characters

Which I guess might be the diacritics in some of the countries? I did not do any debugging but maybe that will be something of interest to you or make you think of something in particular.

xexyl:
I just tried removing the diacritics and I got the same warnings.  I guess the obvious answer is to have the code print the adjacent characters so you can find out which ones are giving it trouble.

I'm going for the night but maybe it'll give you some ideas. If nothing else it doesn't seem to be the diacritics (I assume those characters have diacritics) in the country names.

ilyakurdyukov:
> Yes you did. I had:
> 
> `ret = printf("\nYou've ignored one or more warnings. If you update your files so that these warnings\n" "aren't triggered or if different warnings are triggered, it is very likely that the\n" "answers file will cause invalid results and we STRONGLY recommend that you re-enter\n" "your answers!\n");`
> 
> And also
> 
> `printf("\nAlthough you've not ignored any warnings be aware that if you add or change a file that\n" "triggers a warning the next time, then this can result in improper answers being provided.\n");`
> 
> And there might have been others.

So you're angry because I'm removing **your** warnings. But that warning is completely deprecated with the new method.

Because: "answers file will cause invalid results"

The answers will NEVER give wrong results with my method, either any warning are triggered or not triggered.

I explained it, but you keep insisting that these warnings be returned.

And why I strongly dislike the warnings you made, and are strongly against them:

These warnings do not solve the problem, instead of solving it, you leave the problem as it is and think that this is a good solution?

I think warning the user is not a solution. Is this something that can't be fixed? No, it can be! And I don't see your intent to make a real fix other than these warnings. You're happy not fixing the problem, but I'm not happy with that solution.

So I made a fix, and you are angry at me for changing your code - is this the real problem, what are you against my patch? Not because anything else? I removed that warning because it's obsolete now, because the problem solved.

> I personally think it's better to work on the things that still haven't been done. Ilya seems to be experienced in JSON so he could go for that part.

So, you want me to go do the other part and give up to fix this just because you don't want someone to edit your code, even if your code doesn't solve the problem and someone has a solution? Not fair at all.

ilyakurdyukov:
The answers file can now be read from stdin.

And the user is asked for final confirmation if there are any warnings and the -i option is used.

But it's clear to me that Cody refuses anyone to edit his code. That's the problem, because I offered a solution to all his complaints about this patch. This is very unconstructive behavior, because if we avoid changing other people's code, we won't be able to improve anything.

His warnings should no longer be here because the answer file is now consistent and cannot be broken by the different warnings layout.

ilyakurdyukov:
The new code should be edited with additional comments, improved warnings and error messages, and fixed error codes.

But I'm not in the mood to do it. Right now I'm very stressed and tired of this pointless conflict over the outdated warnings.

xexyl:
I will try looking at the rest later but I wanted to say I'm not angry with you. I know you're sick right now and you should take it easy until you're feeling better. I know all too well how hard that can be to do but it's important! I don't want to contribute to you not feeling well. If this discussion makes you think I'm angry it's time to take a step back - which like I said in email might be good anyway.

I hope you're feeling all better soon.

xexyl:
Will try now to reply because it seems to be important. I want to reiterate with you I'm not angry at all.

> > Yes you did. I had:
> > `ret = printf("\nYou've ignored one or more warnings. If you update your files so that these warnings\n" "aren't triggered or if different warnings are triggered, it is very likely that the\n" "answers file will cause invalid results and we STRONGLY recommend that you re-enter\n" "your answers!\n");`
> > And also
> > `printf("\nAlthough you've not ignored any warnings be aware that if you add or change a file that\n" "triggers a warning the next time, then this can result in improper answers being provided.\n");`
> > And there might have been others.
> 
> So you're angry because I'm removing **your** warnings. But that warning is completely deprecated with the new method.

Please don't make it all about me. This is one of the reasons I asked you what it is you want. Remember the judges also liked these warnings. I'm not saying they can't be improved but I think having warnings is important - yes even with your changes.

> 
> Because: "answers file will cause invalid results"
> 
> The answers will NEVER give wrong results with my method, either any warning are triggered or not triggered.

But that's the problem: the user should have the choice of deciding to confirm that they want to ignore the warnings _before_ the tarball is formed. Those final warnings you have are good but they're not enough for all users, I can assure you that!

> 
> I explained it, but you keep insisting that these warnings be returned.
> 
What's more important is that the user has the choice of confirming that they want to ignore the warnings - and in so doing it's why the warnings should be there. You have too much faith in users if you think the final warnings are enough!

> And why I strongly dislike the warnings you made, and are strongly against them:
> 
> These warnings do not solve the problem, instead of solving it, you leave the problem as it is and think that this is a good solution?

Neither is a perfect solution: your way doesn't solve the problem either because it removes the choice from the user whereas mine it can cause invalid input which means the user must revisit the answers (because a change was made that warrants this). 
> 
> I think warning the user is not a solution. Is this something that can't be fixed? No, it can be! And I don't see your intent to make a real fix other than these warnings. You're happy not fixing the problem, but I'm not happy with that solution.
> 
It's not a solution. There is no solution to the user ignoring warnings. But warnings are there for a reason and if the user cannot confirm this then there's a problem.

> So I made a fix, and you are angry at me for changing your code - is this the real problem, what are you against my patch? Not because anything else? I removed that warning because it's obsolete now, because the problem solved.
>

I'm not angry and if you think I am you should take a step back. I know you're recovering from a cold and it's not good to be working yourself up when you're unwell (it's never a good idea but especially when you're sick). I've even suggested that some of your ideas are great and I gave a possible solution of how to do it (e.g. since I used ``freopen()`` on stdin).

You say the warnings are obsolete now but that's because you've completely removed the possibility to confirm that the user wants to ignore the warning which is not a solution at all. Once the tarball is formed if the user did not get a prompt about it some are likely to ignore the final warnings. Again I think the final warnings are a good idea but I don't think not letting the user decide is a bad idea too.


 
> > I personally think it's better to work on the things that still haven't been done. Ilya seems to be experienced in JSON so he could go for that part.
> 
> So, you want me to go do the other part and give up to fix this just because you don't want someone to edit your code, even if your code doesn't solve the problem and someone has a solution? Not fair at all.

I'm not against all of your patch, no. What I was trying to say though is as the judges pointed out they could use more help on other parts - as the answers file is fine. Oh sure there are places it can be improved upon but removing the choice from the user is not one of those places. The judges have given us other things to do. As for bringing up JSON I did that because it seems your'e experienced with it - it was nothing but encouragement! Again if you're thinking like this you should take a step back (see above - I have agreed that some of your ideas are great even one that might be annoying to me but is a better idea than what I have nonetheless).

> 
> I think it's impossible to develop the code together when one of the participants is trying to keep their code untouched, even when it is outdated and there is a better solution. And they attack those who try to change their code, coming up with reasons why the new code is worse.

I'm not attacking you. I'm having a discussion with you. I also don't think your patch is entirely useless. We're disagreeing on one point in particular (and this point also leads to your removal of the warnings): that you think the user doesn't have to have the choice of confirming that they want to ignore the warning. That's the thing I'm against. The rest I'm not against. But as for the warnings I gave a possible solution to you (since it seems they annoy you).

Please take good care of yourself and if you're stressed do take a break. I cannot stress that enough! It's so important. As I also said I realised yesterday I've been irritable (though not at any specific person - just a general thing) and that's a reason I wanted to step back too. It's also why I asked you what you exactly wanted.

I hope this helps you out. I'm not against all of your changes. I was encouraging you on the JSON because you're experienced in it (and I'm not) and I know you'll do something great. I'm definitely not attacking you but if you think I am you should take a step back ... and I'll do the same. A clear head from both of us will help the matter.

Feel better soon. The most important thing right now is that you take good care of yourself whilst you're recovering from the cold. Not programming. Not anything else. You have to take care of yourself first. You can address my concerns later (again though I'm not attacking you nor am I against all your patch or ideas - I'm against one part of it).

xexyl:
> The answers file can now be read from stdin.

That's good.

> 
> And the user is asked for final confirmation if there are any warnings and the -i option is used.

Great. I agree with this.

> 
> But it's clear to me that Cody refuses anyone to edit his code. That's the problem, because I offered a solution to all his complaints about this patch. This is very unconstructive behavior, because if we avoid changing other people's code, we won't be able to improve anything.
> 

This is not true. I'm not against anyone editing my code. I have told you things I like about your ideas and patch. I'm only against one thing in your patch.

> His warnings should no longer be here because the answer file is now consistent and cannot be broken by the different warnings layout.

If you've removed the ability for the user to decide whether to ignore the warnings though this is not a better way. Yes it means that the warnings might not be relevant but it's riskier if the user is not **PROMPTED**. That's the problem. They should be prompted. Having a warning at the end is great (and I had one too) but if they're not prompted they're much less likely to think about it and that's not good. Well maybe it's good in that others are more likely to win :) but I'm trying to be fair to everyone.

xexyl:
> The new code should be edited with additional comments, improved warnings and error messages, and fixed error codes.
> 
> But I'm not in the mood to do it. Right now I'm very stressed and tired of this pointless conflict over the outdated warnings.

Please take good care of yourself. You're recovering from a cold. Doing that is more important. I'll give it a break too. Let me put it this way though:

My main complaint is NOT about removing the warnings; it's the reason you removed them: not prompting the user if they want to ignore the warning. Humans have this tendency to ignore warnings but they're MORE likely to ignore warnings if they're NOT prompted about it: if they're not asked if they want to continue anyway. In your code they don't even get a chance to confirm that (in the checking of the files). It's **that** that I am against.

This is not a pointless discussion (yes - discussion).

xexyl:
Okay here's a thought. I might just have a good compromise (or even just fully concede).

I just took a break - had a shower, had some food (well actually liquid - I don't eat until late in the day) - and I was thinking. I'm sorry I'm adding stress to your situation; that's not what I mean to do at all. I want you to feel well.

Last night I slept about 11 hours and I'm still exhausted. Part of the problem is I had a lot of vivid dreams and dreams always exhaust me. Anyway I will not consider any of this today. Instead I'll take it easy. Tomorrow I have something going on but I'll make an effort to look at this Tuesday.

It's true that I think it unwise to remove the prompting of the user about warnings (that's my problem with this patch) but since I won't be ignoring the warnings if it means that much to you I'm seriously considering just letting it go. I mean it's not the end of the world and it doesn't _truly_ affect me or my entries. 

I do have a suggestion though with your warnings that might just solve the problem: maybe instead of including say prog.c is 0 size maybe have the full warning message that the code shows earlier on (in the check functions). That way the user has a better idea of the seriousness of it. That could be a good compromise actually as although it would still not have the prompt it would remove your problems with the confirmation too (which means the answer file will always work - which seems good to me I might add ... again I'm not totally against your ideas!). This way you have your way but it still expresses more warnings to the user.

I'll probably look at the code before Tuesday but I will try not reacting to anything until then. How does that sound to you? I don't want to add to your stress and I'm sorry (it's very clear to me that I have) that I've upset you. I'm so serious about that that I'm almost tempted to just say: add the full warnings and then go for it. That might just be enough to make me happy (it's not as good, maybe, in some ways, as having the prompt, but it at least explains to the user _why_ there's a problem rather than just the one line message).

Hope you have a good night and hope you sleep well and again I'm very sorry that I've upset you. Btw I still think you should go for the JSON code irrespective of this code; I know you'll deliver a great tool! I know I wouldn't as I've never played with parsing JSON nor have I even played with JSON full stop. But it seems like you would do well. I'll take a look at the other requests later on - probably not today.

Cheers and I hope you feel better soon mate! I hope this message helps you out. I really am trying to be reasonable and that's why I'm trying to come up with a solution (and why I've given it thought even when not at the computer).

lcn2:
We will continue to watch your discussion @xexyl and @ilyakurdyukov.  Let us know when you reach a consensus and update the pull request if/as needed.

xexyl:
A short bit ago I had a thought that might help here. After that I looked at the code and saw that it's already there. But looking at the code I see a couple ways to improve it. I'd say go for accepting these and I'll add in my thoughts below.

Here's what I like about the patch that's different from my code:

- It does ask for confirmation if warnings are ignored (after the more recent changes from Ilya). The difference is it's asked right before the tarball is formed. The better thing here is that it doesn't need the 'y' in the answers file so if the user triggers a different warning it still works.
- It has a versions tag and an EOF marker. I think those are nice ideas.

I did notice though that if you ignore a warning you can have the result of:

> Are you sure you want to create a tarball with having these warnings unresolved? [yn]: FATAL[74]: prompt: EOF while reading prompt input

So reading from stdin is incomplete. Here's where I think there could be some improvements (besides making it so this is complete).

- Minor detail: show how many warnings are ignored rather than just 'one or more warnings'. This is easy enough to do.
- Next to the warnings like ``WARNING: prog.c is empty`` show the FULL warning (which is in the check functions).
- It might also be good to prompt for each warning individually before continuing because this way it's really drilled into the user that these can be a problem. I believe that a prompt for each warning GROUP is good; that is to say: all the warnings (if any) of the iocccsize tool should have a single prompt as should the prog.c being empty as well as the Makefile warning. This would be better than what I had and I think it improves Ilya's code too.
- I also think that the answers file should not be recorded until after confirmations are verified (clarified: Ilya's confirmations are verified). On the other hand this might not be necessary since the answers are not added to the answer file which means that the other information can be there by itself (which means that the answers file can in fact be used again).

Since Ilya is stressed and also these are my summarised thoughts I'll add these. Thus I would say you should go ahead and accept these changes and I'll work on the additions. This I think will make us both happy and I think the changes together will make the tool all the better.

How does that sound to you @ilyakurdyukov ? I hope this will make you happy and again I'm sorry if I caused you any stress. With a clearer head it's a lot easier to see your ideas as an improvement (I was against not prompting the user to ignore warnings and since you've added that - I did not see that yesterday as I was too tired - I think it's good but I still think I can update it a bit to make it a bit more rounded). 

I'm going to work on these changes now on Ilya's code and once this is accepted into the repo I will just copy the mkiocccentry.c in that directory to my fork and then I will do a PR.

xexyl:
Okay I just did the following against Ilya's changes and I think this will work out; it'll be a good compromise between us and I hope it'll work for him.

* Warnings

He had it so that there's a summary of warnings that are displayed after processing the files. However the warnings that were in the check functions were also printed. When reading from the answers file he had it so that it did not prompt you: it just went right ahead and proceeded but still showed the warnings.

What I did is make it so the full warnings are not displayed until the end (after processing) and prompt individually for each warning (as if it wasn't automated except that everything other than the warnings are automated much like how he had it only I show the full warnings to show the importance of it). 

If there aren't any warnings there's no problem: the creation of the tarball is done; else it's automated but at the end the user can decide whether or not to ignore the warnings. This means that those who don't have any warnings in their files can automate it without any prompting but if they do change it they will be prompted. This satisfies my disquiet about not having the prompts but it also means the answers file will always work (which satisfies Ilya's concern).

I also modularised the warnings: each warning is in its own function which prints the warning and prompts if the user wishes to proceed. This way there's no need to duplicate the code in the two places.

* The creation of the tarball

I made it so even if it's automated (answers file used) the user is prompted if the directory is correct (correct files and so on). I think this is a good idea but this is certainly a matter of opinion: it could go either way and I'd be happy; whatever the general consensus is. I'm okay with this either way.

* stdin versus answers file

I think that given that the answers file has the tag and EOF marker that maybe it's not necessary that if there are warnings that doing ``< answers`` needs to work; if they want that they can use the option to read in the answers file which makes it easier and built-in. I have a possible idea of how it might be done where it would work but I'm not sure if it's worth it.

* Summary

Once Ilya's PR is accepted I will commit and do another PR against his and I think we can both be satisfied. 

* Other unrelated information that might mean I can't be so easily working on the tool(s)

I seem to be having a problem with my charger on my laptop; my power bank works but this could be a serious drain so I'm not sure how much I can do until this can be worked out. It was working strangely before but after a power outage it no longer charges at all so I need to get a replacement (and that might not be the problem because it seems also that there's a possible building wiring fault on the line). I hope to get to the other stuff sooner than later.

lcn2:
This is great progress @xexyl and @ilyakurdyukov!  Thanks!

Let us know when the pull request has been updated / ready for us to merge.

xexyl:
@lcn2 Please merge it. Once that's done I'll fetch upstream on my fork and then commit my changes and make a pull request.

lcn2:
> @lcn2 Please merge it. Once that's done I'll fetch upstream on my fork and then commit my changes and make a pull request.

done, @xexyl .. thanks!

lcn2:
> @lcn2 Please merge it. Once that's done I'll fetch upstream on my fork and then commit my changes and make a pull request.

See #44 .  This was expected and the issue was created to track the issue.  We presume this will be addressed in your future pull request.

PR #43, opened 2022-01-31
Fix high bit mkiocccentry

xexyl:
The problem was in a comment: it had '–' instead of '-'.

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -4106,7 +4106,7 @@ check_extra_data_files(struct info *infop, char const *entry_dir, char const *cp
 	}
 
 	/*
-	 * basename must only use POSIX Fully portable characters: A–Z a–z 0–9 . _ - and the + character
+	 * basename must only use POSIX Fully portable characters: A-Z a-z 0–9 . _ - and the + character
 	 */
 	for (j=0; j < base_len; ++j) {
 	    if (!isascii(base[j]) ||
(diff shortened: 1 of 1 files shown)


xexyl:
I also fixed this in my changes to Ilya's patch (which I've not committed yet - won't until you accept that PR) so this shouldn't be added back by mistake.

lcn2:
Good catch @xexyl, thanks!

xexyl:
Welcome and glad I could help! I've always been good at catching these types of things though in this case I partly cheated: I made a temporary adjustment to iocccsize that does the following:

- If it didn't trigger the warning print the character.
- Else print the warning and exit.

So I saw that the problem was in the comment. I have noticed this very character before though so it was a simple matter of checking whether to not it was a ``-`` or something else (I use vim so just ``/-``).

What I'm curious to know is how did it even get there?

lcn2:
> Welcome and glad I could help! I've always been good at catching these types of things though in this case I partly cheated: I made a temporary adjustment to iocccsize that does the following:
> 
> * If it didn't trigger the warning print the character.
> * Else print the warning and exit.
> 
> So I saw that the problem was in the comment. I have noticed this very character before though so it was a simple matter of checking whether to not it was a `-` or something else (I use vim so just `/-`).
> 
> What I'm curious to know is how did it even get there?

It was probably a cut and paste error from a terminal window where it was rendered, we guess.

xexyl:
That makes sense. I have often had to fix mobile / so-called ‘smart quotes’ in files and indeed from copy pasting so I could see how this might have happened. 

I use vim (used to use vi or some variant) - which maybe I already mentioned -  so it’s a simple fix but one I have done many times over the years. 

Well at least it’s solved now. 

I really am going to sleep now. Will look at everything else tomorrow. Good night!

PR #45, opened 2022-02-02
Answers enhancements and fixes

xexyl:
The third commit fixes `mkiocccentry-test.sh` as well. Since I have provided logs for each of the commits I will not include them here but if it doesn't show up in git log (as happened once) I'll make a comment here.


lcn2:
Thank you @xexyl

PR #46, opened 2022-02-02
Minor fixes

xexyl:
Although it works the feature test macro should be enabled to expose the WEXITSTATUS() macro.

I also fixed some typos in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,13 +82,13 @@ SED= sed
 
 # how to compile
 #
-CFLAGS= -O3 -g3 -pedantic -Wall -Wextra
+CFLAGS= -D_XOPEN_SOURCE -O3 -g3 -pedantic -Wall -Wextra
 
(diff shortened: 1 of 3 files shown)


xexyl:
As for the commit that overhauls warning messages I'm aware of one problem that should be addressed but I'm too tired to do that today; I'll see about it tomorrow. The summary is thus:

Whereas the other warnings are complete even when using the answers file it appears that the warnings for the Makefile are not complete in this case. This is because the `inspect_Makefile()` function has the warnings itself and so I missed it. Thus only the summary is shown when using the answers file.

This should be easily fixed but I am afraid it will have to be done another day (hopefully tomorrow). Note that the warnings are shown but since they're ignored due to the answers file they're easy to miss.

lcn2:
Thank you @xexyl your help is much appreciated.

xexyl:
I’m very happy to help!

Issue #47, opened 2022-02-03
On -a answers always overwriting answers


xexyl:
I think this is a mistake for the following reasons:

- It no longer reminds the user how to use the answers file.
- And maybe more importantly it means that a simple typo will overwrite the answers. This has happened to me more than once and although it's only test mode it could be annoying when the contest is running.

I suppose this is up to discussion but I believe the answers file should not be overwritten for these reasons.

Is there a reason you made it so it always overwrites the answers file?

xexyl:
I had an idea to compromise if there's a good reason to have it always overwrite the answers file:

The tool could warn that it's about to overwrite the file and prompt them if they want to proceed or if they want to instead use the input option.

Another possibility (maybe this is a good idea anyway, I don't know - it just popped into my head though it's probably unnecessary) is a helper script that you can run instead of the tool directly. I'm not sure what this would look like and I also don't know if it's even necessary (it probably is not) but since it occurred to me I figured I'd bring it up.

lcn2:
You make good points, @xexyl.  The 'overwrite the file and prompt them' would work.   And possibly the `mkiocccentry-test.sh` might (or might not) need updating, depending on how this is done.

A pull request would be appreciated.

xexyl:
I'm not sure if I'll have time for this today but if not I should be able to tomorrow. Well, technically I have the time but I'm in the middle of something and although it's not important I'd like to finish it and after that I have to leave for the day. 

As for how it might be done I would think it would simply use the `yes_or_no()` function and if they confirm it'll overwrite it else it'll give them a message how to use it (which would be how it used to be). Another possibility is that there's a force option to overwrite the answers file so for example only overwrite the answers file IF the user also specifies a force option (what letter that would be I don't know but I suppose `-f`).

This is assuming there's a good reason to have it ever overwrite it in the first place. What was the thinking behind this? I'm sure there's a good reason for it but the reasons I gave are why I didn't have it overwrite it.

I also had a thought on another option that could be added which might be good for the test modes: an option to ignore all warnings (the warnings would still be noted in the .info.json file of course). I bring this up because this might be the right time to do it if modifying the script. Something tells me it wouldn't be desirable to have test mode automatically ignore (as in not display) warnings but maybe there's a use for the option in some situations?

xexyl:
[...]

PR #50, opened 2022-02-04
Resolve issue #47 + add ignore warnings option

xexyl:
If -a is specified and -A is NOT specified, warn the user if the answers
file already exists and ask them if they want to overwrite it. If they
say yes the answers file will be overwritten. If -A is specified or the
answers file does not already exist, the answers file is written to
whether or not it exists. -A implies -a (falls through in switch).

To make it easier to test add an option that allows the user to ignore
warnings. When this option -W is specified, ignore warnings once and
warn them that they're doing this, telling them how to fix this (in case
they did it by accident) and reminding them that the judges will not
ignore the warnings.

### DIFF
--- a/mkiocccentry-test.sh
+++ b/mkiocccentry-test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
-# A simple mkioccccentry test,
-# or an example of how to automate the process!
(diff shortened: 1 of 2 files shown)


lcn2:
Thanks @xexyl

PR #52, opened 2022-02-04
Modularise utility/helper functions

xexyl:
See the git log for more details as well as the rationale behind this.


lcn2:
Sorry @xexyl , we are also in the process of doing this right now, so we need to reject this pull request.

xexyl:
How funny and ironic is that! That's okay though. 

Btw: is it desirable to have a make depend? I have some Makefile black magic that can give do this easily. I actually did it but it also used the files in this pull request so it obviously won't be valid any more and maybe not desirable but figured I'd mention it.

lcn2:
A make depend would be useful.  We suspect we forgot some of the dependences in our last push.

Feel free to update util.h and util.c as needed, @xexyl 

Sorry for the inconvenience.

xexyl:
Nothing to be sorry about - but I appreciate it!

I'll update the Makefile then and you can take a look at it. Give me a little bit of time to put it together.

PR #54, opened 2022-02-04
Add make depend

xexyl:
This shows how it works. I have left out the rule_count.o because the
rule is special. Anyway when a new object file is added you can add it
to the OBJFILES variable and then run make depend. It uses awk but it
could be done with cut too (and probably should be fixed). I'm sure you
get the idea behind this and you can adapt it to your purposes if
necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ modules.order
 Module.symvers
 Mkfile.old
 dkms.conf
+
+# Makefile backup (make depend specific)
+Makefile.orig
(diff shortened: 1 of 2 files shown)


xexyl:
If this doesn't work exactly for your purposes I've no problem if you reject it and do it differently but this is what I had in mind. I use it in some other projects and it's very useful - as long as you don't forget to run `make depend` of course!

xexyl:
Oh one more note is that since this does modify the Makefile and since Makefile is a dependency itself it means that everything has to be recompiled. I don't know if you have a reason to have it there but I wanted to point it out since it tripped me up for a moment - I didn't expect it and tried locating the reason and then saw that Makefile itself is a dependency. I'm not sure if this somewhat defeats the purpose of the make depend or not - I'll leave that up to you to decide.

lcn2:
Thank you for this improvement, @xexyl

xexyl:
﻿
Happy to help!

PR #55, opened 2022-02-04
Move cmdprintf() to utils

xexyl:
This because the future tarball checker is supposed to call filenamechk.


xexyl:
Hopefully this does not cause any problems with the other commit being there but if it does I'm happy to fix it.

lcn2:
Thanks.

PR #56, opened 2022-02-05
Initial txzchk (issue #42)

xexyl:
This is incomplete but the following changes were made for this commit.
It's a lot and if there are any problems (as is possible with such
changes) I will fix them. I've tested various tarballs (size, not
starting with a directory, having any special files etc.) but much could
be done still. Because there's a lot I'm going to make a comment in the
GitHub pull request. I'm sorry if for all the text there but I think
it's important! I'll fix any problems and I'll tackle the things not
completed another time (probably not today) if the judges (or someone
else) does not (there are a number of things left to do).

* mkiocccentry

    Change usage message a bit: since the line was so long for the
    synopsis I just have [options] and below the options are shown (as
    they always were).

    The tool now runs (in the form_tarball() function) the txzchk tool
    (see below).

    The locating of the tools (tar, cp, ls) were moved to util.c/util.h
    in a separate function (that also finds other tools).

* mkiocccentry.1

    Formatting fixes.

* txzchk

For discussion here see the comment in the pull request. This will
include possible issues as well as things that aren't yet implemented.

* Makefile

    Add missing util.o to make clean rule.

    Add txzchk to Makefile.

Additionally there were some typo fixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ CFLAGS= -O3 -g3 -pedantic -Wall -Wextra
 # where and what to install
 #
 DESTDIR= /usr/local/bin
-TARGETS= mkiocccentry iocccsize dbg_test limit_ioccc.sh filenamechk
+TARGETS= mkiocccentry iocccsize dbg_test limit_ioccc.sh filenamechk txzchk
 TEST_TARGETS= dbg_test
 OBJFILES = dbg.o util.o
 SRCFILES = $(patsubst %.o,%.c,$(OBJFILES))
@@ -138,6 +138,9 @@ dbg_test: dbg.c dbg.h Makefile
 
 filenamechk: filenamechk.c limit_ioccc.h dbg.o util.o
 	${CC} ${CFLAGS} filenamechk.c dbg.o util.o -o $@
+
+txzchk: txzchk.c limit_ioccc.h rule_count.o dbg.o util.o Makefile
+	${CC} ${CFLAGS} txzchk.c rule_count.o dbg.o util.o -o $@
 
 limit_ioccc.sh: limit_ioccc.h Makefile
 	${RM} -f $@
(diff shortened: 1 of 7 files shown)


xexyl:
Before even reviewing the pull request please let me type out the comments I referred to. Give me a few minutes!

lcn2:
Thank you @xexyl

xexyl:
* txzchk

# What's implemented so far

It checks the file size (not the accumulated file size of each file in the tarball!) just like form_tarball() does in mkiocccentry.

It checks that the first file in the tarball is a directory with the permissions: drwxr-xr-x (that was my understanding but if this is incorrect it can be fixed).

It checks that any additional files start with a '-' (i.e.  normal file).

# What's not implemented yet

It does not verify that the proper files are in the tarball.

It does not tally the total file size (of all the files) to verify that the total is not over the limit.

It does not check for setuid/setgid bits or sticky bit.

Anything else that's needed but not listed in what's implemented.
        
# Potential issues

The error messages are almost certainly not what the judges want but I'm afraid I cannot read their minds (nor would I want to - and they definitely would not want to read mine! :) )

I wasn't sure if it should list all errors at the end or just abort at the first error. At first I had started a tally and printed the results at the end but then I used the err() function which meant that this code was not reached in the case of an error. Thus this code should be modified to either show all the problems at the end OR remove the printing of the number of issues (number of issues returned from the function). I'm leaving that in just in case it should report all the issues.

There might be some better ways to go about certain things.
        
It might be annoying to the user to see the output of tar twice?  Perhaps the form_tarball() function should merge the two: that is to say that whilst testing the tarball (which runs tar) it's there that the contents should be shown (if there's a problem this is reported instead)?

xexyl:
> Thank you @xexyl

Welcome. I did not expect it to go that quickly! It's as I said not complete but a good amount is there. Just check the above comment and you'll get a better understanding of what I mean. Much has to be done and I can certainly do some of it including fixing any issues. However that will have to be tomorrow I'm afraid - unless it's serious and I can find the energy.

PR #57, opened 2022-02-05
Remove redundant tarball listing output

xexyl:
Since the txzchk is always executed by mkiocccentry and since txzchk
prints out the contents of the tarball (since it has to parse the output
of tar) there's no need to explicitly invoke tar -tJvf. If the txzchk
tool is changed so it doesn't show the output of the tar listing then
this will have to be added back but this way it's no longer showing the
same information twice.

Typo fix(es).

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -5634,7 +5634,8 @@ write_author(struct info *infop, int author_count, struct author *authorp, char 
  * form_tarball - form the compressed tarball
  *
  * Given the completed entry directory, form a compressed tar file for the user to submit.
- * Remind the user where to submit their compressed tarball file.
+ * Remind the user where to submit their compressed tarball file. The function
+ * shows the listing of the tarball contents via the txzchk tool.
  *
  * given:
  *      work_dir        - working directory under which the entry directory is formed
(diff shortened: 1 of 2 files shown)


xexyl:
That's it for the day on my behalf I think. Tomorrow I'll look at writing man pages for the two new tools, update the readme and maybe look at additional tests to add to `txzchk`. Perhaps you'll have more ideas by the time I'm back at it.

lcn2:
Thank you @xexyl  And thanks for fixing the typo .. we forgot to spell check.

xexyl:
I just put my glasses back on quickly to say I had a good thought on how to check for special bits on the files. Instead of checking for specific characters (necessitating the need to play with chmod - even though I believe I know which characters they are) I can just check for any character that’s not the right ones. 

The function strcspn comes to mind (I think that’s the one). I could also use strchr of course to check for invalid characters also. 

I will work on this in the morning. I will also look at the other requirements and see what I come up with. I can definitely make the man pages and update the readme file. 

As for typos no worries. I have a language defined in my .vimrc file so it’s easy to spot these things. Plus my eyes tend to focus on mistakes in text. I don’t know why or how but they do. 

Ps: I hope this reads okay. Although I have my glasses on I have my distance glasses on - not my reading glasses. 

I would  also use formatting to make some of this stand out but I discovered that when emailing the reply it doesn’t seem to register on the website.

PR #59, opened 2022-02-07
txzchk updates and fixes (including several checks added)

xexyl:
Many changes have been made. Several checks have been added but not all are listed below. The original push had these changes:

*   File sizes test added to txzchk

txzchk now tallies the size of all the files in the tarball and rounds it up to the next multiple of 1024 and warns if it's > MAX_DIR_KSIZE (also if it's < 0 - which is an error).

The function that rounds it up (round_to_multiple()) uses off_t as it's meant for file sizes but perhaps it could be changed to be more general (maybe long long).

*   Bump version in fnamchk and mkiocccentry

The change in filename format seems like a reason to update the version.

*   t_flag_used -> tar_flag_used (txzchk.c)

Previously only fnamchk bool had this format.

*   Make txz filename format consistent w/dir name

In anticipation of the txzchk verifying that the directory name is correct, change the format of the filenames to be:

        "entry." ("test" or a valid UUID) "\-" entry_number "." timestamp ".txz"

This is because the directory name in the tarball is "test" or valid UUID followed by a '-' and then the entry number instead of a '.'. The tool fnamchk used to print out either "test" or a valid UUID followed by a '.' followed by the entry number; now it uses a '-' instead of a '.' which is exactly how the directory name is. Thus the txzchk tool can get the expected directory name in the tarball by calling the fnamchk tool.
  
What's not done yet is moving fnamchk call to before the test for the directory name tests. If however I extract all the fields in advance I can perform all the tests that need to be done (except the directory name) before calling fnamchk and in this way I can still have fnamchk return an error if the filename is not correct. On the other hand fnamchk could be changed (this was not done) to warn on invalid filenames rather than error out (thus preventing txzchk from continuing which is why fnamchk call was moved to after all the tests txzchk performs).

Both fnamchk.1 and mkiocccentry.c have been updated to account for this change.

Instead of doing == true and == false just have e.g. if (test) and if (!test).

To be consistent with txzchk_flag_used make the other flag_used variables in this form too. This can be of help when looking later at the code as it can remind exactly what the flag means.

### DIFF
--- a/fnamchk.1
+++ b/fnamchk.1
@@ -6,7 +6,7 @@ fnamchk \- IOCCC compressed tarball filename sanity check tool
 .SH DESCRIPTION
 \fBfnamchk\fP verifies that an IOCCC compressed tarball is properly named.
 .PP
-The program validates that the filename is correct, in either the form of \fI"entry." ("test" or a valid UUID) "." entry_number "." timestamp ".txz"
+The program validates that the filename is correct, in either the form of \fI"entry." ("test" or a valid UUID) "\-" entry_number "." timestamp ".txz"
 \fP where \fBentry_number\fP is an integer between 0 and MAX_ENTRY_NUM (see \fIlimit_ioccc.h\fP) and where \fBtimestamp\fP is an integer number of seconds since the epoch.
 .PP
 More specifically, the filename must:
@@ -16,7 +16,7 @@ Start with "entry."
 Followed by either test or a UUID string of the form xxxxxxxx-xxxx-4xxx-axxx-xxxxxxxxxxxx
 Where "x" are [0-9-f] characters. And yes, there is a "4" (UUID version 4) and an "a" (UUID variant 1) in there.
 .PP
-Followed by "."
+Followed by "\-"
 .PP
 Followed by an decimal entry number between 0 and MAX_ENTRY_NUM (see \fIlimit_ioccc.h\fP) inclusive.
 .PP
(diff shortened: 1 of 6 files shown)


lcn2:
Wow @xexyl looks like you put in a lot of good work.  We will process the pull request and look at things later as we are deep in changes to https://github.com/ioccc-src/winner (1223 files changed so far!).

The `txzchk` is not a simple tool to write.  It is good that you are taking the time to write well, even if it is for the [IOCCC](https://www.ioccc.org).  We have been trying to make the code in this repo to serve as a GOOD example of what we call `literate programming`.

xexyl:
> Wow @xexyl looks like you put in a lot of good work. We will process the pull request and look at things later as we are deep in changes to https://github.com/ioccc-src/winner (1223 files changed so far!).

Keep up the great work!

> 
> The `txzchk` is not a simple tool to write. It is good that you are taking the time to write well, even if it is for the [IOCCC](https://www.ioccc.org). We have been trying to make the code in this repo to serve as a GOOD example of what we call `literate programming`.

Thank you! I take that as a great compliment - even more from the IOCCC judges than anyone else! 

Do you mind elaborating on what you mean by `literate programming` though? Anyway thank you for the wonderful compliment! 

I do have more work to do with it but I'll continue that probably tomorrow. I'm kind of tempted to now but I think it's better I take a break as I did a lot today and better to take it easy (esp as I have been awake so many hours and I don't sleep well).

Thank you again!

PR #60, opened 2022-02-08
txzchk: Clean up, bug fixes and more tests added

xexyl:
* Clean up and bug fix txzchk
    
Make warning messages more consistent in format: always show the tool's name and the txzpath. For example if there's no .info.json file it'll now show e.g.: 
    
            Warning: txzchk: test.txz: no .info.json found
    
...whereas before it showed the function name and did not show the tarball path (in this case the fnamchk would issue an error). I feel that showing the utility name is more useful than 'check_tarball()'.  It's possible I have missed a warning. I tried to do the same with errors but I spent less time working on those as those are fatal and the program ends.
    
Removed the check_files() function; the tests are now all in check_tarball().
    
check_tarball() now issues a warning if fnamchk fails so that we can continue if it does fail (though currently it's the last step of the function).
    
txzchk now always shows how many warnings were issued even if fnamchk fails (which is considered an issue).
    
txzchk now exits with 1 if there are issues found.

* txzchk: Check for NUL bytes before end of line

### DIFF
--- a/txzchk.1
+++ b/txzchk.1
@@ -35,7 +35,7 @@ Specify path to the IOCCC \fBfnamchk(1)\fP tool.
 If this option is not specified the program looks under the current working directory and \fI/usr/local/bin\fP.
 .SH EXIT STATUS
 .PP
-\fBmain()\fP returns 0 on success; if there's an error the end is not reached.
+\fBmain()\fP returns 0 if no issues are found and 1 if there are issues found; if there's an error the end is not reached.
 .SH FILES
 \fItxzchk.c\fP
 .RS
--- a/txzchk.c
+++ b/txzchk.c
@@ -759,7 +759,7 @@ check_tarball(char const *tar, char const *fnamchk)
 	}
 	/*
 	 * attempt to find !isdigit() chars (i.e. the tarball listing includes
-	 * the owner name of the files
+	 * the owner name of the file).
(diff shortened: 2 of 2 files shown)


xexyl:
Btw the reason I changed `issues` to be `total_issues` is I am thinking that it might be nice to print out the tarball contents and after that parse each line and report any issues so that the warnings are not interspersed with the tar listing.

Originally I had the thought that I would have a list of issues so I needed to rename the `issues` variable. But then I thought of another way to do it. Whichever way I do it I don't know but it's for a later commit. It's possible I do more today but I'm not sure of this. If not today then should definitely do tomorrow. 

These commits fixed a bug and made (I hope) warning messages easier to parse at the command line as they now have a general prefix like:

> Warning: txzchk: test.txz:

Where `test.txz` is the name of the tarball. It's not the only prefix but that's the general one. For example there's also:

> Warning: txzchk: ./fnamchk test.txz failed with exit code: 3

But I think that might be one of the only ones that is a different format (perhaps it should be changed?) if not the only one.

xexyl:
More tests added just now.

xexyl:
Just made the final push for the day (most likely). Hopefully tomorrow I can add detection of subdirectories. That might be the final test to add but I'm not certain of this yet.

PR #62, opened 2022-02-09
Add support for linux tar output format

xexyl:
There are three new functions: parse_line() which is the driver of the
other two; it takes the line read in from the tar command (or text file)
(readline() in check_tarball() which calls parse_line()) and checks for
special bits first, warning if any are found. Then it determines if the
output is linux (6 fields) or BSD/macOS (9 fields) and calls the
appropriate function depending on this. The check comes down to the
second field having a '/' in it.

What I will probably do next is move the checks that are done when
parsing the line (not all tests can be done in the same location) into
their own function and call that function from both parse_linux_line()
and parse_bsd_line(). Either way make test should now succeed on both
linux and macOS (and hopefully BSD). After this clean up subdirectories
still have to be detected.


lcn2:
Thank you @xexyl 

We will be happy to test some more this tomorrow / later this week.

xexyl:
Thank you as well.

That sounds good. Just to be sure:

`make test` did work on my end under linux as well as the output of linux under macOS. It works okay for you too, right?

PR #65, opened 2022-02-10
Major clean up of txzchk - output + modularity

xexyl:
Show all warnings _after_ parsing files. This is done by saving every
line to another linked list so that after the entire tarball contents
are displayed (via the readline() function) the checks are run on the
list of lines rather than each line as the line is read. There are still
some warnings that are shown before and after rounding the total file(s)
size to a 1024 multiple but I tried to move as many warnings together as
possible (some warnings have to be detected whilst parsing each line;
others can be done at the end and some fit better when parsing each
line).

Move duplicate code into separate functions.

I tried also to make warning messages a more consistent format but more
can be done in this department.

These changes have been tested under linux (CentOS) and macOS with the
tarballs provided by the judges on GitHub as well as some text files
(via the -T option) I created and with the exception of subdirectories
(which I hope to work on later today) everything _should_ be detected.

### DIFF
--- a/txzchk.c
+++ b/txzchk.c
@@ -54,6 +54,7 @@ struct info {
     bool has_prog_c;
     bool has_remarks_md;
     bool has_Makefile;
+    bool has_correct_directory;
     unsigned dot_files;
 } info;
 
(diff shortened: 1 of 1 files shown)


lcn2:
Good stuff, @xexyl!

We will test and see how it goes.

PR #66, opened 2022-02-11
txzchk: bug fixes and man page updates

xexyl:
I've updated the code to use `strtok_r`() instead of `strtok()`.

The self-deprecating humour in the man page fits me well as I enjoy laughing at myself.

I also removed some ambiguity in a variable name.

Changed the detection of `../` in paths a bit (now checks for just `..`) and fixed the dot file warning to say only 'file' if only one dot file was found (i.e. it used to say files but now it only shows that if > 1 found).

### DIFF
--- a/iocccsize.1
+++ b/iocccsize.1
@@ -48,6 +48,10 @@ Source file to the \fBiocccsize\fP tool.
 .RS
 Script that tests the \fBiocccsize\fP tool.
 .RE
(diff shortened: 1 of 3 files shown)


lcn2:
Good work, @xexyl !

xexyl:
Thank you! I will get back to the other things when you reply - today or else tomorrow.

PR #67, opened 2022-02-11
NULL pointer checks in txzchk.c

xexyl:
The pointer saveptr should never be NULL but I've added it to the NULL
pointers check in both parse_bsd_line() and parse_linux_line() to be
complete (I forgot to do this when changing strtok() to strtok_r()).


lcn2:
Good catch, @xexyl !

xexyl:
Thanks. I'm not even sure why I thought of it - I just did.

PR #70, opened 2022-02-12
Initial JSON tool checkers + add my URLs in txzchk.c

xexyl:
As for adding my URLs to `txzchk.c`: Thank you again Landon for adding my name to txzchk.c! I'm honoured and privileged and I'm happy I could help!

I think the other commits will be of help for a good start on the json tools. 

`make depend` was fixed as well (it reported an error after one of the recent changes in the `Makefile`) and I also executed it (so the dependencies have been updated at the end of the file).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ DESTDIR= /usr/local/bin
 TARGETS= mkiocccentry iocccsize dbg_test limit_ioccc.sh fnamchk txzchk jauthchk jinfochk
 MANPAGES = mkiocccentry.1 txzchk.1 fnamchk.1 iocccsize.1
 TEST_TARGETS= dbg_test
-OBJFILES = dbg.o util.o mkiocccentry.o iocccsize.o fnamchk.o txzchk.o jauthchk.o
+OBJFILES = dbg.o util.o mkiocccentry.o iocccsize.o fnamchk.o txzchk.o jauthchk.o jinfochk.o
 SRCFILES = $(patsubst %.o,%.c,$(OBJFILES))
 
 all: ${TARGETS} ${TEST_TARGETS}
@@ -303,3 +303,4 @@ iocccsize.o: iocccsize.c
 fnamchk.o: fnamchk.c limit_ioccc.h dbg.h util.h
 txzchk.o: txzchk.c limit_ioccc.h dbg.h util.h
 jauthchk.o: jauthchk.c limit_ioccc.h dbg.h util.h json.h
+jinfochk.o: jinfochk.c limit_ioccc.h dbg.h util.h json.h
--- a/Makefile
+++ b/Makefile
@@ -112,10 +112,10 @@ CFLAGS= -D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE -std=c99 -O3 -g3
 #
 MANDIR = /usr/local/share/man/man1
 DESTDIR= /usr/local/bin
-TARGETS= mkiocccentry iocccsize dbg_test limit_ioccc.sh fnamchk txzchk
+TARGETS= mkiocccentry iocccsize dbg_test limit_ioccc.sh fnamchk txzchk jauthchk jinfochk
 MANPAGES = mkiocccentry.1 txzchk.1 fnamchk.1 iocccsize.1
 TEST_TARGETS= dbg_test
-OBJFILES = dbg.o util.o mkiocccentry.o iocccsize.o dbg_test.o fnamchk.o txzchk.o
+OBJFILES = dbg.o util.o mkiocccentry.o iocccsize.o fnamchk.o txzchk.o jauthchk.o
 SRCFILES = $(patsubst %.o,%.c,$(OBJFILES))
 
 all: ${TARGETS} ${TEST_TARGETS}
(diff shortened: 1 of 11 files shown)


xexyl:
I'm going to take a break and then I'll see about reading the requirements for each tool. In the meantime now at least:

- Help message is implemented.
- Parsing of options is done.
- Verifying that a file is passed in and that it exists, is a regular file and is readable.
- It attempts to open it (reporting an error if `NULL` is returned from `fopen()`) and then at the end of the respective function calls `fclose()` (reporting a warning if it fails).

The return value of each tool depends on a new variable in the respective structs: `issues`: at the end of `main()` there's `exit(info.issues != 0);` (for `jinfochk`) and `exit(author.issues != 0);` (for `jauthchk`. This means the `info` and `author` variables are not unused (first commit they were) but it also seems like a good idea to have them there. Perhaps even `mkiocccentry` could write the number of issues to the respective file though how that might be done if the file is written first I don't know (but I've not read the requirements in full yet either).

Later today I hope to begin parsing of the files (I think the parsing for both will at least in some locations be pretty similar and I think `strtok_r()` will be of use) but this is a good start.

xexyl:
As for the most recent commit (`off_t` one) I wrote in the changelog as a possible issue but one which I did not know if there was a reason behind it (so did not touch it). Bringing it up here in case it is an issue:

```
    Note that in limit_ioccc.h the #define RULE_2A_SIZE had a cast to
    ssize_t and I changed it to off_t; however in iocccsize.c itself in
    stand-alone mode neither of the RULE_* defines have a cast; they're
    just:
    
         #define RULE_2A_SIZE           4096    /* IOCCC Rule 2a */
         #define RULE_2B_SIZE           2503    /* IOCCC Rule 2b */
    
    I am not sure if this is a problem but it's inconsistent with what it is
    in limit_ioccc.h. I will mention this in a comment on GitHub to see if
    it should also be fixed.
```

The reason I changed the cast to `off_t` btw: in `mkiocccentry.c` the same variable in the info struct is assigned the value of `file_size()` which is of type `off_t`.

xexyl:
Okay as you will see now the only thing that has to be done now is writing the parser/checker of the files: everything else should be in order. This means that `mkiocccentry`:

- Accepts `-j` and `-J` for `jinfochk` and `jauthchk` (if not it looks for `./jinfochk` and `./jauthchk` first before then looking under `/usr/local/bin`).
- Makes sure that both tools exist, are a file and are executable.
- After writing the `.info.json` and closing the file the tool runs `jinfochk` on it (with `-q` option that I added: to prevent seeing welcome messages etc.); if any errors are reported it aborts (right now since the tools only parse options and make sure that the file exists it will not report any errors - unless one of the tools somehow can't read the file anyway).
- If writing `.info.json` succeeds and the test passes the same procedure is done for `.author.json` with the `jauthchk` tool.

If the above all succeeds `mkioccentry` will only then create the tarball. Once that's done it of course runs `txzchk` which will simultaneously run `fnamchk` and also show the user the output of their tarball.

As you'll also see I added debug output to `txzchk` that you might find useful when analysing the entries. Of course I arbitrarily selected the debug levels but if you want to see the messages look for the function `show_txz_info()`.

I'm not sure if I'll start writing the parsers today. I might make it an early day since I'm waking up so early regardless of when I go to sleep and tomorrow will be a long (but fun) day. Before I do do the parsing though I have to read the requirements in the two issues you opened. When I get to that point I'll write in those threads.

lcn2:
> As for the most recent commit (`off_t` one) I wrote in the changelog as a possible issue but one which I did not know if there was a reason behind it (so did not touch it). Bringing it up here in case it is an issue:
> 
> ```
>     Note that in limit_ioccc.h the #define RULE_2A_SIZE had a cast to
>     ssize_t and I changed it to off_t; however in iocccsize.c itself in
>     stand-alone mode neither of the RULE_* defines have a cast; they're
>     just:
>     
>          #define RULE_2A_SIZE           4096    /* IOCCC Rule 2a */
>          #define RULE_2B_SIZE           2503    /* IOCCC Rule 2b */
>     
>     I am not sure if this is a problem but it's inconsistent with what it is
>     in limit_ioccc.h. I will mention this in a comment on GitHub to see if
>     it should also be fixed.
> ```
> 
> The reason I changed the cast to `off_t` btw: in `mkiocccentry.c` the same variable in the info struct is assigned the value of `file_size()` which is of type `off_t`.

The values found inside iocccsize.c are for when neither MKIOCCCENTRY_USE nor IOCCCSIZE_STANDALONE is defined.
The iocccsize.c goes back to Anthony Howe issues related to his original code.  As such those values are not used.  See the big comment above them.

xexyl:
> > The reason I changed the cast to `off_t` btw: in `mkiocccentry.c` the same variable in the info struct is assigned the value of `file_size()` which is of type `off_t`.
> 
> The values found inside iocccsize.c are for when neither MKIOCCCENTRY_USE nor IOCCCSIZE_STANDALONE is defined. The iocccsize.c goes back to Anthony Howe issues related to his original code. As such those values are not used. See the big comment above them.

Thanks. I noticed the comment after I wrote the above.

xexyl:
> Thanks @xexyl !

Happy to help! Waiting on the replies to the two issues and then I'll hopefully get to working on those tools - but I am almost certainly done for the day. I'll probably write more tomorrow.

Have a great day!

PR #71, opened 2022-02-13
Various updates and fixes

xexyl:
txzchk now checks for invalid chars in filenames (full path): that is it
now only allows fully POSIX portable chars (see mkiocccentry.c) as well
as '/' (since we check on the full path and not just the basename). I.e.
it checks on the field 'char *filename' in struct file. Note that as
soon as an invalid char is detected it breaks out of the loop (so a file
with two invalid chars only triggers one warning and issue).

txzchk also now checks for files called just '.'. These files are
counted as BOTH an invalidly named dot file AND a file called just '.'
(which is clearly abuse).

The show_txz_info() function has been updated for both of these changes.
As well instead of showing if there were any unacceptable dot files it
shows how many.

### DIFF
--- a/fnamchk.1
+++ b/fnamchk.1
@@ -1,8 +1,8 @@
-.TH fnamchk 1 "06 February 2022" "fnamchk" "IOCCC tools"
+.TH fnamchk 1 "13 February 2022" "fnamchk" "IOCCC tools"
 .SH NAME
 fnamchk \- IOCCC compressed tarball filename sanity check tool
 .SH SYNOPSIS
-\fBfnamchk fnamchk [\-h] [\-v level] [\-V] filepath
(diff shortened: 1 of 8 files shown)


xexyl:
I just realised that you merged this - and I have been making more commits. I have to fix this by making a new branch and then doing another pull request.

I also see in the git log that it only has the summary and not the full text. I guess that's because I make a branch first. I'll add the previous commit logs to the new commit to get everything in order. I just have to copy the files first to another location and then get the most recent version of the main repo and then will re-commit. More to follow soon I hope!

xexyl:
Okay I apologise for the problem with not making a branch: it appears there's a problem so what I'm going to do is revert the change and then re-commit into a new branch with all the changes. That way everything should be able to go through to the main repo. Sorry about that!

PR #74, opened 2022-02-14
jauthchk, jinfochk, txzchk return value update

xexyl:
The tools now exit with the number of issues found rather than 0 for no issues and 1 for one or more issues found.

Updated the man pages for this change.

Typo fix in json.c.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ txzchk
 test-set
 jauthchk
 jinfochk
+entry.*.tx[zt]
 
 # Prerequisites
 *.d
--- a/jauthchk.1
+++ b/jauthchk.1
@@ -1,4 +1,4 @@
-.TH jauthchk 1 "13 February 2022" "jauthchk" "IOCCC tools"
+.TH jauthchk 1 "14 February 2022" "jauthchk" "IOCCC tools"
 .SH NAME
 jauthchk \- IOCCC tool to validate .author.json file found within an entry directory
 .SH SYNOPSIS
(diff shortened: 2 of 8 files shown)


xexyl:
I think you have some interesting thoughts here. First the reason I did it this way (and below I can see how it's a problem):

```shell
exit codes:

	0		no errors or warnings detected
	>0		some error(s) and/or warning(s) were detected

```

So the idea in my head (at the time!) was that the number of issues is the exit code. But I can see how the exit code might be better if it's more static: 1 for error and 0 for not. This way it can also be more easily tested in a script (I did not think of that until just now - not even reading your reply made me think of that).

But as for the reporting the last problem: how do you suggest the exit codes are defined? I mean since the problem can be any number of issues (this goes for the tarball checker as well as the json checkers). Do you have a suggestion? I can see how that might be good: whatever caused it to abort. At the same time though for `txzchk` it might not always abort (depends on if it's run directly or indirectly). Since the tool also tells you how many issues were detected perhaps this (the number of issues being the exit code) is not necessary.

You're right that exit codes should be defined. I was thinking of the `> 0` part as a counter and did not go beyond that. Do you want to come up with a list of exit codes that can be used or how do you propose to proceed?

Maybe the exit code indicates the TYPE of problem rather than what problem in particular? For example the issues in the tarball might be given a type (fatal, warning, trivial - or whatever) and depending on what caused the error it would use that exit code.

In the meantime I'm making a minor change to the files now so I remember to at least change it back to how it was: return 1 if there are issues. If you come up with a system I'll be happy to implement it. Since you have more ideas what to expect perhaps you would like to come up with the exit codes that can then be added?

As for the warnings in `json.c` no worries. I'll be going for the day but I'm happy to implement a better system for exit codes (for now though I'm going to make it so total_issues != 0 but I won't commit it yet until we come up with something).

PR #75, opened 2022-02-14
Change exit codes

xexyl:
We can talk about it here. See commit log:

>    Until exit codes can be defined, revert back to just returning 1 if
>    there's an error and 0 if not.

I'm not sure what exit codes there should be - I did not think of it at the time but you're right that they should be more static. I won't go for anything else until either tomorrow or Wednesday though. 

Do you want to define exit codes? Maybe `mkiocccentry` could have some too? Let me know and I'd be happy to implement them. But I'm done for the day.

Enjoy the rest of your day!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ txzchk
 test-set
 jauthchk
 jinfochk
+entry.*.tx[zt]
 
 # Prerequisites
 *.d
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ DESTDIR= /usr/local/bin
 TARGETS= mkiocccentry iocccsize dbg_test limit_ioccc.sh fnamchk txzchk jauthchk jinfochk
 MANPAGES = mkiocccentry.1 txzchk.1 fnamchk.1 iocccsize.1
 TEST_TARGETS= dbg_test
-OBJFILES = dbg.o util.o mkiocccentry.o iocccsize.o fnamchk.o txzchk.o jauthchk.o jinfochk.o
+OBJFILES = dbg.o util.o mkiocccentry.o iocccsize.o fnamchk.o txzchk.o jauthchk.o jinfochk.o json.o
 SRCFILES = $(patsubst %.o,%.c,$(OBJFILES))
 DSYMDIRS = $(patsubst %,%.dSYM,$(TARGETS))
 
@@ -136,8 +136,8 @@ rule_count.c: iocccsize.c
 rule_count.o: rule_count.c limit_ioccc.h Makefile
 	${CC} ${CFLAGS} -DMKIOCCCENTRY_USE rule_count.c -c
 
-mkiocccentry: mkiocccentry.c limit_ioccc.h json.h rule_count.o dbg.o util.o Makefile
-	${CC} ${CFLAGS} mkiocccentry.c rule_count.o dbg.o util.o -o $@
+mkiocccentry: mkiocccentry.c limit_ioccc.h json.h rule_count.o dbg.o util.o json.o Makefile
+	${CC} ${CFLAGS} mkiocccentry.c rule_count.o dbg.o util.o json.o -o $@
 
 iocccsize: iocccsize.c limit_ioccc.h Makefile
 	${CC} ${CFLAGS} -DIOCCCSIZE_STANDALONE iocccsize.c -o $@
@@ -310,3 +310,4 @@ fnamchk.o: fnamchk.c limit_ioccc.h dbg.h util.h
 txzchk.o: txzchk.c limit_ioccc.h dbg.h util.h
 jauthchk.o: jauthchk.c limit_ioccc.h dbg.h util.h json.h
 jinfochk.o: jinfochk.c limit_ioccc.h dbg.h util.h json.h
+json.o: json.c dbg.h util.h json.h
--- a/Makefile
+++ b/Makefile
@@ -82,31 +82,35 @@ SED= sed
 RPL= rpl
 TRUE= true
 
+# C source standards being used
+#
+STD_SRC= -D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE
+
 # how to compile
 #
 # note the feature test macros are required to compile on more systems e.g.
 # CentOS
-CFLAGS= -D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE -std=c99 -O3 -g3 -pedantic -Wall -Wextra
+CFLAGS= ${STD_SRC} -std=c99 -O3 -g3 -pedantic -Wall -Wextra
 
 # We test by forcing warnings to be errors so you don't have to (allegedly :-) )
 #
-#CFLAGS= -D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE -std=c99 -O3 -g3 -pedantic -Wall -Wextra -Werror
+#CFLAGS= ${STD_SRC} -std=c99 -O3 -g3 -pedantic -Wall -Wextra -Werror
 
 # NOTE: There are some things clang -Weverything warns about that are not relevant
 # 	and this for the -Weverything case, we exclude several directives
 #
-#CFLAGS= -D_BSD_SOURCE -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE -std=c99 -O3 -g3 -pedantic -Wall -Wextra -Werror -Weverything \
+#CFLAGS= ${STD_SRC} -std=c99 -O3 -g3 -pedantic -Wall -Wextra -Werror -Weverything \
 #     -Wno-poison-system-directories -Wno-unreachable-code-break -Wno-padded
 
 # NOTE: If you use ASAN, set this environment var:
 #	ASAN_OPTIONS="detect_stack_use_after_return=1"
 #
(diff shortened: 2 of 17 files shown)


lcn2:
> Do you want to define exit codes? Maybe mkiocccentry could have some too? Let me know and I'd be happy to implement them.

Probably yes.

Code such as:

```c
exit(author.issues != 0);
```

Should be changed to use an exit specific, exit codes such as:

```c
if (author.issues != 0) {
    exit(123);  /* or some such number */
}
```

So that when the tool exits with code 123 (of whatever constant is picked), someone viewing the source can determine the circumstances are the fatal error / issue was detected.

Remember that some exit codes should NOT be used unless ...:

- 0 - all is OK
- 127 - a call to an external program (via interfaces `system()`)
- >= 256 (these codes are invalid as the exit code is only an 8 bit value)
- == 0 mod 256 (a non-zero exit code will appear to the user as an exit 0 (no error) was called)

A useful convention is to reserved:

- 0 - all is OK
- >= 1 and <= 9 - some startup problem such as bad or invalid command line args, missing files, etc.
- >= 250 and <= 255 - reserve for significant process errors such as a failure for error reporting to report an error
- >= 256 - invalid exit codes that should never be used because their impact will be as if (code mod 256) was done

We use a tool, `seqcexit`

```sh
usage: /usr/local/bin/seqcexit [-h] [-v lvl] [-b bottom] [-t top] [-n] [-s] [-c] file.c [file2.c ...]

	-h		print help and exit 
	-v lvl		verbose / debugging level (def: 0)
	-b bottom	bottom exit code after wrap around (must be >=0 and < bottom and != 127) (def: 10)
			    NOTE: The sequenced exit codes in file.c can start at 0. The bottom  value
				  only applies when the codes exceed top and need to wrap around.
	-t top		top exit code range (must be > bottom  and < 256 and != 127) (def: 249)
	-n		do not change, nor create files
	-s		keep a copy of the original filenmame as filename.orig.c
	-c		continous sequencing across files (def: always reset code on a new file)

	file.c ...	C source file(s) to process

NOTE: Exit 0 can only be used once at the first exit code use: if bottom  == 0.
      Exit codes > 255 are not used: instead the next code will be set to bottom , or set to 1 if bottom  == 0.
      Exit 127 is not used: instead next value is used: usually 128 unless top == 128 in which case bottom  is used.
```

To automate sequencing exit codes.  (You may have noticed the /*ooo*/ and /*coo*/ comments that the tool uses).

It does NOT create unique exit codes if you have many (>250) code points, but it does mean that one has far fewer places to look in order to found a location of an exit point.

This is why use of explicit exit code constants is also import.  Code such as this is not helpful:

```c
exit(guess_what_value_this_variable_had_at_the_point_exit_was_called);
```

Better would be:

```c
if (something_bad_happened || foo > bar || value == magic) {
    .. stuff to report / fprintf to say stderr or a log file ...
    exit(23);
}
```

Changes are there will be few, or perhaps just one location for `exit 23`.  Along with any error or log messages, something better than "the program crashed and/or didn't work" can be determined.  :-)

xexyl:
I am going to reply to this quickly over email. More can be said tomorrow or the next day. 

I noticed that tool that you use. Is it open source or can I get it for free?

I agree with your thoughts though. I will elaborate another time. I will also reply to the other messages later as well. 

I don’t know if you saw but I finally updated my 2020 entries with an update to the cake recipe. I hope the html has the header and footer correct but if not please let me know and I will fix it. The Makefile now works for Snake as well which is good. There were some other fixes. 

Anyway if I can get access to that tool I would be happy to make use of it. Either way I will address the error codes another day  If today is any indication of tomorrow I probably won’t get to it until Wednesday but then I should be able to do a good amount  

Going now. Have a great night!

lcn2:
Thanks @xexyl

xexyl:
> > Do you want to define exit codes? Maybe mkiocccentry could have some too? Let me know and I'd be happy to implement them.
> 
> Probably yes.

I think that's a good idea. Perhaps some `#define`s are in order? (Perhaps you already did that and I haven't seen it yet.)

> 
> Code such as:
> 
> ```c
> exit(author.issues != 0);
> ```
> 
> Should be changed to use an exit specific, exit codes such as:
> 
> ```c
> if (author.issues != 0) {
>     exit(123);  /* or some such number */
> }
> ```
> 
> So that when the tool exits with code 123 (of whatever constant is picked), someone viewing the source can determine the circumstances are the fatal error / issue was detected.

Makes sense yes.

> 
> Remember that some exit codes should NOT be used unless ...:
> 
> * 0 - all is OK
> * 127 - a call to an external program (via interfaces `system()`)
> * > = 256 (these codes are invalid as the exit code is only an 8 bit value)
> * == 0 mod 256 (a non-zero exit code will appear to the user as an exit 0 (no error) was called)
> 

Indeed. I not only saw those in the tools (when using `system()` - prior to when I got involved in this repo) but also have seen some of them in the wild (obviously I've seen 0 and 1). The `>= 256` I discovered not too long ago.

As for the last one:

> == 0 mod 256 (a non-zero exit code will appear to the user as an exit 0 (no error) was called)

I am having a difficult time parsing that in my head. Perhaps it's the wording but perhaps I'm not awake enough yet. Would you please elaborate what you're trying to say here? Maybe the `== 0 % 256` is what's throwing me off (the `==` in particular might be it).


> A useful convention is to reserved:
> 
> * 0 - all is OK

Of course.

> * > = 1 and <= 9 - some startup problem such as bad or invalid command line args, missing files, etc.


> * > = 250 and <= 255 - reserve for significant process errors such as a failure for error reporting to report an error
> * > = 256 - invalid exit codes that should never be used because their impact will be as if (code mod 256) was done

As above maybe we need some `#define`s - perhaps in `util.h`? Should these be done in `err()` or should they be done elsewhere?

I'm not sure what exact ones to define right now; if you have ideas maybe you can do it or else I can do some later on when more awake (maybe when working on one or more of the tools).

> 
> We use a tool, `seqcexit`
> 
> ```shell
> usage: /usr/local/bin/seqcexit [-h] [-v lvl] [-b bottom] [-t top] [-n] [-s] [-c] file.c [file2.c ...]
> 
> 	-h		print help and exit 
> 	-v lvl		verbose / debugging level (def: 0)
> 	-b bottom	bottom exit code after wrap around (must be >=0 and < bottom and != 127) (def: 10)
> 			    NOTE: The sequenced exit codes in file.c can start at 0. The bottom  value
> 				  only applies when the codes exceed top and need to wrap around.
> 	-t top		top exit code range (must be > bottom  and < 256 and != 127) (def: 249)
> 	-n		do not change, nor create files
> 	-s		keep a copy of the original filenmame as filename.orig.c
> 	-c		continous sequencing across files (def: always reset code on a new file)
> 
> 	file.c ...	C source file(s) to process
> 
> NOTE: Exit 0 can only be used once at the first exit code use: if bottom  == 0.
>       Exit codes > 255 are not used: instead the next code will be set to bottom , or set to 1 if bottom  == 0.
>       Exit 127 is not used: instead next value is used: usually 128 unless top == 128 in which case bottom  is used.
> ```
> 
> To automate sequencing exit codes. (You may have noticed the /_ooo_/ and /_coo_/ comments that the tool uses).

I noticed not only the comments but also the reference to the tool. I'll look it up and see if I can find the tool so I could use it (I guess that might be an idea - if nothing else to see more details). I however at that point assumed it was a tool that you had written. 

> 
> It does NOT create unique exit codes if you have many (>250) code points, but it does mean that one has far fewer places to look in order to found a location of an exit point.
> 
> This is why use of explicit exit code constants is also import. Code such as this is not helpful:
> 
> ```c
> exit(guess_what_value_this_variable_had_at_the_point_exit_was_called);
> ```
> 
> Better would be:
> 
> ```c
> if (something_bad_happened || foo > bar || value == magic) {
>     .. stuff to report / fprintf to say stderr or a log file ...
>     exit(23);
> }
> ```

Ah, then maybe it's not good to have `#define`s?

> 
> Changes are there will be few, or perhaps just one location for `exit 23`. Along with any error or log messages, something better than "the program crashed and/or didn't work" can be determined. :-)

True. That's always an issue indeed.

lcn2:
> Ah, then maybe it's not good to have #defines?

Generally, yes.

xexyl:
In other words magic numbers. Perhaps it should be left to you then. 

How do you suggest I write it though so that it’s easy for you to update? I guess if I do it like I did for `txzchk` it will suffice?

lcn2:
> In other words magic numbers. Perhaps it should be left to you then.
> 
> How do you suggest I write it though so that it’s easy for you to update? I guess if I do it like I did for `txzchk` it will suffice?

Sadly yes, constants.  But not magic numbers, simple constants that differ from most other exit constants.

lcn2:
The one of:

```c
err(3, __func__, "passed NULL arg");
```

Will name the function from which it was called (that is what `__func__` resolves to).  Combined with if `3` is a fall unique exit code, then the resulting error message and/or exit code will point to where the problem occurred and what happened.

I.e., by either `exit 3` or the `name_of_function .. passed NULL arg`, you will know where and what happened.

xexyl:
> The one of:
> 
> ```c
> err(3, __func__, "passed NULL arg");
> ```
> 
> Will name the function from which it was called (that is what `__func__` resolves to). Combined with if `3` is a fall unique exit code, then the resulting error message and/or exit code will point to where the problem occurred and what happened.
> 
> I.e., by either `exit 3` or the `name_of_function .. passed NULL arg`, you will know where and what happened.

Yes I understand that. But thank you for clarifying - just in case. I was just saying how I interpret magic numbers - right or wrong or just different.

Well as you'll see in the pull request I think I'm done for the day. Sorry for the delay! I think tomorrow should be good though.

PR #77, opened 2022-02-15
Various updates

xexyl:
Moved the file scope variables of `txzchk.c` to new file `txzchk.h`.

I also fixed the man pages for commit 29789c5b.

I forgot to note that I also ran `make depend` (this in turn added an additional dependency rule that was neglected in a more recent update).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ dbg_test: dbg.c Makefile
 fnamchk: fnamchk.c dbg.o util.o Makefile
 	${CC} ${CFLAGS} fnamchk.c dbg.o util.o -o $@
 
-txzchk: txzchk.c rule_count.o dbg.o util.o Makefile
+txzchk: txzchk.c txzchk.h rule_count.o dbg.o util.o Makefile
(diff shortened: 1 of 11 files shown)


xexyl:
I noticed that I did not move the usage message variable. This is because I hadn't yet thought to do an `#ifdef TXZCHK_C` to the header file (probably it wasn't until I moved the function prototypes - so `sanity_chk()` that I did that).

Perhaps this same change could be done for the other tools and in particular the `mkiocccentry.c` one as that has a lot of file scoped variables and functions. Just to keep it a bit cleaner.

Do you think that's a good idea Landon? I think it is and I'll be happy to do it but later on -  perhaps today but if not tomorrow. That's all I'm doing for now but I expect (but am not certain) that I will return to this later on. Still haven't looked at the stuff you added yesterday (other than running `make test`).

lcn2:
Good catches, @xexyl

xexyl:
> Good catches, @xexyl

No worries. I have more I haven’t committed yet as well. If I do they will likely be done tomorrow.

PR #80, opened 2022-02-18
Many checks added to jinfochk and jauthchk

xexyl:
Much progress made! See the git log for information on some things that should be updated and for what is done (I did not list everything that's not done).


lcn2:
Good catch on the typos too, @xexyl

xexyl:
Thanks for merging it! I'll fetch it on my fork and pull locally so that I can work on it. I'm not sure when that'll be but expect more tomorrow at the latest! I guess I'll start out (like you suggested) with adding test cases.


And sure - no problem. My eyes tend to zoom in on typos though of course I've made my fair share as well. Some are easier than others to find since I have in vim the language configured; others when checking comments (which is where they usually are) I just see them.

PR #82, opened 2022-02-19
Tools updates

xexyl:
There are several bug fixes in these commits too. See the git log for details. I'm also going to raise some issues in the `jinfochk` thread though I'm not sure if I'll do that right away.

Unfortunately I didn't get as many tests in as I would like but the fixes are of course more important. I probably am done for the day but we'll see. I'll continue tomorrow if not again today.  I'm hoping that I can get these tools done within the next week or so but I'm unsure of it other than I don't think it should be too many more updates needed (more will probably occur but I don't think many more updates have to be done to get all the tests in place).


lcn2:
Excellent work @xexyl !

xexyl:
Thank you very much! I really appreciate the words of encouragement etc.!

xexyl:
I see that the git log did not carry over though. That's really unfortunate as there are some important points to consider in the log. Do you have any idea why this sometimes happens, @lcn2 ? And do you have any suggestions on what can be done to make it so this is not a problem in the future?

Perhaps a changelog file in the repo is in order? That would require making sure to update it too but maybe something to consider? Just a thought though.

lcn2:
Not sure why this happened.  The commit was made via the GitHub iOS app instead of the GutHub web site.  Perhaps that is why?

We suggest you make a minor commit, perhaps a version change and put into that commit the log that should have gone into the previous commit.

xexyl:
> Not sure why this happened. The commit was made via the GitHub iOS app instead of the GutHub web site. Perhaps that is why?

Nope. The pull request was on the website; the git commit was not: I do it from the command line. This has happened to me before as well.

> 
> We suggest you make a minor commit, perhaps a version change and put into that commit the log that should have gone into the previous commit.

I could do that. I also have the actual commit ids so I could include those as well.

PR #83, opened 2022-02-19
Many tools updates

xexyl:
I'm not sure if this will go okay: unfortunately I started with the same branch name and it looks like this has all the commits. If this is so and it's a problem please let me know and I'll delete the branches here and locally and then re-commit. I had this problem before too. There seems to be some problems with using git at the command line with GitHub - or maybe I'm just unlucky. Maybe I should just delete the fork and re-fork and do the commit again? Let me know but if so I'll get to it tomorrow in the morning.

I shortened the commit log significantly but it's still long which probably shows how I believe in thoroughness.


xexyl:
Actually I'll close the pull request now and do that later. I'll update the version of the two tools in order to do this commit; I have the commit log I used for this so it's just a matter of updating the repo locally.

lcn2:
No worries!

xexyl:
I opened a new request a few minutes ago to get it out of the way.

PR #88, opened 2022-02-22
Various updates, fixes and clean up

xexyl:
I've not added any additional tests to the `jinfochk` and `jauthchk` tools but I think these updates are better to have. I think that I will probably get the tarball check added today but I'm not sure that if I do when it will be.

The `check_common_json_fields()` function has had its definition updated so that each tool can keep a log of problems to report at the end if quiet mode is not set. Note that this function does not yet add anything to the structs and some of it will very possibly change but this will take a bit more thought and time.

There were some typo fixes as well. I also actually thought of an additional test for `jinfochk` but I'll note that on that issue - in a few minutes I should expect (though I have to open the file again to make sure I am saying it right).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,10 +4,16 @@
 
 Created this CHANGES.md markdown document.
 
-Added -T flag to mkiocccentry to print the IOCCC entry tool set release tag:
+Added -T flag to `mkiocccentry`, `fnamchk`, `txzchk`, `jstrencode` and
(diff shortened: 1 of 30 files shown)


lcn2:
Nice with @xexyl !  We will use the _create a merge commit_ option for this pull request.

xexyl:
> Nice with @xexyl ! We will use the _create a merge commit_ option for this pull request.

Thank you! And yes that merge commit does seem to be working.

PR #89, opened 2022-02-23
Minor updates and fixes

xexyl:
Move common JSON fields to json_common struct

That is to say that the fields that are written to .info.json and
.author.json are now in a common struct which is in both struct info and
struct author. Please do note that in mkiocccentry.c these still share
memory (or rather the info struct is used): previously in the functions
that refer to author and info structs it referenced e.g. info.ioccc_id
or infop->ioccc_id and the memory was only allocated in the info struct.
Now it's just info.common.ioccc_id or infop->common.ioccc_id instead.
This change is primarily for jinfochk and jauthchk in a future commit.

### DIFF
--- a/jauthchk.c
+++ b/jauthchk.c
@@ -451,7 +451,7 @@ check_author_json(char const *file, char const *fnamchk)
 		 */
 	    }
 	    /* handle regular field */
-	    if (check_common_json_fields(program_basename, file, NULL, &author, fnamchk, p, value)) {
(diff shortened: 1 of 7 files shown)


xexyl:
Expect at the least an update from me later on today. I had another difficult night but I'm hoping that I can do a bit more today at least. I'll say it in the `jinfochk` thread.

lcn2:
Thanks @xexyl

xexyl:
Thanks as well! Hope you had a good sleep.

PR #91, opened 2022-02-25
Improved extraction of JSON fields and bug fixes

xexyl:
New version for both jinfochk and jauthchk: 0.7 2022-02-25. This is a
big update in the way fields are parsed and checked.

Now when going through the json files we add all field : value pairs;
the common fields (those expected in both .info.json and .author.json)
are added to the found_common_json_fields list; else for jinfochk the
list is found_info_json_fields and for jauthchk it's
found_author_json_fields (see below for the rationale behind this naming
scheme). Duplicate values are added to the field (the same field: if the
field is "ioccc_contest" and it's seen twice then the field will have
two values rather than two fields each with one value).  After going
through the file we iterate through the two lists: first the uncommon
fields and then the common fields, checking for problems. After that we
free the lists and if any problems are found the file is reported as
invalid (with dbg() function with DBG_LOW). The exit status is 0 for no
issues found and 1 for issues found still.

As for why the name found_(common|info|author)_json_fields name: it's
because I had the thought that it might be good to have tables (arrays)
of fields to make the checking if the field should be added to the list
or not cleaner and additionally the arrays might also have the number
required and could also be used to detect how many were found. This
detail has not been fully thought out so it'll be in a future commit (or
not) but this meant that the list name (info|author|common)_json_fields
wouldn't work for the found items.


xexyl:
Okay closing this for the time... I've left out some important things by mistake. I'll reopen in a while.

xexyl:
There. Fixed those. Now I'll be afk a while. Might do more later .. not sure of that.

lcn2:
Note work, @xexyl .. we done.  Plus no merge issues on our part.  :-)

xexyl:
> Note work, @xexyl .. we done. 

Thank you!

> Plus no merge issues on our part. :-)

Yes that was very weird! FWIW: I always run `make` before pushing and I try doing `make test` - I don't always remember to do that but I try to. So if there's a problem again like that you might want to see if it's on your end. Still doesn't hurt to ask of course. :)

Hope you're having a nice sleep!

PR #92, opened 2022-02-26
Complete all common json fields checks

xexyl:
The first commit cleans up commits 8c7c56f6 and a99522d5 (the former of which fixed some issues in the latter) and also makes more consistent names of variables and parameters.

Fix memory leaks. This includes referencing struct json_field's char *value instead of 'struct json_value *values' (which is what's used).  'value' has now been removed. 

In functions the parameter for the field name is now called 'name' and the parameter for the value is now called 'val'. In the functions (most of them - the ones that directly act on a struct json_field * and/or struct json_value *) the struct json_field * is called 'field' and the struct json_value * is called 'value'. Any 'char *' for the value is called 'val' (just like in the parameter). In the check_(info|author)_json() functions the variable names are 'p' (for field) and 'value' (for field's value - i.e. the part after the ':' in the name:value pairs); and the struct json_field * is called 'field' and the struct json_value * is called 'field_value'. Similarly in the check_(info|author)_json() functions the 'char *value' is now 'char *val' and the 'struct json_value *field_value' is now 'struct json_value *value'. However the field name is still a 'char *p' as it's not only used for names.

Now all the functions that refer to the found_ lists that did not have found_ before the list name with the exception of get_common_json_field(); this is because at this point the field is not found and it's only checking if the current value of 'p' is one of the common fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,9 @@ For more information and examples, try:
 	    man ./jinfochk.1
 
 
+NOTE: This tool is currently unfinished.
+
+
 ##  jauthchk
 
 The official **IOCCC** .author.json sanity checker tool.
@@ -68,4 +71,4 @@ For more information and examples, try:
 	    man ./jauthchk.1
 
 
-
+NOTE: This tool is currently unfinished.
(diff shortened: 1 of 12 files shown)


lcn2:
Lots of good stuff in this pull request, @xexyl

xexyl:
> Lots of good stuff in this pull request, @xexyl

Thank you!

PR #93, opened 2022-02-27
Release new version of jinfochk and jauthchk

xexyl:
New version for these tools: 0.9 2022-02-27.

The `struct json_field` now has an `int field_type` which can be one of:
`JSON_NUMBER`, `JSON_BOOL`, `JSON_STRING`, `JSON_ARRAY`, `JSON_ARRAY_NUMBER`, `JSON_ARRAY_BOOL` or `JSON_ARRAY_STRING`. `JSON_NULL` also exists: this is only as an end of list marker. Note that the array versions are not
handled yet because arrays are not handled yet and the JSON_ARRAY_
macros might be changed but I needed to distinguish that they're part of
an array. The `JSON_STRING` and `JSON_ARRAY_STRING` are used to decide
whether or not to remove '"'s from the values (they're always removed
from the field names). It used to be done via a bunch of `strcmp() != 0`
results i.e. by a long list of names.

Added list for both .info.json and .author.json. All the fields in the
file are added but because the arrays are not handled the array fields
will not be added to the list (this also means that anything not in the
list will not be added; however in the future - once arrays are dealt
with - these will be reported as an issue but this cannot happen yet as
it would make the tools fail).

Clean up the adding fields to lists and checking fields lists. This is
via the new functions:

```c
    /*
     * get_info_json_field  -           check if name is a .info.json field
     *                                  and if it is add it to the found_info_json
     *                                  list.
     *
     * given:
     *
     *      file    - the file being parsed (path to)
     *      name    - the field name
     *      val     - the value of the field
     *
     * returns:
     *      1 ==> if the name is a .info.json field
     *      0 ==> if it's not one of the .info.json fields
     *
     * NOTE: Does not return on error (NULL pointers).
     */
```
and

```c
    /*
     * check_found_info_json_fields - check that all the fields in the
     *                                  found_info_json_fields table have valid
     *                                  values
     *
     *  given:
     *
     *          file        - .info.json file we're checking
     *          test        - if test mode: ignore some checks
     *
     *
     * returns:
     *      >0 ==> the number of issues found
     *      0 ==> if no issues were found
     */
```

... as well as the author equivalents. The functions for the common
fields version have also been cleaned up some.

New function in util.c:

```c
    /* is_number        - if the string str is a number
     *
     * given:
     *
     *      str         - string to check
     *
     * For our purposes a number is defined as: a string that starts with either a
     * '-' or '+' and beyond that no other characters but [0-9] (any count).
     *
     * returns:
     *
     *      true    ==> str is a number
     *      false   ==> str is not a number
     *
     * NOTE: This function does not return on error (str == NULL). An empty string
     * is considered not a number.
     */
```
... which is used in the json check functions specific to json field
type. Maybe this should be done in other places as well.

In test mode don't complain about different `INFO_VERSION` (in jinfochk)
and `AUTHOR_VERSION` (in jauthchk).

Typo fixes and comments updates.

Some additional sanity checks have been added. One that comes to mind is
we warn about empty values (this does not count as an issue because the
checks below will trigger that - this is only to show why it would
happen).

The `find_json_field_in_table()` function no longer can erroneously find a
field with an empty name and you can now get the number of elements via
passing an empty string or a NULL pointer (if empty it's set to NULL).

There might be other changes I've forgotten.

### DIFF
--- a/dbg.h
+++ b/dbg.h
@@ -59,7 +59,7 @@ typedef unsigned char bool;
  * backward compatibility
  *
  * Not all compilers support __attribute__ nor do they support __has_builtin.
- * For example, MSVC, TenDRA and Little C Compiler doesn't support __attribute__.
+ * For example, MSVC, TenDRA and Little C Compiler don't support __attribute__.
  * Early gcc does not support __attribute__.
  *
  * Not all compiles have __has_builtin
--- a/jauthchk.c
+++ b/jauthchk.c
@@ -167,6 +167,45 @@ main(int argc, char **argv)
     exit(issues != 0); /*ooo*/
 }
 
+/* check_author_json_fields_table	    - perform some sanity checks on the
+ *					      author_json_fields table
+ *
+ * This function checks if JSON_NULL is used on any field other than the NULL
+ * field. It also makes sure that each field_type is valid. More tests might be
+ * devised later on but this is a good start (27 Feb 2022).
+ *
+ * This function does not return on error.
+ */
+static void
+check_author_json_fields_table(void)
+{
+    size_t loc;
+
+    for (loc = 0; author_json_fields[loc].name != NULL; ++loc) {
+	switch (author_json_fields[loc].field_type) {
+	    case JSON_NULL:
+		if (author_json_fields[loc].name != NULL) {
+		    err(5, __func__, "found invalid data in author_json_fields table #0");
+		    not_reached();
+		}
+		break;
+	    case JSON_NUMBER:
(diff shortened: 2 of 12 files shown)


xexyl:
If I do anything else today it will probably be adding sanity checks of boolean mismatches etc. I'm not sure if I will though. If not expect that probably tomorrow.

lcn2:
Nice work @xexyl

xexyl:
> Nice work @xexyl

Thank you!

PR #94, opened 2022-02-28
jinfochk and jauthchk: additional checks added

xexyl:
Add empty_override and rule 2a/2b size mismatches: If empty_override ==
true and either the rule 2a or 2b size != 0 then it's an issue; if false
and either rule 2a or 2b size == 0 it's an issue. This has also been
added to mkiocccentry: if either size is 0 then warn about empty prog.c.

The 'test_mode' field is now common to both .info.json and .author.json
and we verify test_mode consistency with fnamchk: that is to say that if
test_mode == true the tarball filename must start with "entry.test-"; if
false it must be a UUID. This is done by adding two options to fnamchk:
-u: if filename is a test mode filename then it's an error; -t: if
filename is a UUID filename it is an error.

I've bumped the version of fnamchk, jinfochk, jauthchk and mkiocccentry
for the above changes.

### DIFF
--- a/fnamchk.1
+++ b/fnamchk.1
@@ -2,7 +2,7 @@
 .SH NAME
 fnamchk \- IOCCC compressed tarball filename sanity check tool
 .SH SYNOPSIS
-\fBfnamchk [\-h] [\-v level] [\-V] [\-T] filepath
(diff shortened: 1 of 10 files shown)


lcn2:
Good stuff, @xexyl

xexyl:
> Good stuff, @xexyl

Thank you!

PR #99, opened 2022-03-05
Add sanity

xexyl:
...because sometimes we're all a little insane, some of us are a lot
insane and code is often very insane. :-)

What this commit does is add sanity.{h,c} which has a new function:
ioccc_sanity_chk() which runs sanity checks on all the tables: the
location table, the UTF-8 POSIX map and the three JSON fields tables.

The jinfochk and jauthchk fields tables checks were moved to json.c
because the tools which have a sanity check function all run the new
ioccc_sanity_chk (because if something is not right in one or more of
the tables it suggests there's a problem in the toolkit) and jinfochk.c
and jauthchk.c obviously have main() each as well as verbosity_level.


lcn2:
Thanks.  We will consider moving `check_utf8_posix_map()` to `utf8_posix_map.c` and expand its use to fill in the string lengths that are to be added.

xexyl:
> Thanks.
> 
> Correction: We will update `check_utf8_posix_map()` I'm `utf8_posix_map.c` and expand its use to fill in the string lengths that are to be added.

I know I already said it but thank you.

PR #100, opened 2022-03-06
JSON tests, typo and other fixes and updates 

xexyl:
Make the new json test script work: due to some missing directories it would abort.

This commit also includes a quiet option for `mkiocccentry` which is used in `make test`. At first I thought it would be useful with the `ioccc_sanity_chk()` (now suffixed as `_chks()` as are all the others) but then I remembered the messages in that function are via `dbg()` so there's no need for a quiet mode there. Still seems like something that might be good to have so I kept it in.

See the git log for other fixes (typos).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ dkms.conf
 
 # Makefile backup (make depend specific)
 Makefile.orig
+
+# macOS files
+.DS_Store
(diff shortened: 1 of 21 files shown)


lcn2:
Thanks @xexyl , we will start work on `default_handle()` in util.c.

xexyl:
> Thanks @xexyl , we will start work on `default_handle()` in util.c.
> 
> Expect a commit with a code stub (before the full code is ready) as there are a few things a few other fines the need to change in preparation

Thanks for the note and for the merge. I'll look at it tomorrow.

PR #103, opened 2022-03-08
Various bug fixes and updates

xexyl:
The common json field author_count is no longer common but instead only
in .author.json.

The common json field iocccsize_version is no longer common but instead
only in .info.json.

In the functions check_found_info_json_fields() and
check_common_json_fields() no longer allow missing fields to be not a
problem if test mode is enabled.  This was causing some tests to fail
and since the problematic fields are no longer common this should be
okay. Because not all fields are checked in jauthchk this was not done
so this is not a problem. The test option does still allow for different
versions of the tools to be there however.

This commit resolves a number of cases in the test_JSON directories
specific to jinfochk. As jauthchk does not yet parse the authors array
I'm not sure where the test files stand with it. I am aware of some
test_JSON files specific to jinfochk that are still a problem which will
be worked out in a future commit. There seems to be an issue with the
test script which has caused some confusion and this will have to be
worked out as well.

### DIFF
--- a/jauthchk.c
+++ b/jauthchk.c
@@ -787,6 +787,7 @@ check_found_author_json_fields(char const *file, bool test)
     size_t loc = 0;	/* location in the author_json_fields table */
     int issues = 0;
     size_t val_length = 0;  /* current value length */
+    int author_count = 0;
 
     /*
      * firewall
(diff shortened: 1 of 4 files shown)


lcn2:
Thanks @xexyl and sorry if we misled you about this field being common in the past.

xexyl:
> Thanks @xexyl and sorry if we misled you about this field being common in the past.

No worries! As things are worked on we find requirements can change. That goes for life in general but programming as well. And I believe in fact this is one of the goals of the contest: that sometimes there are mismatches and problems with requirements.

PR #108, opened 2022-03-10
Various fixes

xexyl:
The loop did not check the final character because the for loop was:

     for (i=start; i < len-1; ++i) {

Where len was strlen(str). The loop is now correctly:

     for (i=start; i < len; ++i) {

This does fix one issue in test_JSON/info.json/bad/info.title-2.json.
However there still is another problem in that file which is not yet
picked up: that will be addressed in a future commit.


xexyl:
Sorry about the mix-up requiring merging. I was trying to fix the commit log but of course I had already pushed so this was a mistake. If I had an easy way to undo that I would but I'm not sure if this is possible. I guess this doesn't really matter to you but it does to me.

I'm done for the day though for certain: I'm just too tired and I have to be out in a bit and after that I'll be even more tired. Very probable it'll be an even earlier night than usual though I've been considering trying to stay up a bit later for several days in a row to see if I can somehow sleep in past 2. Either way if I go for flex and bison I very possibly won't be doing any real changes to the parser for some days but we'll see what I come up with. I might take it easy for a few days anyway.

I do feel good though that several tests have been resolved and I will worry about more changes tomorrow or the following days.

lcn2:
Thanks and good morning, good afternoon or good evening depending on your time zone and latitude @xexyl

xexyl:
Seems the pull request was not closed. I'll do that and then make a new pull request with the change to `txzchk` plus the sequenced exit codes.

And good morning to you as well!

PR #111, opened 2022-03-12
Write new function for json warnings: jwarn()

xexyl:
/*
     * jwarn - issue a JSON warning message
     *
     * given:
     *      program name of program e.g. jinfochk, jauthchk etc.
     *      name    name of function issuing the warning
     *      code    warning code
     *      line    line number of the calling file (__LINE__ macro)
     *      fmt     format of the warning
     *      ...     optional format args
     *
     * Example:
     *
     *      jwarn(program, __func__, 1, __LINE__, "unexpected foobar: %d", value);
     *
     * NOTE: We warn with extra newlines to help internal fault messages stand out.
     *       Normally one should NOT include newlines in warn messages.
     */
    void
    jwarn(char const *program, char const *name, int code, int line, char const *fmt, ...)
    {
    }

This function will construct a warning message for JSON specific issues.
Example output:

    jinfochk: check_found_common_json_fields: {JSON-0001}: line 2308: field 'mkiocccentry_version' found 2 times but is only allowed once in file test_work/test-0/.info.json

This would be generated by changing:

            warn(__func__, "field '%s' found %ju times but is only allowed once in file %s", common_field->name, (uintmax_t)common_field->count, file);

to be:

	    jwarn(program, __func__, 1, __LINE__, "field '%s' found %ju times but is only allowed once in file %s", ...);

All the JSON warn() calls need to be converted to this function and
unique warn codes have to be decided upon and documented. That will
likely come next but I'm not sure if that's today. The messages should
possibly also be simplified but I'll worry about this another time.

### DIFF
--- a/json-test.sh
+++ b/json-test.sh
@@ -232,7 +232,7 @@ if [[ ! -f $JCODECHK ]]; then
     exit 4
 fi
 if [[ ! -x $JCODECHK ]]; then
-    echo "$0: ERROR: jcodechk.sh not exeutable: $JCODECHK" 1>&2
+    echo "$0: ERROR: jcodechk.sh not executable: $JCODECHK" 1>&2
     exit 4
 fi
 
--- a/json.c
+++ b/json.c
@@ -1639,6 +1639,86 @@ struct json_field author_json_fields[] =
 };
 size_t SIZEOF_AUTHOR_JSON_FIELDS_TABLE = TBLLEN(author_json_fields);
 
+/*
+ * jwarn - issue a JSON warning message
+ *
+ * given:
+ *	program	name of program e.g. jinfochk, jauthchk etc.
+ *	name	name of function issuing the warning
+ *	code	warning code
+ *	line	line number of the calling file (__LINE__ macro)
+ *	fmt	format of the warning
+ *	...	optional format args
+ *
+ * Example:
+ *
+ *	jwarn(program, __func__, 1, __LINE__, "unexpected foobar: %d", value);
+ *
+ * NOTE: We warn with extra newlines to help internal fault messages stand out.
+ *	 Normally one should NOT include newlines in warn messages.
+ */
+void
+jwarn(char const *program, char const *name, int code, int line, char const *fmt, ...)
+{
+    va_list ap;		/* argument pointer */
+    int ret;		/* libc function return code */
+    int saved_errno;	/* errno at function start */
+
+    /*
+     * save errno so we can restore it before returning
+     */
+    saved_errno = errno;
+
+    /*
+     * start the var arg setup and fetch our first arg
+     */
+    va_start(ap, fmt);
+
+    /*
+     * firewall
+     */
+    if (program == NULL) {
+	program = "((NULL program))";
+	warn(__func__, "\nWarning: in jwarn(): called with NULL program, forcing name: %s\n", program);
+    }
+    if (name == NULL) {
+	name = "((NULL name))";
+	warn(__func__, "\nWarning: in jwarn(): called with NULL name, forcing name: %s\n", name);
+    }
+    if (fmt == NULL) {
+	fmt = "((NULL fmt))";
+	warn(__func__, "\nWarning: in jwarn(): called with NULL fmt, forcing fmt: %s\n", fmt);
+    }
+
+    ret = fprintf(stderr, "%s: %s: {JSON-%04d}: line %d: ", program, name, code, line);
+    if (ret < 0) {
+	(void) fprintf(stderr, "\nWarning: in jwarn(%s, %s, %d, %d, %s, ...): fprintf returned error: %d\n", program, name, code, line, fmt, ret);
+    }
+    ret = vfprintf(stderr, fmt, ap);
+    if (ret < 0) {
+	(void) fprintf(stderr, "\nWarning: in jwarn(%s, %s, %d, %d, %s, ...): fprintf returned error: %d\n", program, name, code, line, fmt, ret);
+    }
+    ret = fputc('\n', stderr);
+    if (ret != '\n') {
+	(void) fprintf(stderr, "\nWarning: in jwarn(%s, %s, %d, %d, %s, ...): fputc returned error: %d\n", program, name, code, line, fmt, ret);
+    }
+    ret = fflush(stderr);
+    if (ret < 0) {
+	(void) fprintf(stderr, "\nWarning: in jwarn(%s, %s, %d, %d, %s, ...): fflush returned error: %d\n", program, name, code, line, fmt, ret);
+    }
+
+    /*
+     * clean up stdarg stuff
+     */
+    va_end(ap);
+
+    /*
+     * restore previous errno value
+     */
+    errno = saved_errno;
+    return;
+}
+
 
 /* find_json_field_in_table	    - find field 'name' in json table
  *
--- a/json.h
+++ b/json.h
@@ -261,6 +261,8 @@ extern int get_common_json_field(char const *program, char const *file, char *na
 extern int check_found_common_json_fields(char const *program, char const *file, char const *fnamchk, bool test);
 extern struct json_field *new_json_field(char const *name, char const *val);
 extern struct json_value *add_json_value(struct json_field *field, char const *val);
+extern void jwarn(const char *program, char const *name, int code, int line, const char *fmt, ...) \
+	__attribute__((format(printf, 5, 6)));		/* 2=format 3=params */
 /* free() functions */
 extern void free_json_field_values(struct json_field *field);
 extern void free_found_common_json_fields(void);


xexyl:
Question for you Landon:

What should be done about the warning code? I mean to say: if it's out of range what should be done? Should a specific warning code be used to indicate that the warning code is out of range? Perhaps (because it amuses me though maybe it would cause some problems for believers? I'm not sure and I'd like you feedback on this too) 666?

Then if someone passes in say 100000 as the code it would be changed to 666 (or whatever) and then that code could be specifically checked for in the script so that it can be fixed. What are your thoughts on this and the number?

xexyl:
Also what if the line number < 0 ?

lcn2:
Good question about out of range `code` values being passed to `jwarn()` @xexylL

If the `code argument passed is `< 0`, then have the _firewall_ code of `jwarn()` throw an `err()` and abort.

We don't suggest that your object to large codes, just let them print as:

```c
" ... {JSON-%04d} ... "
```

with a minimum of 4 digits (so the format prints leafing 0's if needed).

We can change our `jcodechk` to allow JSON codes of 4 or more digits.

> BTW: The reason for wanting a minimum of 4 digits is to make it less likely that the string is part of some normal JSON text.

lcn2:
If you do implement the idea of printing the _offending line idea_ that we raised elsewhere, then perhaps:

```c

    /*
     * jwarn - issue a JSON warning message
     *
     * given:
     *      program name of program e.g. jinfochk, jauthchk etc.
     *      name    name of function issuing the warning
     *      code    warning code
     *      ignore  If non-NULL, list of code values to ignore
     *      line    If non-NULL, beginning of offending JSON line
     *      linenum line number within the JSON file that is a problem
     *      fmt     format of the warning
     *      ...     optional format args
     *
     * Example:
     *
     *      jwarn(program, __func__, 1, ignore, linenum, line, "unexpected foobar: %d", value);
     *
     * NOTE: We warn with extra newlines to help internal fault messages stand out.
     *       Normally one should NOT include newlines in warn messages.
     *
     * NOTE: The ignore pointer points to an array of code values that, if
     * one of them matches code, then this function will just return
     * as if no warning was raised.  The ignore array is terminated by a value < 0 such as -1.
     *
     * NOTE: If is OK to pass line and exclude pointers as NULL.
     */
    void
    jwarn(char const *program, char const *name, int code, int const *ignore, chat const *line, int linenum, char const *fmt, ...)
    {
    }
```

BTW: We think it would be more useful to print the line number within the JSON blob being parsed, than the `__LINE__` number of the C code that raised it.

xexyl:
> Good question about out of range `code` values being passed to `jwarn()` @xexylL
> 
> If the `code argument passed is `< 0`, then have the _firewall_ code of `jwarn()`throw an`err()` and abort.

Will do. I hope I remember anyway. I'm making a mental note of it anyway for when I work on it next.

> 
> We don't suggest that your object to large codes, just let them print as:
> 
> ```c
> " ... {JSON-%04d} ... "
> ```
> 

This is indeed exactly how I have it.

> with a minimum of 4 digits (so the format prints leafing 0's if needed).
> 
> We can change our `jcodechk` to allow JSON codes of 4 or more digits.
> 

True.

> > BTW: The reason for wanting a minimum of 4 digits is to make it less likely that the string is part of some normal JSON text.

Ah. That sounds reasonable.

xexyl:
> If you do implement the idea of _printing the offending line_ idea that we raised elsewhere, then perhaps:
> 
> ```c
>     /*
>      * jwarn - issue a JSON warning message
>      *
>      * given:
>      *      program name of program e.g. jinfochk, jauthchk etc.
>      *      name    name of function issuing the warning
>      *      code    warning code
>      *      ignore  If non-NULL, list of code values to ignore
>      *      line    If non-NULL, beginning of offending JSON line
>      *      linenum line number within the JSON file that is a problem
>      *      fmt     format of the warning
>      *      ...     optional format args
>      *
>      * Example:
>      *
>      *      jwarn(program, __func__, 1, ignore, linenum, line, "unexpected foobar: %d", value);
>      *
>      * NOTE: We warn with extra newlines to help internal fault messages stand out.
>      *       Normally one should NOT include newlines in warn messages.
>      *
>      * NOTE: The ignore pointer points to an array of code values that, if
>      * one of them matches code, then this function will just return
>      * as if no warning was raised.  The ignore array is terminated by a value < 0 such as -1.
>      *
>      * NOTE: If is OK to pass line and exclude pointers as NULL.
>      */
>     void
>     jwarn(char const *program, char const *name, int code, int const *ignore, chat const *line, int linenum, char const *fmt, ...)
>     {
>     }
> ```
> 
> BTW: We think it would be more useful to print the line number within the JSON blob being parsed, than the `__LINE__` number of the C code that raised it.


Thanks. This is probably true once the messages are absolutely unique (this probably already is the case) and can be identified easily (this is too). I think right now having the line number is good though because it can tell me where to look to verify if it's an issue with the code or if it's the file (in the cases it's not obvious). This will all be changing in time though so perhaps this is a good idea for later. I'll save the above in a text file for later reference.

lcn2:
We suggest @xexyl, that you think about this and rest on it.

We hope these suggestions help you help other humans who encounter problems with IOCCC related JSON files.

Rest well!

xexyl:
> We suggest @xexyl, that you think about this and rest on it.
> 
> We hope these suggestions help you help other humans who encounter problems with IOCCC related JSON files.
> 
> Rest well!

Thank you! I will rest as best as I can anyway. Have a nice day! And yes the suggestions are good so thanks for that too. Probably more from me tomorrow.

lcn2:
Likewise @xexyl, rest well.

BTW: We added the `filename` argument idea because we through you might need it for printing filenames under the `JSON_test` tree.

xexyl:
> Likewise @xexyl, rest well.
> 
> BTW: We added the `filename` argument idea because we through you might need it for printing filenames under the `JSON_test` tree.

Thanks for letting me know. I just was going to include filename as a parameter so that it doesn't have to print filenames but I can always check for NULL and only print if it's != NULL.

You rest well too!

xexyl:
Oh you mean you suggested it for that reason. I thought maybe you updated it. Right. Well I'll worry about it next time.

Have a nice day!

lcn2:
> This is probably true once the messages are absolutely unique (this probably already is the case) and can be identified easily (this is too).

You might want to create a `CODE(x)` macro to help single out those `{JSON-9999}` numbers.  So instead of:

```c
    jwarn(program, ..., 9999, ...);
```

You can write:

```c
    jwarn(program, ..., CODE(9999), ...);
```

It looks a little less like s magic number AND it helps you see and search for JSON codes in your C code: even though `CODE(1234)` is just a macro for `1234`.

```c
#define CODE(x) (x)
```

xexyl:
> > This is probably true once the messages are absolutely unique (this probably already is the case) and can be identified easily (this is too).
> 
> A last thought before we sleep, @xexyl:
> 
> You might want to create a `JCODE(x)` macro to help identify those JSON code numbers. So instead of:
> 
> ```c
>     jwarn(program, __func__, 263, ignore, filename, linenum, line, "unexpected foobar: %d", value);
> ```
> 
> You can write:
> 
> ```c
>     jwarn(program, __func__, JCODE(263), ignore, filename, linenum, line, "unexpected foobar: %d", value);
> ```
> 
> It looks a little less like some magic number AND it helps you see and search for JSON codes in your C code: even though `JCODE(1234)` is just a macro for `1234`.
> 
> ```c
> #define JCODE(x) (x)
> ```

Great idea! I'll also add two defines for min and max codes.

lcn2:
With commit df9c29de3a826ed0693bd63f38bbd88325ad9a6f, we added support for ignore codes.

We made is easier to parse on the command line by just doing `-W code`.

    Added -W code support to jinfochk and jauthchk

    The -W code may be supplied on the ./jinfochk and ./jauthchk command lines:

```sh
    usage: ./jinfochk [-h] [-v level] [-q] [-V] [-T] [-s] [-F fnamchk] [-t] [-W code] ... file
            ...
            -W code                 Add code to the list of JSON error code to ignore
            ...

    usage: ./jauthchk [-h] [-v level] [-V] [-q] [-T]  [-s] [-F fnamchk] [-t] [-W code] ... file
            ...
            -W code                 Add code to the list of JSON error code to ignore
            ...
```

    The jinfochk(1) and jauthchk(1) man pages have been updated.

    Added the is_code_ignored(code) to json.c so that jwarn() code can determine
    if the code number passed to it should be ignored:

```c
    /*
     * is_code_ignored - determine of a code is to be ignored
     *
     * given:
     *      code    - code to test if code should be ignored
     *
     * returns:
     *      true ==> code is in ignore_code_set[] and should be ignored
     *      false ==> code is not in ignore_code_set[] and should NOT be igored
     */
    bool
    is_code_ignored(int code)
    {
        ...
    }
```

We did some initial testing and the code looks good.  However you may want to try it so more as we have to rush off to a meeting.

xexyl:
> With commit [df9c29d](https://github.com/ioccc-src/mkiocccentry/commit/df9c29de3a826ed0693bd63f38bbd88325ad9a6f), we added support for ignore codes.
> 
> We made is easier to parse on the command line by just doing `-W code`.
> 
> ```
> Added -W code support to jinfochk and jauthchk
> 
> The -W code may be supplied on the ./jinfochk and ./jauthchk command lines:
> ```
> 
>  ```shell
>     usage: ./jinfochk [-h] [-v level] [-q] [-V] [-T] [-s] [-F fnamchk] [-t] [-W code] ... file
>             ...
>             -W code                 Add code to the list of JSON error code to ignore
>             ...
> 
>     usage: ./jauthchk [-h] [-v level] [-V] [-q] [-T]  [-s] [-F fnamchk] [-t] [-W code] ... file
>             ...
>             -W code                 Add code to the list of JSON error code to ignore
>             ...
> ```
> 
>  ```
> The jinfochk(1) and jauthchk(1) man pages have been updated.
> 
> Added the is_code_ignored(code) to json.c so that jwarn() code can determine
> if the code number passed to it should be ignored:
> ```
> 
>  ```c
>     /*
>      * is_code_ignored - determine of a code is to be ignored
>      *
>      * given:
>      *      code    - code to test if code should be ignored
>      *
>      * returns:
>      *      true ==> code is in ignore_code_set[] and should be ignored
>      *      false ==> code is not in ignore_code_set[] and should NOT be igored
>      */
>     bool
>     is_code_ignored(int code)
>     {
>         ...
>     }
> ```
> 
> We did some initial testing and the code looks good. However you may want to try it so more as we have to rush off to a meeting.

Thanks! I haven't looked at it so I don't know if you can specify it more than once but if not do you think it should be changed to do it? I think it's a good idea if we're going to go this way. Hope the meeting goes/went well.

lcn2:
@xexyl: You can use multiple `-W code` on the command line.  You can even duplicate codes (although useless to do so) with no harm:

```sh
./jinfochk ... -W 123 -W 234 -W 123 -W 1212 foo.json
```

BCNU :-)

xexyl:
> @xexyl: You can use multiple `-W code` on the command line. You can even duplicate codes (although useless to do so) with no harm:
> 
> ```shell
> ./jinfochk ... -W 123 -W 234 -W 123 -W 1212 foo.json
> ```
> 
> BCNU :-)

Yes I took a look. Thanks.

xexyl:
But what do you mean by BCNU though?

lcn2:
BCNU is hacker speak for:

> be seeing uou

as in:

> Be C(s)eeiNg yoU

and thus:

> BCNU

It a shorthand for early UNIX chats where we waned to minimize typing.

If you were really in a rush, too used:

> -oo-

Now:

-oo-

xexyl:
> BCNU is hacker speak for:
> 
> > be seeing uou
> 
> as in:
> 
> > Be C(s)eeiNg yoU

That's one I've never seen. Thanks!

> 
> and thus:
> 
> > BCNU
> 
> It a shorthand for early UNIX chats where we waned to minimize typing.
> 
> If you were really in a rush, too used:
> 
> > -oo-
> 
> Now:
> 
> -oo-


Over and out? That's one I'm familiar with. Or is there another meaning?

xexyl:
> BTW: @xexyl, if do implement the _printing the offending line_ idea that we raised elsewhere, then perhaps:
> 
> ```c
>     /*
>      * jwarn - issue a JSON warning message
>      *
>      * given:
>      *      program name of program e.g. jinfochk, jauthchk etc.
>      *      name    name of function issuing the warning
>      *      code    warning code
>      *      ignore  If non-NULL, list of code values to ignore
>      *      filename   If non-NULL, the name of the JSON file being parsed
>      *      line    If non-NULL, beginning of offending JSON line
>      *      linenum line number within the JSON file that is a problem
>      *      fmt     format of the warning
>      *      ...     optional format args
>      *
>      * Example:
>      *
>      *      jwarn(program, __func__, 1, ignore, filename, linenum, line, "unexpected foobar: %d", value);
>      *
>      * NOTE: We warn with extra newlines to help internal fault messages stand out.
>      *       Normally one should NOT include newlines in warn messages.
>      *
>      * NOTE: The ignore pointer points to an array of code values that, if
>      * one of them matches code, then this function will just return
>      * as if no warning was raised.  The ignore array is terminated by a value < 0 such as -1.
>      *
>      * NOTE: If is OK to pass ignore, filename and/or line as NULL pointers to this function,
>      *.      as those NULL values have meaning to this function to ignore certain things.
>      */
>     void
>     jwarn(char const *program, char const *name, int code, int const *ignore, char const *filename, char const *line, int linenum, char const *fmt, ...)
>     {
>     }
> ```
> 
> BTW: We think it would be more useful to print the line number within the JSON blob being parsed, than the `__LINE__` number of the C code that raised it.
> 
> **UPDATE**: If you do add the `filename` argument, the reason to allow NULL is when the JSON code being parsed is from just a block of memory. If you are parsing from `stdin` then we recommend using the string `"((stdin))"` or something like that. If `filename` is NULL, then print with the filename `"((NULL))"`
> 

That's a good idea. Alternatively it could just be that if it's NULL it's not printed since it might suggest that the caller doesn't want a filename printed or there is no file (not sure where that would come from but maybe if test code would have a block of json text it would be no file though it could also be `"((NULL))"` as well).


> Here is a concept example for how to use the filename:
> 
> ```
> # file: foo.json line: 23:    { "title" : "!!!title!!!" },
> {JSON-0031}: invalid characters found in title
> ```
> 
Ah. Now that could be something indeed! I might actually add that then so that the line filename is on a different line so that it's not such a long line. I think I will do that. Thanks!


> **UPDATE**: While `code` < 0 is invalid (will throw an `err()` call), you might want to reserve the code of 0 for some special future case. Don't make the code of 0 invalid, just start using code numbers starting at 1.

Good idea. I'll do that. I'm going to (I think) start working on it in a bit and I'll start from this discussion.

lcn2:
> Over and out? That's one I'm familiar with.

Yes.

xexyl:
> > Over and out? That's one I'm familiar with.
> 
> Yes.

Thanks. I've come up with an idea for now to have line numbers even if the line number idea will change once a proper json parser is in place. Right now working on it but will see if I finish it. It's not complicated but it's a lot to think about - all the changes - and I'm having a problem right now with what 'doesn't looks right' even though yes I know that the message format can all change.

PR #116, opened 2022-03-16
Version handling and json errors/warnings updates

xexyl:


xexyl:
This will probably be all I get done today: the last few days have been rather difficult and that includes today. I was hoping to work on `jchk` but I don't foresee this happening at all today and I don't even think I will start it.

I do believe that I might do that before working on the parser but I'm not sure of that. I've made some decisions on warning/error codes (usage of reserved codes and also to not use the json warn/error functions for something that is likely to not be needed or changed once the json parser is done) but these could change in time.

lcn2:
Hope your next days are better for you @xexyl

xexyl:
> Hope your next days are better for you @xexyl

Thank you!

PR #120, opened 2022-03-19
Makefile and JSON updates + some minor fixes

xexyl:
Add to Makefile (right after declaring variables):

    -include makefile.config

This tries to includes makefile.config for any local/user specific
configurations. The - makes it not an error if the file does not exist
(since it normally would not exist). This is important for me when I
start using flex(1) and bison(1) under my MacBook Pro (this is not a
problem on my linux boxes because the versions of flex(1) and bison(1)
are not out of date like the default ones under macOS are) as MacPorts
has an unusual prefix /opt/local.

The Makefile now defines the variables FLEX and BISON:

    FLEX = flex
    BISON = bison

However in my makefile.config (lower case as I don't need it earlier in
the directory listing) I override the variables like so:

	FLEX = /opt/local/bin/flex
	BISON = /opt/local/bin/bison

This means that whenever the Makefile refers to ${FLEX} it will (in my
fork of the repo / dev directory on my MacBook Pro) substitute in
/opt/local/bin/flex and when the Makefile refers to ${BISON} it'll
substitute in /opt/local/bin/bison but it won't affect other systems.
In the master repo on this same system it would substitute in just 'flex' and
'bison' respectively (again unless I add a file makefile.config to
override these variables).

This ensures that the proper paths are used when I'm working under my
MacBook Pro (which is the system I primarily use for this) but under
other systems the default path will be used (though I intend to make it
so you don't have to have flex/bison installed to compile the json
parser the .y and .l files will still be added to the repo).

And yes this means that at that point in the Makefile you can override
things (this is another reason it's before any rules). Perhaps it should
come after the CFLAGS but I'll let Landon decide if that's a good idea
or not.

The makefile.config has been added to .gitignore for obvious reasons.

There were also some typo fixes in the Makefile.


xexyl:
I'm going to take a break now and when I come back I hope to (but am not certain I will be able to) work on the other changes I have and then commit those. We shall see. But this is an important change anyway for when I start working on the parser which I hope will be sooner than later and preferably even soon.

xexyl:
Okay commit 36096e8cd is probably the last commit of the day. I don't particularly feel up to reading right now either I'm afraid but I hope to get back to the book tomorrow if I don't later on today. Even so, the Makefile update is the first step to working on the parser as it prepares the Makefile for when I need flex and bison.

lcn2:
We will accept this pull request.

We recommend that 'Makefile.config' be renamed 'Makefile.local', however to reflect local configuration such as yo described.

xexyl:
> We will accept this pull request.
> 
> We recommend that 'Makefile.config' be renamed 'Makefile.local', however to reflect local configuration such as yo described.

Done.

PR #122, opened 2022-03-20
Simplify jparse wrt string versus file args

xexyl:
If you want jparse to parse a string use -S <string>; anything left on
the command line will be assumed to be a file. Just like before - means
stdin. You can specify any file more than once and that includes stdin
(and just like before it takes care to not close stdin and makes sure to
clear the EOF and error flags).

If -S not used and there are no files specified (nothing left on the
command line) it is an error.

A note of interest on the subject of -S when used with -s: -s is
considered for the next -S arg only (for strings):

    /*
     * XXX Rather than having an option to disable strict mode so that
     * in the same invocation we can test some strings in strict mode
     * and some not strict after each string is parsed the strict mode
     * is disabled so that so that another -s has to be specified prior
     * to the string. This does mean that if you want strict parsing of
     * files and you specify the -S option then you must have -s after
     * the string args.
     *
     * But the question is: should it be this way or should it be
     * another design choice? For example should there be an option that
     * specifically disables strict mode so that one can not worry about
     * having to specify -s repeatedly? I think it might be better this
     * way but I'm not sure what letter should do it. Perhaps -x? If we
     * didn't use -S for string it could be S but we do so that won't
     * work.
     */

What this means is:

    ./jparse -s -S '{},' -S '{},'

would parse (once parser is done that is: right now it simply writes the
string/file's contents to stdout) the first '{},' in strict mode and the
second would not be in strict mode; if you were to do:

    ./jparse -s -S '{},' file

it would parse the '{},' in strict mode and file would not be in strict
mode; and if you were to do:

    ./jparse -s -S '{},' -s file

it would parse the both '{},' and file in strict mode.

The reason I chose it this way (and it's certainly up for discussion) is
that whereas before the options were specified first (thus if you wanted
strict mode you would have it) with -S you don't have that so you have
to specify it before (obviously). But what if you don't want all parsing
to be in strict mode? The other option as I said is an option
specifically to disable strict mode but whether that's better or more
desirable I don't know and if it is I don't know what letter to use. Of
course right now strict mode is unused so this is just for the future
but this way it exists for when it's possible.

### DIFF
--- a/jparse.c
+++ b/jparse.c
@@ -36,6 +36,188 @@
  */
 #define REQUIRED_ARGS (0)	/* number of required arguments on the command line */
 
+/* parse_file	    - parses file file
+ *
+ * given:
+ *
+ *	filename    - filename to parse
+ *
+ * If filename is NULL or the filename is not a readable file (or is empty) or
+ * if the file is not a C string or if read_all() returns NULL the function
+ * warns but does nothing else.
+ *
+ * NOTE: Until the JSON parser is written this only prints the string out.
+ */
+static void
+parse_file(char const *filename)
+{
+    size_t length = 0;		/* bytes read if file is a regular file and can be read */
+    size_t outputlen = 0;	/* number of bytes sent to stdout */
+    FILE *input_stream = NULL;	/* FILE * that refers to filename if filename is a regular readable file */
+    char *input = NULL;		/* data of read in file */
+    bool is_stdin = false;	/* true if reading from stdin (filename == "-") */
+    int ret;
+
+    /*
+     * firewall
+     */
+    if (filename == NULL) {
+	/* this should actually never happen */
+	warn(__func__, "passed NULL filename");
+	++num_errors;
+	return;
+    } else if (*filename == '\0') { /* strlen(filename) == 0 */
+	warn(__func__, "passed empty filename");
+	++num_errors;
+	return;
+    }
+
+    is_stdin = !strcmp(filename, "-");
+    if (!is_stdin && !is_file(filename)) {
+	warn(__func__, "passed filename that's not actually a file: %s", filename);
+	++num_errors;
+	return;
+    } else if (!is_stdin && !is_read(filename)) {
+	warn(__func__, "passed filename not readable file: %s", filename);
+	++num_errors;
+	return;
+    }
+
+    errno = 0;
+    input_stream = is_stdin ? stdin : fopen(filename, "r");
+    if (input_stream == NULL) {
+	warnp(__func__, "couldn't open file %s, ignoring", filename);
+	++num_errors;
+	return;
+    }
+
+    dbg(DBG_MED, "about to read from %s", is_stdin?"stdin":filename);
+    input = read_all(input_stream, &length);
+
+    if (input_stream == stdin) {
+	/*
+	 * If it's stdin we have to clear the EOF and error flags:
+	 * this is because if the user specifies - again they would
+	 * get an error like:
+	 *
+	 *	    Warning: read_all: EOF found at start of reading stream
+	 *	    FATAL[35]: main: error reading data in file -
+	 *
+	 * But that's an error itself.
+	 */
+	clearerr(stdin);
+    } else {
+	/*
+	 * If it's not stdin we close the stream as we no longer need
+	 * it: we either read in all the data or there was an error but
+	 * we close the file before we check for NULL input (meaning
+	 * read_all() failed).
+	 */
+	errno = 0;
+	ret = fclose(input_stream);
+	if (ret != 0) {
+	    warnp(__func__, "error in fclose on file %s", filename);
+	}
+	input_stream = NULL;
+    }
+
+    if (input == NULL) {
+	warn(__func__, "error while reading data from %s", is_stdin?"stdin":filename);
+	++num_errors;
+	return;
+    }
+    dbg(DBG_MED, "%s read length: %ju", is_stdin?"stdin":filename, (uintmax_t)length);
+
+    if (!is_string(input, length + 1)) {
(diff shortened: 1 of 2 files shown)


xexyl:
I'm taking a break now and I _might_ read a bit after that but I'm not sure of that. I might need another day off from that as yesterday was really hectic and the night was not easy either.

lcn2:
Ok, `-s` is strict mode whereas `-S` is refers to a string argument to parse, @xexyl?

xexyl:
> Ok, `-s` is strict mode whereas `-S` is refers to a string argument to parse, @xexyl?

Precisely.

xexyl:
But should the other tools be updated to have `-S` be strict so that this tool can use `-s` for strings? I don't mind.

lcn2:
> But should the other tools be updated to have `-S` be strict so that this tool can use `-s` for strings? I don't mind.

It might be a good idea to be consistent, yes.

xexyl:
> > But should the other tools be updated to have `-S` be strict so that this tool can use `-s` for strings? I don't mind.
> 
> It might be a good idea to be consistent, yes.

Well right now they are consistent but with `-s` for strict and `-S` for string. Should this be reversed though?

lcn2:
Avoiding the mistake of `scp` vs `ssh` ... which one uses `-p` for port and which one uses `-P` for port?  *sigh*

xexyl:
> Avoiding the mistake of `scp` vs `ssh` ... which one uses `-p` for port and which one uses `-P` for port? _sigh_

ssh uses `-p`. Or was that just expressing frustration at the differences?

xexyl:
Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.

lcn2:
> > > But should the other tools be updated to have `-S` be strict so that this tool can use `-s` for strings? I don't mind.
> > 
> > 
> > It might be a good idea to be consistent, yes.
> 
> Well right now they are consistent but with `-s` for strict and `-S` for string. Should this be reversed though?

Either way is fine.  Maybe pick the one that is more likely to be used (string mode?) for the lower case `-s` and the less likely to be used (strict mode?) for the UPPER CASE `-S`?

xexyl:
> > > > But should the other tools be updated to have `-S` be strict so that this tool can use `-s` for strings? I don't mind.
> > > 
> > > 
> > > It might be a good idea to be consistent, yes.
> > 
> > 
> > Well right now they are consistent but with `-s` for strict and `-S` for string. Should this be reversed though?
> 
> Either way is fine. Maybe pick the one that is more likely to be used (string mode?) for the lower case `-s` and the less likely to be used (strict mode?) for the UPPER CASE `-S`?

I was thinking the same thing actually. I'll try remembering to update that. Thanks.

lcn2:
> Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.

_sigh_

xexyl:
> > Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.
> 
> _sigh_

Well let's just be thankful `chown`, `rm`, `chmod` and `cp` (maybe others) all have consistent options for recursion. I think that would be even worse if they didn't.

lcn2:
> > > Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.
> > 
> > 
> > _sigh_
> 
> Well let's just be thankful `chown`, `rm`, `chmod` and `cp` (maybe others) all have consistent options for recursion. I think that would be even worse if they didn't.

Oh yes,. that is another POSIX thing .. the `-r` stuff and the mode names for `chmod` and how `chown` does dual ownership and group with the `:`, etc.

xexyl:
> > > > Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.
> > > 
> > > 
> > > _sigh_
> > 
> > 
> > Well let's just be thankful `chown`, `rm`, `chmod` and `cp` (maybe others) all have consistent options for recursion. I think that would be even worse if they didn't.
> 
> Oh yes,. that is another POSIX thing .. the `-r` stuff and the mode names for `chmod` and how `chown` does dual ownership and group with the `:`, etc.

Nice bit of trivia though perhaps I already knew it - not sure. Thanks!

lcn2:
> Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.

There is an **excuse** given to us as to why `scp` uses `-P` that does not come to mind because when it was stated, it was so ridiculous that it wasn't worth remembering.

The `-p` vs `-P` in `ssh` as if they made a nice looking cake, and they used salt instead of sugar because they both look similar, and then rationalized after the fact that the cake was intended to be made that way.

xexyl:
> > Because yes it's strange and I forgot about `scp` using `-P`. Seems silly that two tools in the same suite would use two different letters for the same purpose.
> 
> There is an **excuse** given to us as to why `scp` uses `-P` that does not come to mind because when it was stated, it was so ridiculous that it wasn't worth remembering.
> 

Well in that case it's probably right that you made it bold as to say it's ridiculous.

> The `-p` vs `-P` in `ssh` is if they made a nice looking cake, and they used salt instead of sugar because they both look similar, and then rationalized after the fact that the cake was intended to be made that way.

That makes sense yes. That is your description of it makes sense. And it's also funny and it reminds me of a funny (though not to her at the time) story with my auntie years ago at the house:

She was going to put salt on her steak and she inadvertently ended up putting sugar on it. I've mentioned this to her in more recent times and she laughs about it too.

PR #129, opened 2022-03-25
lexer grammar and other fixes and updates

xexyl:
Separate floating point and integers. This involves a new type in the
union and changing the regexps for the number types. There's a new
comment about the numbers and how to address the type given to simplify
the rules in the parser but this will be worked out later.

Fix the regexp for the floating point/exponent. At least some of it's
been fixed; there might be more cases I have missed but looking at
syntax highlighting in vim I have caught several invalid number strings.
Another minor adjustment there was from a misreading when adding
(somewhat hastily and possibly I was extra tired without realising it)
the original grammar. Specifically remove '""?' from the front. I'm not
sure if there was any harm in having it but it seems unnecessary anyway.

Add an int token_type in the parser for (for now at least) the
punctuation like '{', '}', '[', ']', ':', ',' etc. The assignment occurs
in the lexer actions and currently is not used. It might be useful to
change this to be a type of variable in the union (perhaps with changing
how it's assigned) perhaps in a struct but this will be figured out on a
later date.

Add commented out option in jparse.l that might be of use later but for
now is a problem ('%option nodefault').


xexyl:
This is probably all I'll get done today and I'm afraid I probably won't be reading any either. I did fix some of the files for compilation errors and warnings in the book's source code but I got less sleep last night and right now I don't have the energy or motivation to work on it more. This could possibly change but this is far from certain.

Tomorrow I should be able to do a bit before my cousin calls here but after that (as I said before) I won't be able to do anything.

xexyl:
Okay now I'm taking a break. No idea what I'll do here later on in the day.

xexyl:
...okay I didn't stop then but I am stopping now. These were minor things but I don't think I'll be doing anything else that takes thought if anything at all until tomorrow.

xexyl:
> > Would it be possible please for you to update the tool to act on *.y and *.l files as well ...
> 
> The `seqcexit` should the care what the filename is, so the `make seqcexit` could be extended to include the `bison` and `flex` input files.
> 
> Or are we missing something here?

Looking at the tool you have it set so it only updates C files.

lcn2:
> > > Would it be possible please for you to update the tool to act on *.y and *.l files as well ...
> > 
> > 
> > The `seqcexit` should the care what the filename is, so the `make seqcexit` could be extended to include the `bison` and `flex` input files.
> > Or are we missing something here?
> 
> Looking at the tool you have it set so it only updates C files.

Sorry (tm Canada :-)), fixed in the new release.

PR #130, opened 2022-03-25
Convert two errors in jparse.l to warnings

xexyl:
The errors in question are when the file couldn't be read in (i.e.
read_all() returns NULL) and if there are embedded NUL byte(s) in the
file.

I also increment num_errors in in these cases as well as some other
warnings that did not have this done already. This allows for the parser
to continue to the next file or string to actually make testing easier.

Sequenced exit codes (which did not end up needing to modify any other
files but it was done anyway).

### DIFF
--- a/jparse.c
+++ b/jparse.c
@@ -1925,15 +1925,18 @@ parse_json_string(char const *string)
     if (string == NULL) {
 	/* this should never happen */
 	warn(__func__, "passed NULL string");
+	++num_errors;
 	return;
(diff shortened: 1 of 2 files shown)


xexyl:
It might still be nice if `seqcexit` would look at `.l` and `.y` files though. I'm really going now.

Enjoy the rest of your day!

lcn2:
seqcexit v1.5 2022-03-25 will now process any filename.

It was a mistake to restrict to just *.c files, sorry (tm Canada :-)).

PR #137, opened 2022-03-31
Fix jfloat -t and jint -t (due to commit 2c47e4)

xexyl:
When running ./jint -t and ./jfloat -t all was okay and you'd get with
debug level DBG_LOW all tests PASSED. However the exit code was 13
because when sequencing exit codes in commit 2c47e4 the /*ooo*/ comment
was missing so the exit code was changed to (for both) 13.


xexyl:
Ah..you did indeed fix this so I'm closing this. Thank you for addressing it! And if you've been awake all night I hope you go get some sleep!

PR #138, opened 2022-03-31
make depend + typo fix in jfloat.h

xexyl:
Several updates had the need for make depend. For example updating
jfloat.h would not recompile jfloat.c. In particular the following were
generated:

    +utf8_test.o: utf8_test.c utf8_posix_map.h util.h dbg.h limit_ioccc.h \
    +  version.h
    +jint.o: jint.c jint.h dbg.h util.h json.h limit_ioccc.h version.h
    +jint.test.o: jint.test.c json.h
    +jfloat.o: jfloat.c jfloat.h dbg.h util.h json.h limit_ioccc.h version.h
    +jfloat.test.o: jfloat.test.c json.h

jfloat usage message referred to 'int' instead of 'float' (yank/paste
and forgetting to update the string?).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -555,6 +555,9 @@ utf8_posix_map.o: utf8_posix_map.c utf8_posix_map.h util.h dbg.h \
   limit_ioccc.h version.h
 sanity.o: sanity.c sanity.h util.h dbg.h location.h utf8_posix_map.h \
   json.h
-jparse.o: jparse.c util.h dbg.h jparse.tab.h
-jparse.tab.o: jparse.tab.c jparse.h dbg.h util.h json.h sanity.h \
-  location.h utf8_posix_map.h limit_ioccc.h version.h jparse.tab.h
+utf8_test.o: utf8_test.c utf8_posix_map.h util.h dbg.h limit_ioccc.h \
+  version.h
+jint.o: jint.c jint.h dbg.h util.h json.h limit_ioccc.h version.h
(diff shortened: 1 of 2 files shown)


xexyl:
And this will be all I'll do today.

lcn2:
Good catch @xexyl

xexyl:
> Good catch @xexyl

I am happy to help as always!

This is just another form of me seeing typos. More generally my eyes zoom in on wrong words and spelling errors (and also differences) and other such things. 

I was talking with my mum about this not too long ago and she told me that I have done this since I started to learn to read; it’s always been there. And as I said I have other gifts with reading as well. 

Funny to think about because my vision is truly awful (since birth) but as I have pointed out for years I can still see and that’s what matters. I would be devastated if I lost any of my senses but the inability to read I think would crush me more than other lost senses. Hopefully I never lose any (my hearing is also not that great and that’s also since birth but at least I can still hear). 

I notice details that other people simply miss (I already gave you some examples now that I think about it). I would be great at proofreading but I would not enjoy it if I couldn’t pick which books and texts to proofread so it would become a burden and then I wouldn’t do so well. 

Anyway as for continuing the work on the tools. I am hoping that I can do more next week. Monday and Tuesday I have some things going on but I will probably have a bit of time those days anyway. I think I just need a bit of time to figure out how I want to proceed. It’s quite possible that I do a little bit before then and it’s possible that I need more time but we shall see. 

Will be going to sleep soon. Have a great night!

PR #141, opened 2022-04-02
Minor clean up parser/lexer + make seqcexit

xexyl:
The functions clearerr_or_fclose() and print_newline() have been moved
to util.c so that the definitions are not needed in the lexer.

The print_newline() function now takes a bool output_newline. I'm aware
of another file that prints a newline based on a bool but that file does
not use this routine. Whether util.c is the best file I don't know and
I'm more than fine with it elsewhere.

Sequenced exit codes.

### DIFF
--- a/jint.c
+++ b/jint.c
@@ -398,7 +398,7 @@ main(int argc, char *argv[])
 	 */
 	ival = malloc_json_conv_int_str(input, &retlen);
 	if (ival == NULL) {
-	    err(13, __func__, "malloc_json_conv_int_str() is not supposed to return NULL!");
+	    err(12, __func__, "malloc_json_conv_int_str() is not supposed to return NULL!");
 	    not_reached();
 	}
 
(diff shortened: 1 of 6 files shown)


xexyl:
I did do a slight bit of reading and this is what made me think to do this. However as much as I want to read/work on this right now I think I'm going to end it for the day. I hope to do more tomorrow but this seemed like a worthy thing to get out of the way.

I'm pretty sure there are other places where code checks if file is `stdin` or not only to call `clearerr()` or `fclose()` and it could now use this function probably. An example where this might be used is `txzchk` but I'm not 100% sure of that.

I know one tool at least uses the `-n` option to prevent newlines being printed and that tool ould use `print_newline()` as well.

These can be done another time though.

xexyl:
Okay and now I'm really done...or at least I don't plan on doing any more today.

lcn2:
Thanks @xexyl and good day!

xexyl:
> Thanks @xexyl and good day!

To you too my friend!

PR #146, opened 2022-04-07
Remove artefact from shell_cmd/pipe_open + typos

xexyl:
It's not certain but I'm afraid this might be all I do today .. still trying to recover from the past few days (as of the day before yesterday).


xexyl:
Okay now I'm taking a break and I don't expect to return to this today .. was not a good night last night.

I've not had a chance to think much about the parser this past week but I do hope I can return to that within the next few days.

PR #148, opened 2022-04-08
Minor updates in json.h + new error functions

xexyl:


xexyl:
I believe that the new error functions should be discussed @lcn2 in the sense of error codes. This includes the (to be written) json counterparts.

Should these functions have error codes? If so should the `seqcexit` be modified or should a new system be devised for making the codes unique? Maybe something like the json error/warning codes? Or perhaps there shouldn't be error codes?

The `seqcexit` is for exit codes so maybe it shouldn't be modified? I don't know: I'm happy to give my thoughts later (not sure now) but ultimately I'll leave it up to you.

I give more details in the commit log as for the purpose of these functions as well as why they're not used yet (but several places in code come to mind immediately as to where this should be used).

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Sure. Any thoughts on how to make use of the error codes in these new functions? I put them there as I think it has some value but there still remains the question of how to define them and I would love to hear your thoughts on it. 

And with that I am putting the phone down. Good night my friend!

PR #151, opened 2022-04-12
Spelling fix in sorry.tm.ca.h + various updates

xexyl:
Specifically honor -> honour. There are two reasons but one was only
argued to plead my case of the other (yes this might be a bit
manipulative :( in the mind of some but my reasoning is below and I
think it's more than fair).

The first one is that I'm the one working on the parser and since I use
British English and it also goes against my honour to spell honour honor
:) and in fact the removal of 'u' in American English goes against my
very nature so I think it's fair to keep my honour by spelling it
honour not honor.

The second one is that honor is a US spelling so I argued that to truly
honour Canadians one should actually use their spelling! :)

Of course as a logophile and amateur philologist I will point out that
yes I do know that American English actually keeps some of the original
English more than British English (and maybe Canadian English as well:
not sure) but I'm not going to start a logomachy over a simple spelling
especially when I get my way. :)

Thank you Landon for not objecting to this change - I appreciate it a
lot!

### DIFF
--- a/bfok.sh
+++ b/bfok.sh
@@ -165,7 +165,7 @@ if [[ ! -f $VERGE ]]; then
     exit 3
 fi
 if [[ ! -x $VERGE ]]; then
-    echo "$0: ERROR: verge not a executable: $VERGE" 1>&2
+    echo "$0: ERROR: verge not an executable: $VERGE" 1>&2
     exit 3
 fi
 
--- a/jauthchk.c
+++ b/jauthchk.c
@@ -724,7 +724,7 @@ check_found_author_json_fields(char const *json_filename, bool test)
 	 * If the field is not allowed in the list then it suggests there is a
 	 * problem in the code because only author fields should be added to the
 	 * list in the first place. Thus it's an error if a field that's in the
-	 * author list is not a author field name.
+	 * author list is not an author field name.
 	 */
 	if (author_field == NULL) {
 	    jerr(JSON_CODE_RESERVED(10), NULL, __func__, __FILE__, NULL, __LINE__,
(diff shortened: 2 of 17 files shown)


xexyl:
This _might_ be all I do today: not sure. I'm actually wondering about waiting on the structs to be finished so I have something to work with. This would mean that when going through the book I also have something to work on. I think this might be the better way to go about it and it would also give me some time to do other things too though what those things are I don't know. Certainly some will be in this repo but I mean other things as well.

Do you have any thoughts on this?

xexyl:
Actually I'm not sure if commit 5c14ab5 is totally correct or complete and I meant to cancel it for that reason (i.e. didn't meant to push it but tried pulling from what was pushed) but unfortunately it was pushed instead. The reason I'm not sure it's correct/complete is after looking at the use of the function I see that other variables that are used in comparison with the return value of the `size_t` are `int`. I could change those but that involves more updates and I'm not up to doing that right now.

Even so it seems that returning the same type as the variable is a good idea (esp as it was not cast to int) - I'm just not sure if it's complete.

xexyl:
> This _might_ be all I do today: not sure.

As for this I think this is likely the case even though I'd like to do more ... just don't feel up to it. Maybe that'll change but very likely it will not. I would still like your thoughts on my question above though.

xexyl:
Okay now I really am done until tomorrow - though I'll probably reply to any comments you write. Have a good day!

lcn2:
busy with things .. thanks for the commit @xexyl

xexyl:
> busy with things .. thanks for the commit @xexyl

Take your time - hope all is okay!

PR #152, opened 2022-04-13
Various updates and fixes

xexyl:
I removed the `#define JSON_` in json.h and instead  we use the enum element_type with the prefix JTYPE_.

Additionally the JTYPE_ enum values are powers of two. This might not be
strictly necessary: the reason I used it is because currently the
checkers check for json numbers and not specific types so this way I
could do JTYPE_INT|JTYPE_FLOAT. Once a proper parser is complete this
might not be necessary at all - not sure yet how the checker functions
will work.

There still remains the issue that the parser has an enum with the
prefix JSON_ but I cannot control that and it's not yet worked out how
to integrate the two but using that enum is not yet worked out either as
nothing in the parser is worked out yet.

Running make prep passes all tests. Whether this change breaks some of
the working tests in test_JSON or not I do not know: I'm not too worried
about that because the parser isn't written yet and so what matters now
is that make test passes and it does: any files that are not made by
mkiocccentry can be tested later.

As for renaming the JTYPE_ to JSON_ this might be possible but it might
require changing the names of the symbols in jparse.y; I will worry
about that later - or not - but either way it's low priority right now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,23 @@ MANDIR = /usr/local/share/man/man1
 DESTDIR= /usr/local/bin
 TARGETS= mkiocccentry iocccsize dbg limit_ioccc.sh fnamchk txzchk jauthchk jinfochk \
 	jstrencode jstrdecode utf8_test jparse jint jfloat verge
-MANPAGES = mkiocccentry.1 txzchk.1 fnamchk.1 iocccsize.1 jinfochk.1 jauthchk.1
+
+# man pages
+#
+# Currently we explicitly specify a variable for man page targets as not all
+# targets have a man page but probably all targets should have a man page. When
+# this is done we can change the below two uncommented lines to be just:
+#
+#   MANPAGES = $(TARGETS:=.1)
+# 
+# But until then it must be the next two lines (alternatively we could
+# explicitly specify the man pages but this makes it simpler). When a new man
+# page is written the MAN_TARGETS should have the tool name (without any
+# extension) added to it.  Eventually MAN_TARGETS can be removed entirely and
+# MANPAGES will act on TARGETS.
+MAN_TARGETS = mkiocccentry txzchk fnamchk iocccsize jinfochk jauthchk
+MANPAGES= $(MAN_TARGETS:=.1)
+
 TEST_TARGETS= dbg utf8_test
 OBJFILES= dbg.o util.o mkiocccentry.o iocccsize.o fnamchk.o txzchk.o jauthchk.o jinfochk.o \
 	json.o jstrencode.o jstrdecode.o rule_count.o location.o sanity.o \
@@ -778,8 +794,8 @@ clobber: clean
 distclean nuke: clobber
 
 install: all
-	${INSTALL} -m 0555 ${TARGETS} ${DESTDIR}
-	${INSTALL} -m 0644 ${MANPAGES} ${MANDIR}
+	${INSTALL} -v -m 0555 ${TARGETS} ${DESTDIR}
+	${INSTALL} -v -m 0644 ${MANPAGES} ${MANDIR} 2>/dev/null
 
 tags: ${ALL_CSRC} ${H_FILES}
 	-${CTAGS} -- ${ALL_CSRC} ${H_FILES} 2>&1 | \
--- a/Makefile
+++ b/Makefile
@@ -175,12 +175,20 @@ GENERATED_HSRC= jparse.tab.h
 GENERATED_OBJ= jparse.o jparse.tab.o
 FLEXFILES= jparse.l
 BISONFILES= jparse.y
-SRCFILES= $(patsubst %.o,%.c,$(OBJFILES))
+# This is a simpler way to do:
+#
+#   SRCFILES =  $(patsubst %.o,%.c,$(OBJFILES))
+#
+SRCFILES= $(OBJFILES:.o=.c)
 ALL_CSRC= ${LESS_PICKY_CSRC} ${GENERATED_CSRC} ${SRCFILES}
 H_FILES= dbg.h jauthchk.h jinfochk.h json.h jstrdecode.h jstrencode.h limit_ioccc.h \
 	mkiocccentry.h txzchk.h util.h location.h utf8_posix_map.h jparse.h jint.h jfloat.h \
 	verge.h sorry.tm.ca.h
-DSYMDIRS= $(patsubst %,%.dSYM,$(TARGETS))
+# This is a simpler way to do:
+#
+#   DSYMDIRS= $(patsubst %,%.dSYM,$(TARGETS))
+#
+DSYMDIRS= $(TARGETS:=.dSYM)
 SH_FILES= iocccsize-test.sh jstr-test.sh limit_ioccc.sh mkiocccentry-test.sh json-test.sh \
 	  jcodechk.sh vermod.sh bfok.sh
 
--- a/iocccsize-test.sh
+++ b/iocccsize-test.sh
@@ -26,7 +26,7 @@ usage()
 	echo "usage: $0 [-h] [-b] [-v lvl] [-V] [-t tool]"
 	echo
 	echo "	-h	    print usage message and exit 2"
-	echo "	-b	    make disclean all (def: make all)"
+	echo "	-b	    make distclean all (def: make all)"
 	echo "	-v lvl	    set debugging level to lvl (def: no debugging)"
 	echo "	-V	    print tool version and exit 3"
 	echo "	-t tool	    test with tool (def: test with ./iocccsize)"
--- a/jauthchk.c
+++ b/jauthchk.c
@@ -444,10 +444,8 @@ check_author_json(char const *file, char const *fnamchk)
 		break;
 
 	    can_be_empty = (common_field && common_field->can_be_empty) || (author_field && author_field->can_be_empty);
-	    is_json_string = (common_field && (common_field->field_type == JSON_CHARS ||
-			      common_field->field_type == JSON_ARRAY_CHARS)) ||
-			     (author_field && (author_field->field_type == JSON_CHARS ||
-			      author_field->field_type == JSON_ARRAY_CHARS));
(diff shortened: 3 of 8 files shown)


xexyl:
If commit 132ed26 conflicts with the changes you're working on feel free to reject it: I only did it as something to do and I know it needs to be done. If something should be changed (it was quickly done and I don't know all of what you're doing) please also let me know.

I don't mind one way or the other. Hope you're having a nice sleep my friend!

xexyl:
BTW on the subject of the json numbers:

For the IOCCC this is probably unnecessary but for wider uses is it maybe an idea to use a bignum library (or even our own bignum code)? I ask this since json numbers can be much bigger than what C supports. 

I suppose this can come later even but just a thought. I've never worked with bignums but if you have maybe you'd like to work on that after the structs are done and ready for the parser? I know at least one or two previous winning entries have used them and I know one even made their own but I never have played with them. I probably _should_ but it's never been even halfway on my priority list: it's always been close to the bottom in fact and there are plenty of other things higher than it which I have yet to reach as well!

Just a thought that has crossed my mind a number of times and since I did make a change for numbers in this commit I figured I'd bring it up here. For future reference though: should I bring these types of things up in new issues or is the way I've done it so far good? I don't mind: whatever works better for you is better for me.

xexyl:
At the risk of stating the obvious :) if commit 132ed26 needs to be rejected then so too does ceb6db9.

xexyl:
FWIW commit 9a91bb3 is possibly the last one I'll do today. Not sure of that but it's quite possible as I am trying to avoid any conflicts with your imminent changes to `json.{c,h}` and I'm also kind of holding off for the parser until the structs are completed as it'll make things clearer and easier to work with.

xexyl:
As for commit c0fa0e1:

I'm not sure if this is correct actually: if it actually is a typo or not. The commit log:

```git
    Typo fixes in jstr-test.sh
    
    I saw this when the script was first introduced but I couldn't be sure
    if it was a typo or not. But since I can't figure out what 'foles' means
    I can only think it was supposed to be 'holes'. If this is not correct
    I"m happy to correct it to what it should be even if that's 'foles'.
```

This will be all I do today I believe. I've saved the errors in shellcheck under CentOS (other than the one I reported which I know I'm not supposed to touch) so I can look at them another time.

Have a good day!

xexyl:
Okay now I really am done. Again if the json changes conflict with what you're doing I can do those again later.

Good day!

xexyl:
FYI: My cousin will be here this Saturday so I won't be able to do much then and it might be for some days after that.

Anyway more from me tomorrow I'm sure.

lcn2:
We have a problem with the concept that JTYPE_ enum values are powers of two.

The union is setup hold one valid struct for an element at a time.  

We don't see why the _ JTYPE_INT|JTYPE_FLOAT_ overlap should be considered.  The fact that a floating point number can be an integer is prolly not important.  The application calling the parse understands the context of the JSON and can make a decision about how to process a value.

In the case of the JSON files `.info.json` and `.auth.json` we can safely reject as invalid a size of the form `123.0`.

xexyl:
> We have a problem with the concept that JTYPE_ enum values are powers of two.
> 
> The union is setup hold one valid struct for an element at a time.
> 
> We don't see why the _ JTYPE_INT|JTYPE_FLOAT_ overlap should be considered. The fact that a floating point number can be an integer is prolly not important. The application calling the parse understands the context of the JSON and can make a decision about how to process a value.
> 
> In the case of the JSON files `.info.json` and `.auth.json` we can safely reject as invalid a size of the form `123.0`.

Now this should be able to be fixed because we don't have any values that should be floating point. The idea was so that one could do `JTYPE_INT|JTYPE_FLOAT` but this is no longer even needed. I'll do a quick change.

lcn2:
> > We have a problem with the concept that JTYPE_ enum values are powers of two.
> > The union is setup hold one valid struct for an element at a time.
> > We don't see why the _ JTYPE_INT|JTYPE_FLOAT_ overlap should be considered. The fact that a floating point number can be an integer is prolly not important. The application calling the parse understands the context of the JSON and can make a decision about how to process a value.
> > In the case of the JSON files `.info.json` and `.auth.json` we can safely reject as invalid a size of the form `123.0`.
> 
> Now this should be able to be fixed because we don't have any values that should be floating point. The idea was so that one could do `JTYPE_INT|JTYPE_FLOAT` but this is no longer even needed. I'll do a quick change.

Thanks!

xexyl:
> > > We have a problem with the concept that JTYPE_ enum values are powers of two.
> > > The union is setup hold one valid struct for an element at a time.
> > > We don't see why the _ JTYPE_INT|JTYPE_FLOAT_ overlap should be considered. The fact that a floating point number can be an integer is prolly not important. The application calling the parse understands the context of the JSON and can make a decision about how to process a value.
> > > In the case of the JSON files `.info.json` and `.auth.json` we can safely reject as invalid a size of the form `123.0`.
> > 
> > 
> > Now this should be able to be fixed because we don't have any values that should be floating point. The idea was so that one could do `JTYPE_INT|JTYPE_FLOAT` but this is no longer even needed. I'll do a quick change.
> 
> Thanks!

Sure - and done. Took longer to come up with a good commit log than to make the change.

lcn2:
Thanks for your understanding @xexyl

xexyl:
> Thanks for your understanding @xexyl

Of course my friend! It's not even necessary now. It was in the original thought but it's not now.

Am I right in thinking that I should hold off on updating json.{c,h} ?

Also I wonder: should I add a comment in jparse.l and jparse.y in section three of the files so that it refers to the generated code above - since the code in section three I did write. Of course I'd have to come up with a good comment to have and it's probably not necessary but it did cross my mind.

xexyl:
> > Thanks for your understanding @xexyl
> 
> Of course my friend! It's not even necessary now. It was in the original thought but it's not now.

Naturally even if I felt it was necessary it would be fine. I do want to say the original purpose though. The reason it was this way is because the old incomplete parser had just `JSON_NUMBER` so I was trying to fit the old into the new more specific enum. This is because the tables had JSON_NUMBER so how to make sure all numbers are valid?

But as I noted in the comment the `is_number()` function (which is quite primitive) doesn't even work for floating point numbers so I removed that case statement. However if ever the tables do need (assuming the tables remain) need to have a floating point value it would be expected to be floating point so I could just use `JTYPE_FLOAT`.

Of course the fact it is an enum for the union means that of course it can't be powers of two: this was all me thinking early in the morning and all because I was trying to get rid of the defines but at the same time not break the current code so that `make test` will still work (as it does).

Thank you for pointing this out - it didn't even register until I stepped away a few minutes ago.

That'll be all for today as far as code goes. Anyway if you would let me know if I should not touch `json.{c,h}` until you say it's okay please let me know: don't want to stomp on your imminent changes.

Good day!

lcn2:
> > Thanks for your understanding @xexyl
> 
> Of course my friend! It's not even necessary now. It was in the original thought but it's not now.
> 
> Am I right in thinking that I should hold off on updating json.{c,h} ?

Yes, we have a change to `json.c` and `json.h` in the way ...

lcn2:
We have a change pending for `jfloat` to make is easy for the caller to determine if the floating point value is an integer.

lcn2:
> But as I noted in the comment the is_number() function (which is quite primitive) doesn't even work for floating point numbers so I removed that case statement. However if ever the tables do need (assuming the tables remain) need to have a floating point value it would be expected to be floating point so I could just use JTYPE_FLOAT.

We have a _workaround_ as the values needed for `jauthchk` and `jinfochk` need to be integers .. stay tuned!

xexyl:
> > But as I noted in the comment the is_number() function (which is quite primitive) doesn't even work for floating point numbers so I removed that case statement. However if ever the tables do need (assuming the tables remain) need to have a floating point value it would be expected to be floating point so I could just use JTYPE_FLOAT.
> 
> We have a _workaround_ as the values needed for `jauthchk` and `jinfochk` need to be integers .. stay tuned!

Sure though right now I'm not sure it's necessary. I guess I'll find out. A lot of uncertainty surrounds the parser and the tools!

Looking forward to seeing what you have in mind though.

xexyl:
> > > Thanks for your understanding @xexyl
> > 
> > 
> > Of course my friend! It's not even necessary now. It was in the original thought but it's not now.
> > Am I right in thinking that I should hold off on updating json.{c,h} ?
> 
> Yes, we have a change to `json.c` and `json.h` in the way ...

Thought so. Thanks for confirming! I think many changes even.

xexyl:
> We have a change pending for `jfloat` to make is easy for the caller to determine if the floating point value is an integer.

That sounds really useful too - thanks!

xexyl:
> With commit [fbafe5b](https://github.com/ioccc-src/mkiocccentry/commit/fbafe5bb548aa884495f6471d3bf48a9b3e5636e) the changes have been made:
> 
> ```
> Add JSON string processing, is_number fix, added float as int info
> 
> Previous uses of is_number() needed only integers.
> The old is_number() function is now is_integer().
> 
> Added boolean values to struct floating to indicate that
> the floating point value is an integer.
> 
> Updated jfloat.test and jfloat to test non-integer
> as well as integer values.
> 
> Fixed JSON processing cases where the malloc was not the correct size.
> 
> Added code to convert JSON encoded string to C string
> via malloc_json_conv_string() and malloc_json_conv_string_str().
> 
> Added posix_safe_chk() function to util.c to perform a
> more extensive check on the JSON decoded string value.
> ```

Thanks!

Funny thing: I almost renamed the function `is_integer()` today when pointing out the fact it doesn't deal with floating point.

I'll look at the changes later on probably not until tomorrow. I guess you'll be doing the other structs as well in the coming days?

lcn2:
> 'll look at the changes later on probably not until tomorrow. I guess you'll be doing the other structs as well in the coming days?

yes 👍

xexyl:
> > 'll look at the changes later on probably not until tomorrow. I guess you'll be doing the other structs as well in the coming days?
> 
> yes 👍

Great. Look forward to it!

PR #155, opened 2022-04-14
Fix compilation under CentOS 7 + English updates

xexyl:
The function floorl() requires -lm under CentOS 7 (the proper feature
test macros were already enabled although I have also changed
_XOPEN_SOURCE to be 600 up from 500: not sure if this is strictly
necessary but it doesn't hurt to have it at 600).

The Makefile has had some changes in the way warnings are handled
described below as thus:

First the warning flags -Wall -Wextra -Werror have been moved to the new
variable WARN_FLAGS. After this in WARN_FLAGS specific warnings have
been disabled in particular:

     -Wno-unused-command-line-argument -Wno-poison-system-directories -Wno-unreachable-code-break -Wno-padded

The first one is important for macOS as after adding -lm to the CFLAGS
(via the standard LDFLAGS variable) we get that error for the fewer
objects that don't use -lm. Since many of the objects do link in json.o
(which is what needs -lm) we simply add disable the warning in the
WARN_FLAGS variable. There is at least one warning that we explicitly
disable in some rules as they're not needed in many: for example the
parser needs some warnings disabled.

As for the others they're enabled by -Weverything and although we don't
enable that they're here in case we ever do as they're irrelevant here.

With this commit everything should compile under macOS and also CentOS 7
(and other linux systems should be fine as well: CentOS is in a sense
behind as it's built for stability). In 2024 CentOS 7 will reach EOL and
some of the things we include for CentOS 7 should thus no longer be
needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,17 @@ TRUE= true
 # Makefile parameters #
 #######################
 
+# linker options
+#
+# We need the following linker options for some systems:
+#
+#   -lm for floorl() under CentOS 7.
+#
+# If any more linker options are needed we will try documenting the reason why
+# here.
+#
+LDFLAGS = -lm
+
 # C source standards being used
 #
 # NOTE: The use of -std=gnu11 is because there are a few older systems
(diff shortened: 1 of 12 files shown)


xexyl:
As for commit f8fa3fd:

This commit would resolve the first thing I suggested yesterday prior to
release namely the 1st -> first, 2nd -> second and so on but if this
conflicts with the pending json.c changes this commit should be rejected
and I can do redo it without any issues on my end; I don't mind either
way but wanted to point it out just in case.

xexyl:
I suspect that the above two commits is all I'll do today. Not certain of that but it's quite possible. I don't feel up to much of anything right now but I wanted to get the compilation errors resolved and since I had the English update in `verge.c` I figured I'd finish the job. As the log for that commit says if it conflicts with json.c please reject that commit and I'll do it again on all the files later. 

I've saved the commit log in a text file locally so I don't have to rewrite it in that case.

lcn2:
Offsite so we will delay the pull request until we are at a place where we have code.

xexyl:
> Offsite so we will delay the pull request until we are at a place where we have code.

No worries! Hope all is well. I'll be heading off shortly. Enjoy the rest of your day and I hope nothing is wrong!

PR #157, opened 2022-04-15
Typo fixes, Makefile fixes and other updates

xexyl:
The most important changes are the Makefile fixes. Specifically if one were to update `jint.h` or `jfloat.h` the respective source file would not be recompiled.

Now as for the English updates there is something I will bring up in another comment.


xexyl:
As for English updates. If you note from commit b0d2910 there are some places where I kept the digit(s). This is because shell `$#` shows digits also.

But this raises a possibly valid question: since in some cases it has to remain as digits (to remain consistent) should in messages all others be digits too? In my mind they should be words if the number is < 10 but this might be up to debate for one reason: is it truly consistent? Each sentence is consistent but not all sentences are consistent wrt each other. I don't know. For me I would prefer the writing it out because it looks cleaner and more correct but I figured I'd bring it up anyway.

Either way for args/arguments each error message has a consistent form.

I don't expect to do more than this today but I'm not certain of this either: I don't plan to do more but as has happened many times even after I plan to not do more I end up doing more - sometimes quite a bit more. 

Hope you're having a nice sleep!

xexyl:
As for commit fe5dc253 it's for a couple reasons.

First of all so that there's no spelling error triggered. Second is I think it might be clearer as major.minor.patch is a standard form. 

I had thought of adding a regex but I felt that that would be more complicated and some might not even understand it. There were other thoughts that crossed my mind as possibilities but they all seemed to have some ambiguity. In fact the way I have it now has a source of confusion as well:

Should  `-[12]` be typed explicitly? The idea was that it is the first version and second version so 1 and 2. On the other hand since `-[0-9]+` is also used in versioning it might be confusing in that way too.

I don't know but I think this way might be better. I did think of one other form though:

```
1.0.0 2.0.0
```

but then would people take that literally? I think some might (and I would probably be in that group though with my understanding of commands I would probably realise that it's meant to be two numbers in that form).

Some things to think about anyway.

As for this commit I now believe I'm done for the day. Soon I will be afk for a while and I have other things to do when I get back.

lcn2:
Thanks for the update @xexyl

xexyl:
> Thanks for the update @xexyl

Happy to help!

PR #158, opened 2022-04-16
JSON parser + lexer updates and mkiocccentry fixes

xexyl:
The JTYPE_ enum in jparse.h have been renamed with the prefix JSON_ and
the JSON_ in the lexer and parser have been prefixed instead with JTYPE_
as nobody in their right mind will look at the parser or scanner code
but people will look at the other code. Actually nobody in their left
(wrong) mind would look at the generated code either because anyone who
does will have horrible nightmares about spaghetti code with very long
tentacles attached to the pasta that tries to grab them and pull them
into its mouth before consuming them with insatiable craving for humans
in its eyes. Okay you get the point: JTYPE_ is a better name for the
ugly, offensive code that can't be maintained whereas JSON_ is better
for code that can be maintained and is not ugly or offensive.

In the %union block in the parser change struct json *json to struct
json. Also removed the other members as the struct json has a union
itself. Functions that have yet to be written will be used for this (in
the parser and the lexer).

In the lexer I've naturally removed the assignment of
yylval.union_member as those members no longer exist. I still do print
what it is and I return the now JTYPE_ prefixed types but that's all.
Well for some the token_type int (that probably will be removed at some
point) is also used for the punctuation tokens but that hasn't changed.

### DIFF
--- a/jauthchk.c
+++ b/jauthchk.c
@@ -444,8 +444,8 @@ check_author_json(char const *file, char const *fnamchk)
 		break;
 
 	    can_be_empty = (common_field && common_field->can_be_empty) || (author_field && author_field->can_be_empty);
-	    is_json_string = (common_field && common_field->field_type == JTYPE_STRING) ||
-			     (author_field && (author_field->field_type == JTYPE_STRING));
+	    is_json_string = (common_field && common_field->field_type == JSON_STRING) ||
+			     (author_field && (author_field->field_type == JSON_STRING));
 	    /*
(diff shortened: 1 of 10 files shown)


xexyl:
> I removed have had a use but it shouldn't have been reached at all.

\- have

This was from changing the sentence mid sentence and forgetting to update the rest of it.

lcn2:
🙏 @xexyl

xexyl:
> 🙏 @xexyl

Welcome. I have another change incoming...

PR #161, opened 2022-04-17
Spelling updates and fix commit 463ec301

xexyl:
For non-standard thru -> through:

As part of making this repo ready for public consumption (even though
the code is not ready for that :) ) I've made this change. Besides that
it triggers vim spelllocal highlight which shouldn't happen except
perhaps for American English[0] versus British English[0] (and other
Englishes[1] I guess): whereas the Judges and others use American
English I use British English but they keep my spelling and I do
likewise with theirs.

[0] Well OK 'should' is actually subjective: I would prefer it all be
British English but it is an international contest so we should all be
fair to everyone else with languages ... unless of course someone tries
putting in other computer languages (especially that ghastly C++ or any
references to it :) ).

[1] The astute reader who wants to be (or if they're like me: cannot
help but be :) ) cynical might point out that English is a mass noun so
the plural is the same: Englishes is with this logic incorrect and in
fact it does trigger spelllocal. However it's amusing to me to complain
about spelllocal being triggered and at the same time trigger it when
explaining an exception to the rule.

But there's also a point to make to refute this logic: a mass
noun can also be something that normally is not countable but can be
when referring to different units or types and I'll argue that British
versus American English are two types of the same (though normally I'd
argue that in many respects they're actually two different languages).

Anyway the point is that non-standard spellings should be avoided and
with the exception of computer terms (that might trigger spelllocal in
vim), punning (which I'm doing with Englishes here) and American versus
British or some other English spelllocal should also be avoided.

### DIFF
--- a/dyn_alloc_test.h
+++ b/dyn_alloc_test.h
@@ -60,7 +60,7 @@
 /*
  * usage message
  *
- * Use the usage() function to print the these usage_msgX strings.
+ * Use the usage() function to print the these usage_msg([0-9]?)+ strings.
  */
 static const char * const usage_msg =
     "usage: %s [-h] [-v level] [-V]\n"
--- a/fnamchk.h
+++ b/fnamchk.h
@@ -66,7 +66,7 @@
 /*
  * usage message
  *
- * Use the usage() function to print the these usage_msgX strings.
+ * Use the usage() function to print the these usage_msg([0-9]?)+ strings.
  */
 static const char * const usage_msg =
     "usage: %s [-h] [-v level] [-q] [-V] [-E ext] [-t|-u] filepath\n"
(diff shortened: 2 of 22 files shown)


lcn2:
Thanks @xexyl no issues with this pull (so far).

xexyl:
> Thanks @xexyl no issues with this pull (so far).

I don't see any problem either. Except that:

Something strange with my fork today... having problems with it wanting to merge and having (earlier) conflicts which made me delete the fork and re-fork. I think that might have been a mistake as other problems have happened since then. 

Think tomorrow when I pull out the backup drive I'll open up Time Machine and copy my fork from yesterday and then do a pull and hopefully I can fix it that way - though perhaps not because of the deletion of the fork. 

Going to watch a film - or rather continue a film. Good day! If nothing else I'll look at comments tomorrow.

PR #162, opened 2022-04-17
Fix commit b0813e09

xexyl:
The change was not done in jfloat.h or jint.h and I'm not sure why as I
    was pretty sure I used sed -i on *.h. However I've been having some
    issues with my fork today and when making this commit I noticed some
    additional things that are not right as well. Anyway this commit changes
    usage_msgX to be usage_msg([0-9]?)+ for the reasons I cited in the other
    commit.

The commits that are part of this pull request show some of the problems I'm having with my fork. It's incredibly annoying and I'll try working this out tomorrow but I wanted the change in all the files anyway.

### DIFF
--- a/jfloat.h
+++ b/jfloat.h
@@ -61,7 +61,7 @@
 /*
  * usage message
  *
- * Use the usage() function to print the these usage_msgX strings.
+ * Use the usage() function to print the these usage_msg([0-9]?)+ strings.
  */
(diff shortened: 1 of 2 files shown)


xexyl:
Actually I'm closing this. I'll try recreating the fork tomorrow and see if I can figure out what is going on. Then I'll make this change in the final files. I don't think this should be merged.

PR #163, opened 2022-04-18
JSON parser and scanner updates + other updates

xexyl:
# usage_msgX -> usage_msg([0-9]?)+ in j(int|float).h

This was for some reason not done yesterday. I'm not sure how as I am
sure I used sed -i on *.[ch] but I was having some issues with my fork
(which I hope have been resolved now). As I said in commit b0813e094:

This is strictly because usage_msgX triggers vim spellbad and the regex
explains more clearly what it is ... possibly even to those who don't
know basic regex since they can take a look at some of the files that do
have more than one usage_msg e.g. mkiocccentry's usage messages.


xexyl:
The JSON parser and scanner updates commit is likely the last one for the day but we shall see.

lcn2:
Good improvements @xexyl, hope you rest well.

xexyl:
> Good improvements @xexyl, hope you rest well.

Thank you - for both! Have a great day! I did think of a slight enhancement for the `find_utils()` function I added a while back but it's not that important so if I do it I'll do it during an off moment. Won't happen today either way.

PR #164, opened 2022-04-19
Minor update to Makefile

xexyl:
For the rules used by make prep that require certain tools always use
the form (when the tool is not found) of:

	    echo 'The seqcexit tool could not be found.' 1>&2; \
	    echo 'The seqcexit tool is required for this rule.'; 1>&2; \

Before this commit for seqcexit it said:

	    echo 'If you have not installed the seqcexit tool, then' 1>&2; \
	    echo 'you may not run this rule.'; 1>&2; \

which could be construed as somewhat negative. This same pattern was
used in the shellcheck command. For shellcheck I also suggested that one
could use their OS package manager to install it (after giving the
GitHub repo link).

I've not touched the reset_min_timestamp rule with this wording for the
rpl tool as this rule was made by the judges and only for the judges. I
did however make a typo fix in the comments above this rule (the ALL the
-> ALL the).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -503,7 +503,8 @@ jparse.c: jparse.l jparse.tab.h bfok.sh limit_ioccc.sh verge jparse.ref.c Makefi
 # repo tools - rules for those who maintain the mkiocccentry repo #
 ###################################################################
 
-# things to do before a release, forming a pull, and/or updating the GitHub repo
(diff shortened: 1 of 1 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

To you as well.

PR #168, opened 2022-04-21
Major (mostly JSON) reorganisation, grammar updates

xexyl:
Move JSON parser specific structs and functions from json.{h,c} to
jparsefuncs.{h,c} with the exception of struct json, its union (and the
related structs) and code that acts on these. Perhaps these should also
be moved over and the files renamed jparseutil (or kept as jparsefuncs
which looks nicer to my eyes though possibly it's not as accurately
named: not sure) or something like that?

Move the functions in mkiocccentry.{h,c} that write a json file to
json.{h,c}. Perhaps these could be in jutil.{h,c} (which was the
original name thought up for jparsefuncs.{h,c}) along with the
encode/decode functions: not sure.

Moved lookup_location_name() to location.c where it belongs. This is the
only change not specific to JSON.

Updated Makefile (including make depend) for these changes.

I do want to point out some intentional irony in the comments at the top
of jparsefuncs.{h,c}. It has the quote that I added for txzchk:

    + * "Because sometimes even the IOCCC Judges need some help." :-)

but below that I put:

    + * This is currently being worked on by:
    + *
    + *	@xexyl
    + *	https://xexyl.net		Cody Boone Ferguson
    + *	https://ioccc.xexyl.net
    + *
    + * and
    + *
    + *	chongo (Landon Curt Noll, http://www.isthe.com/chongo/index.html) /\oo/\

when of course Landon is one of the judges ... which means that the judges
are getting help and they're also helping themselves by helping someone
help them! :)

TODO:

* Decide whether or not to move the struct json and the related structs
and functions to jparsefuncs. This might seem like the obvious choice
but the discussion did not originally suggest this so I'm keeping it in
json.{h,c} for now.

* Decide if anything else should be moved over.

* Decide if jparsefuncs is the best name for the above mentioned
changes.


xexyl:
Actually after a regeneration of files (which I had done already) the regex now does trigger an error so the commit https://github.com/ioccc-src/mkiocccentry/pull/168/commits/eb3a47c5bffbbfdc90ec2f8298850a44e764ca33 should be rejected. Possibly I'll roll it back instead: not sure yet.

xexyl:
On the subject of grammar: I saved the links to the comments you made yesterday about this so I can look at them quickly.

As for what else I'll do today here: probably nothing. Right before I went to lie down I started feeling quite down (and certainly off) though I can't figure out why. Thus I'm just going to take it easy. I'll probably reply to comments though.

Hope you're having a good morning!

lcn2:
Thanks @xexyl

lcn2:
Again, this is a good start at a separation between functions supporting the JSON parse tree functions and more general JSON utility functions.

We will look at this is more detail and consider, if needed, a few more moves / changes.

xexyl:
> Thanks @xexyl

Sure.

xexyl:
> Again, this is a good start at a separation between functions supporting the JSON parse tree functions and more general JSON utility functions.
> 
> We will look at this is more detail and consider, if needed, a few more moves / changes.

Yes. The original file is still very long. I think over 3k lines? As I noted it might be an idea to move the `struct json` to these new files? What do you think about that? Of course if you agree with it feel free to do it. Certainly more separation could and probably should be done. The same might be the case for some of the other files as well though maybe not. 

Tomorrow I will look at those functions we discussed and if it all is clear to me (I was in the midst of something during the dialogue so it’s possible that I missed something in the back and forth) I will look into writing it (assuming I don’t work on something else - there’s a lot to do still of course). I also will reread the comments about the grammar. Definitely will change the prefix of the tokens as well. 

Those are the things I am thinking about for tomorrow but as always it depends on a number of factors. 

I hope you have a great night! I will be awake a little longer and then I will go to sleep. Good night my friend!

PR #169, opened 2022-04-22
txzchk fixes + several JSON updates

xexyl:
I'd like in particular to point out commit 1f4dd7687a69fc:

```
    Older bison and POSIX Yacc compatibility
    
    Change parse.error 'detailed' to 'verbose' to support older versions of
    bison. Now bison version 3.0.4 will work which means CentOS 7 bison can
    actually generate the file. The change in verbosity does not seem to
    make a difference for me so this is a worthwhile change. Please note the
    version has not yet been updated in version.h.
    
    Now passes -Wyacc which has been added to the Makefile. This is done by
    making a couple changes described below.
    
    POSIX Yacc does not support the %empty directive so we use /* empty */
    instead.
    
    POSIX Yacc does not support the %define directive but it does support
    the -D option. Thus instead of having in jparse.y (where detailed is now
    verbose as described above):
    
        %define parse.error detailed
        %define parse.lac full
    
    we have in the Makefile:
    
        echo "$$BISON_PATH -d -Wyacc -Dparse.error=verbose -Dparse.lac=full jparse.y"; \
        "$$BISON_PATH" -d -Wyacc -Dparse.error=verbose -Dparse.lac=full jparse.y; \

```

I did try updating the version in `version.h` but for some reason running `make` did not let CentOS bison work though using `./bfok.sh` directly reported the correct exit code (I think I did that). Would you like to change this? I'm not sure if older versions of flex will work but I'll test that: perhaps in a bit as I need to go afk for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,7 +323,7 @@ dbg: dbg.c Makefile
 fnamchk: fnamchk.c dbg.o util.o dyn_array.o Makefile
 	${CC} ${CFLAGS} fnamchk.c dbg.o util.o dyn_array.o -o $@
 
-txzchk: txzchk.c rule_count.o dbg.o util.o dyn_array.o location.o json.o json_chk.o json_util.o \
+txzchk: txzchk.c txzchk.h rule_count.o dbg.o util.o dyn_array.o location.o json.o json_chk.o json_util.o \
 	utf8_posix_map.o sanity.o Makefile
 	${CC} ${CFLAGS} txzchk.c rule_count.o dbg.o util.o dyn_array.o location.o json.o \
 	    json_chk.o json_util.o utf8_posix_map.o sanity.o -o $@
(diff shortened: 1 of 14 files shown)


xexyl:
I think I found the problem with CentOS bison version not being found and if so it would be a problem for those who don't have alternative paths either. Will make the change and test this out. If so I'll commit but then I have to go.

lcn2:
FYI: We are in the process of improving the way `bison` and `flex` are used, replacing the `bfok.sh` script with 2 new scripts and simplifying the make files for `jparse.tab.c`, `jparse.tab.h`, and `jparse.c`.

Thanks for this pull request, @xexyl.  We hope to further improve on it.

xexyl:
> FYI: We are in the process of improving the way `bison` and `flex` are used, replacing the `bfok.sh` script with 2 new scripts and simplifying the make files for `jparse.tab.c`, `jparse.tab.h`, and `jparse.c`.
> 

Sounds good.

> Thanks for this pull request, @xexyl. We hope to further improve on it.

Happy to help and look forward to seeing what you have in mind!

xexyl:
BTW just noticed an interesting thing with the Makefile:

```shell
$ make parser
rm -f jparse.tab.c jparse.tab.h
make jparse.tab.c jparse.tab.h
rm -f jparse.tab.c jparse.tab.h
...
```

does the same thing twice. Since you'll be updating the Makefile maybe it'll be fixed anyway but in case you hadn't noticed.

PR #178, opened 2022-04-26
JSON parser updates

xexyl:
As for the first commit where the lexer and parser ironically declare that they're ugly:

**XXX Unfortunately there are some commits from yesterday that have already been merged but for some reason I had to do a merge on my end as well so you might want to NOT merge the commit that remove the `-T` from usage message in `jparse` if you can (maybe it's not a problem though?). I'm not sure if not merging the merge branch one will be a problem or not. If I have this problem again I'll once again re-fork the repo which I almost did but was too tired to bother with it.** 

Anyway this is the commit log:

    This is done by changing the (yy|YY) prefix to (ugly|UGLY) in the lexer
    and parser. As the following comment says:
    
        /*
         * As we utterly object to the hideous code that bison and flex generate we
         * point it out in an ironic way by changing the prefix yy to ugly_.
         */
    
    This is done in flex and bison via:
    
        %option prefix="ugly_"
        %define api.prefix {ugly_}
    
    respectively. This creates an ironic statement in the lexer:
    
        #define ugly_lex_ALREADY_DEFINED
    
    and an ironic statement in the C pre-processor in the parser:
    
        #if ! defined UGLY_STYPE && ! defined UGLY_STYPE_IS_DECLARED
        # define UGLY_STYPE_IS_DECLARED 1
    
    which is to say that both the lexer and parser actually declare that
    they're ugly! :))
    
    For the curious:
    
        $ grep -ic ugly jparse*ref.[ch]
        jparse.ref.c:108
        jparse.tab.ref.c:34
        jparse.tab.ref.h:24
    
    which is a huge amount of ugliness (though it's much more ugly than what
    it admits :) )!
    
    As well I've made a minor change to sorry.tm.ca.h that's not even worth
    describing.


xexyl:
That last commit is almost certainly the last of the day. I've had my fun and I hope you appreciate it. I am going to do something else now. Good day!

xexyl:
Oh totally OT I found a useful link about vim that you might like:

https://vim.fandom.com/wiki/Best_Vim_Tips

Has some easter eggs too. I know of another one but I don't think I have it configured on this machine and the fedora box where it is is down.

Anyway good day!

lcn2:
Thanks 🙏 @xexyl

xexyl:
> Thanks 🙏 @xexyl

Very welcome my friend! Glad you appreciated me taking the piss out of flex and bison as well.

lcn2:
> > Thanks 🙏 @xexyl
> 
> Very welcome my friend! Glad you appreciated me taking the piss out of flex and bison as well.

It was fun. :)

xexyl:
> > > Thanks 🙏 @xexyl
> > 
> > 
> > Very welcome my friend! Glad you appreciated me taking the piss out of flex and bison as well.
> 
> It was fun. :)

Happy you agree! It IS ugly after all and the best way to deal with something not pleasant is humour (or at least it's one of the best ways?). :)

xexyl:
I will be going to sleep now but I wanted to wish you a good night before I turn the lights off. 

As well I wanted to say that I will update the comments tomorrow about how it appears that the prefix might suggest that our struct json is ugly. I will work it out so that it works well and then remove the comment about how it might be a problem. I do generally speaking have a way with words so should be fine. 

If you give an update on the struct number in the other thread that would be good but if not I will read the log - this is of course assuming you actually finish it today. 

And with that good night and sleep tight! More from me tomorrow.

lcn2:
To prepare for the merged `struct json_number` we needed to clean out old references to the previous JSON numerics.
This required us to modify the `jparse.y` and `jparse.l` code.   While we were thinking about how the `json_conv_number()` code will work, and we were inside the `bison` and `flex` code, we modified the regex to account the JSON numbers according to the [JSON spec](https://www.json.org/json-en.html).

BTW: While we in that same parser code, made a slight improvement to JSON string regex in an attempt to handle the \" case.

We hope we didn't mess anything up.

With commit 4b883153d851e3ff03c7d18465e669dbe75c0a6e

    Checkpoint on merged struct json_number

    The jint and jfloat code is no longer compiled.
    The jint_gen and jfloat_gen code is no longer compiled.

    Added stub rules for jnumber and rebuild_jnumber_test.

    The make clean and make clobber rules remove, for now,
    compiled jint and jfloat and jint_gen and jfloat_gen files.

    Removed all use of JSON_INTMAX, JSON_UINTMAX and JSON_LONG_DOUBLE.

    Added JSON_NUMBER trial (untested) regex.

    Attempted to improve parser regex for JSON numeric values to handle
    the full JSON number specification.

    While we were in the parser code, made an improvement on the regex
    for JSON strings in an attempt to handle the \" case.

    Replaced use of JTYPE_INT and JTYPE_FLOAT with JTYPE_NUMBER.

    Added stub code for json_conv_number() and json_conv_number_str().

    Replaced use pf ./jint and ./jfloat with ./jnumber in test.sh.

xexyl:
That all sounds great! I will be taking a look shortly. 

I also like the idea that the clobber rule still removes the previous objects and binaries as that will help remove some clutter. It might be an idea for the future to keep doing that and maybe until a proper release? Maybe with a different rule? I don’t know: I just know that I have more than once found binaries (or if we want to be funny we can say binary files that are also text files though I guess not everyone will understand that but I know you will) that no longer are removed and I have to think for a moment why that is. 

Well will let you know how the regexes work. Hope you’re having a great sleep my friend!

xexyl:
> To prepare for the merged `struct json_number` we needed to clean out old references to the previous JSON numerics. This required us to modify the `jparse.y` and `jparse.l` code.
> 
> While we were thinking about how the `json_conv_number()` code will work, and we were inside the `bison` and `flex` code, we modified the regex to better account the JSON numbers according to the [JSON spec](https://www.json.org/json-en.html). We needed to complete the transition from previous JSON numerics to the single JSON number before we started on the interior of the conversion code for the merged `struct json_number`.
> 
> We could have built one single huge `JSON_NUMBER` regex, but instead we opted to built it up in 3 stages. This is similar to the way the _railroad track_ grammar in [JSON spec](https://www.json.org/json-en.html) defines JSON _number_. We even named the parts in accordance with the comments found in the _railroad track_ grammar in [JSON spec](https://www.json.org/json-en.html) defines JSON _number_.

Mind elaborating what you mean by railway track grammar? That sounds good though to have it the way they do even if it might be - well you know. Even so I think it's better to have it broken down for easier reading.

However I did notice that the regex doesn't work for at least one case (I think). If I'm not mistaken `+` is not allowed in front of numbers in json? On the other hand though we have the conversion routines which do check for that so I guess that's okay.

> 
> > BTW: While we in that same parser code, made a slight improvement to JSON string regex in an attempt to handle the` \"` case.   It seems that if the `JSON_STRING` regex is able to grab all valid JSON quoted strings, including JSON strings with backslash escaped quotes (i.e., ` \"`) then that is all the `JSON_STRING` regex really needs to do.  The modified `JSON_STRING` regex seems good enough to pass to the JSON string conversion functions `(json_conv_string()` and `json_conv_string_str()`) and let them to their detailed work.
> 

It looks good to me! Check this out:

```shell
$ ./jparse -s '{"string": "this\"is a test" }'
[...]
string: '"string"'
[...]
string: '"this\"is a test"'
```

So thank you!

> We hope we didn't mess anything up in the parser. It builds and compiles, which was better that it might have been had we left the previous JSON numeric code in place, referring to now non-existent symbols and non-existent structures that would have prevented things from compiling.
> 

It looks like everything is in order in that way yes.

I already replied to the commit parts I think that are useful so no need to repeat it.

lcn2:
> Mind elaborating what you mean by railway track grammar?

We refer to the lines filling most of the page on the left side of the [JSON spec](https://www.json.org/json-en.html).

xexyl:
> > Mind elaborating what you mean by railway track grammar?
> 
> We refer to the lines filling most of the page on the left side of the [JSON spec](https://www.json.org/json-en.html).

Ah I see. Thanks.

xexyl:
And I just noticed why the `+` is accepted... It's not in the number itself. Perhaps it needs to be removed or addressed? It looks like for `./jparse -s '+555'` the number is '555' but the '+' is not there. So this is an issue that must be resolved.

xexyl:
The question is why it's not triggering an error since it's not in the grammar. Perhaps a new token type is needed - one for invalid chars? But still seems like it should already trigger a problem.

lcn2:
> However I did notice that the regex doesn't work for at least one case (I think). If I'm not mistaken + is not allowed in front of numbers in json? On the other hand though we have the conversion routines which do check for that so I guess that's okay.

We don't think numbers such as `+1` will match the `JSON_NUMBER` regex.  The [JSON spec](https://www.json.org/json-en.html) does allow for + after the `e` or `E` (so much for the JSON designers minimalist goals) as does the `JSON_NUMBER` regex.

So we don't see the issue you raised with the `JSON_NUMBER` regex.  Still, the conversion routines do rest to the check as you said.

xexyl:
> > However I did notice that the regex doesn't work for at least one case (I think). If I'm not mistaken + is not allowed in front of numbers in json? On the other hand though we have the conversion routines which do check for that so I guess that's okay.
> 
> We don't think numbers such as `+1` will match the `JSON_NUMBER` regex. The [JSON spec](https://www.json.org/json-en.html) does allow for + after the `e` or `E` (so much for the JSON designers minimalist goals) as does the `JSON_NUMBER` regex.
> 

You're right: it doesn't. I added a comment a few minutes ago about this.

> So we don't see the issue you raised with the `JSON_NUMBER` regex. Still, the conversion routines do rest to the check as you said.

Unfortunately though this will not be picked up since the regex (as you said) does not match the leading `+`.

xexyl:
The question is: what should happen on chars that aren't valid? Should the lexer return a value for invalid input? I'm actually surprised that the `+` goes through without a problem as it's not a match and the parser rules should forbid it.

xexyl:
Okay it is the default rule. But why doesn't the parser complain?

Anyway here's what I mean:

```shell
debug[0]: *** BEGIN PARSE:
'
+
'
Starting parse
Entering state 0
Stack now 0
Reading a token
--accepting default rule ("+")
--(end of buffer or a NUL)
--EOF (start condition 0)
[...]
```

Maybe it's time I add the nodefault option.

lcn2:
The JSON parser function should not allow garbage characters.  I.e., it should not declare _valid JSON_ that contains junk:

```json
{ 
"curds" : "whey",
/* Not allowed &
this is junk */; 
"foo" : 123 }
```

The JSON parser function should NOT return `true` to the above text.

It would be nice if error messages could identify the source of such junk, but at a minimum it should return `false` to the above.

xexyl:
> The JSON parser function should not allow garbage characters. I.e., it should not declare _valid JSON_ that contains junk:
> 
> ```json
> { 
> "curds" : "whey",
> /* Not allowed &
> this is junk */; 
> "foo" : 123 }
> ```
> 
> The JSON parser function should NOT return `true` to the above text.
> 
> It would be nice if error messages could identify the source of such junk, but at a minimum it should return `false` to the above.

Right. I'm not sure how to show the invalid input. I just had an idea though. I'll play with this quickly and see. If not I'll have the change in git diff so I can maybe play with it later. The idea is to return an error token but have the string itself printed in the error.

xexyl:
> > The JSON parser function should not allow garbage characters. I.e., it should not declare _valid JSON_ that contains junk:
> > ```json
> > { 
> > "curds" : "whey",
> > /* Not allowed &
> > this is junk */; 
> > "foo" : 123 }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The JSON parser function should NOT return `true` to the above text.
> > It would be nice if error messages could identify the source of such junk, but at a minimum it should return `false` to the above.
> 
> Right. I'm not sure how to show the invalid input. I just had an idea though. I'll play with this quickly and see. If not I'll have the change in git diff so I can maybe play with it later. The idea is to return an error token but have the string itself printed in the error.

The idea will work I just have to sort out a linker error. But I'll worry about that tomorrow or the next day.

PR #179, opened 2022-04-27
JSON parser updates + fixes in txzchk

xexyl:
The parser updates are merely comments. This includes pointing out (and I neglected to say this in the git log) that JSON_STRING is now correct. But it also addresses the fact that the ugly_ prefix makes it look like our struct json is ugly:

```
    This is what's been added (the first paragraph I just added some _s to
    the text to stress certain words):
    
         * A negative consequence here is that because of the prefix being set to
         * ugly_ there's a typedef in the parser that _might suggest_ that _our struct
         * json_ is ugly:
         *
         *      typedef struct json UGLY_STYPE;
         *
         * At first glance this is a valid concern. However we argue that even if this
         * is so the struct might well have to be ugly because it's for a json parser; a
         * json parser necessarily has to be ugly due to the spec: one could easily be
         * forgiven for wondering if the author of the json specification was on drugs
         * at the time of writing.
         *
         * Please note that we're _ABSOLUTELY NOT_ saying that they were and we use the
         * term very loosely as well: we do not want to and we are not accusing anyone
         * of being on drugs (we rather find addiction a real tragedy and anyone with an
         * addition should be treated well and given the help they need) but the fact is
         * that the JSON specification is barmy and those who are in favour of it must
         * surely be in the JSON Barmy Army (otherwise known as the Barmy Army of the
         * Jocular Sillies On Narcotics :-)).
         *
         * Thus as much as we find the specification objectionable we rather feel sorry
         * for those who are indeed in the JSON Barmy Army and we apologise to them in a
         * light and fun way and we hope that they're not terribly humour impaired.
    
```

I think you'll appreciate the new translation of JSON that I made. :) 

As for the apology I added:

```
        * On the other hand the fact the parser IS for JSON might suggest that the
        * parser SHOULD be ugly. And SINCE IT IS UGLY if one were to ignore the fact
        * that bison and flex do not care what they are parsing this same person could
        * argue, obviously by some perverted logical fallacies, that bison and flex did
        * their job well. :-)
```

The fixes in `txzchk` I will let you look at: minor issues in most cases but if someone was messing about with the tarball or some new format of tar output was discovered it now triggers not only triggers a warning (already did this) but counts it as an issue (did not do this).


xexyl:
I am afraid this is probably all I'll get done today but we'll see. I'm going to have a lie down now and after that I'll get into other things.

I do have a question for you though. Well two actually.

1. Once `struct json_number` and its routines are done can you think of anything else that might be needed before the parser can be written?
2. Since the json grammar has punctuation is there a problem with having an enum type for these symbols? Or can you not think of a reason to have them? I'm not sure yet. Thinking quickly since the punctuation will be part of the others it's probably not needed other than to return to the parser the type but it did cross my mind so I'm asking anyway.

lcn2:
> I am afraid this is probably all I'll get done today but we'll see. I'm going to have a lie down now and after that I'll get into other things.
> 
> 
> 
> I do have a question for you though. Well two actually.
> 
> 
> 
> 1. Once `struct json_number` and its routines are done can you think of anything else that might be needed before the parser can be written?

The compound JSON item (_member_, _object_, _array_) needs to be written.

Those JSON items will need _dynamic array_ references (an additional lookup feature that acts kind of like pointers that is able to maintain the reference even when `realloc()` moves the expanding data block) needs to be added to the _dynamic array_ services.

> 2. Since the json grammar has punctuation is there a problem with having an enum type for these symbols? Or can you not think of a reason to have them? I'm not sure yet. Thinking quickly since the punctuation will be part of the others it's probably not needed other than to return to the parser the type but it did cross my mind so I'm asking anyway.

There should not be a problem if you end up thinking you need to add them.

As far as the JSON parse tree, it is not needed, however, so there shouldn't be references to things like `JTYPE_COMMA` nor should they end up in the `struct json`.

lcn2:
Thanks for the pull request, @xexyl.

We probably won't get to work on [the above mentioned items](https://github.com/ioccc-src/mkiocccentry/pull/179#issuecomment-1111143236).

Today, there is an OT focus on giving the following talk to the maths department of California State University at Sonoma:

[A Grand Coding Challenge!
Finding a new Largest Known Prime](http://www.isthe.com/chongo/tech/math/prime/prime-tutorial.pdf)

xexyl:
> Thanks for the pull request, @xexyl.
> 
> We probably won't get to work on [the above mentioned items](https://github.com/ioccc-src/mkiocccentry/pull/179#issuecomment-1111143236).

No worries.

> 
> Today, there is an OT focus on giving the following talk to the maths department of California State University at Sonoma:
> 
> [A Grand Coding Challenge! Finding a new Largest Known Prime](http://www.isthe.com/chongo/tech/math/prime/prime-tutorial.pdf)

That sounds like fun! And I'll look at that link too.

PR #185, opened 2022-04-29
Fix memory leak in verge.c

xexyl:
In allocate_vers() a string was strdup()d but prior to returning the
copy was not freed.

### DIFF
--- a/verge.c
+++ b/verge.c
@@ -281,9 +281,11 @@ allocate_vers(char *str, long **pvers)
     if (i == len) {
 	/* report invalid version string */
 	dbg(DBG_MED, "version string contained no digits: <%s>", wstr);
-	/* free strdup()d string */
-	free(wstr);
-	wstr = NULL;
+	/* free strdup()d string if != NULL */
+	if (wstr != NULL) {
+	    free(wstr);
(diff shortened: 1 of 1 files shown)


xexyl:
Really going to do something else now. Good day!

lcn2 (changes requested):
We suggest, as a _defense in depth_ technique, that you check that the variable is not NULL before the free and setting to NULL.

xexyl:
I can do that - though it's already checked for NULL earlier on and it aborts if it is. Though one could say that it could have been freed unintentionally and not set back to NULL.

xexyl:
Done. The reason I sometimes don't check is because you can safely free a NULL pointer and I always set to NULL after using free(). But anyway it's done.

xexyl:
> I can do that - though it's already checked for NULL earlier on and it aborts if it is. Though one could say that it could have been freed unintentionally and not set back to NULL.

I did set to NULL though fwiw. I always do. Anyway all fixed.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Welcome.

PR #188, opened 2022-05-01
Minor fixes, txzchk.1 update and add jparse.1

xexyl:
This is all I'm doing today almost certainly as I do not feel so well today. I'll do more tomorrow.

PR #189, opened 2022-05-01
Minor fixes, txzchk.1 update and add jparse.1 #188

xexyl:
Sorry for the confusion of the earlier pull requests. Wanted to clean it up. I'm off to do other things now as I'm having a difficult day. Going to take it easy but I can discuss things with you still.

Good day!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -753,49 +753,51 @@ depend: all
 
 ### DO NOT CHANGE MANUALLY BEYOND THIS LINE
 utf8_posix_map.o: utf8_posix_map.c utf8_posix_map.h util.h dyn_array.h \
-  dbg.h limit_ioccc.h version.h
-jparse.o: jparse.c jparse.h dbg.h util.h dyn_array.h json.h sanity.h \
-  location.h utf8_posix_map.h json_chk.h json_util.h limit_ioccc.h \
-  version.h jparse.tab.h
-jparse.tab.o: jparse.tab.c jparse.h dbg.h util.h dyn_array.h json.h \
+  dbg.h version.h limit_ioccc.h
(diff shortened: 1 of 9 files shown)


lcn2:
Thanks @xexyl .. we may need to process the dbg changes as it is used in other tools as well

xexyl:
> Thanks @xexyl .. we may need to process the dbg changes as it is used in other tools as well

Welcome. I'm not sure what might have to change though as I only made it so the version is available to all the other utils - and `limit_ioccc.sh`. Well I guess I'll find out.

lcn2:
The dbg system is self contained so we will need to do something slightly different for version.h.

See commit 0766073a329b02a1acf5b0fd19fa94de4d77aede

    Release dbg v1.10 2022-05-01

    Changed dbg version string to DBG_VERSION.

    Moved DBG_VERSION back into dbg.h so that using the
    dbg facility only requires one include file (dbg.h)
    and compiling dbg.c to form dbg.o.

xexyl:
> The dbg system is self contained so we will need to do something slightly different for version.h.
> 
> See commit [0766073](https://github.com/ioccc-src/mkiocccentry/commit/0766073a329b02a1acf5b0fd19fa94de4d77aede)
> 
> ```
> Release dbg v1.10 2022-05-01
> 
> Changed dbg version string to DBG_VERSION.
> 
> Moved DBG_VERSION back into dbg.h so that using the
> dbg facility only requires one include file (dbg.h)
> and compiling dbg.c to form dbg.o.
> ```

That makes sense yes. Just to remind in case you forgot: you ran `make depend` after doing this, right?

PR #191, opened 2022-05-02
Lexer/parser/other updates + man page addition

xexyl:


xexyl:
Commit 362a6e9 is likely the last one I do today. Need to take a break and not feeling up to doing any more at least not right now. I do think I feel better than I have the past few days but that's hard to know for sure. In any event though I have other things to do. Good day!

lcn2:
Good idea on the rename of `test.sh`.  That name was a poor choice in highlight.  Thanks for the pull request @xexyl.

xexyl:
> Good idea on the rename of `test.sh`. That name was a poor choice in highlight. Thanks for the pull request @xexyl.

It's easy enough to do though because it corresponds to `make test`. I'll be able to reply to the comments shortly.

PR #192, opened 2022-05-03
Various fixes (mostly JSON specific)

xexyl:
I hope that commit https://github.com/ioccc-src/mkiocccentry/commit/900202f320bf20621c6611f901bbca7a13f8f0ac does not have any mistakes with the timestamp changes. As I noted in a comment in a thread I based the need to NOT change the timestamp on the filenames. I also renamed a file as it gives a better indication of what is wrong: mkiocccentry version is wrong but it said just `author.version`.

I will reply to your comment in the json thread/issue later on. Taking another break and no idea what I'll do after that. Good day!


lcn2:
Good stuff @xexyl 👍

PR #195, opened 2022-05-07
Update JSON parser including one action added

xexyl:
Add function 'char const *json_element_name(enum element_type
json_type)' to print the element type by name rather than a number.

Add to the conversion functions a call (in some cases more than one
call) to dbg() with new debug value (#defined in json.h as DBG_NONE)
JSON_DBG_LEVEL so that I can see the type in each call to the conversion
functions. The reason I named it JSON_DBG_LEVEL is so that it's easy to
remove once this is no longer needed (alternatively it could be changed
to some other debug value if it's worth keeping).

Add action to json_member so that it actually parses a member (it does
NOT add to the parse tree as currently there is no parse tree) or a list
of members. The diff in jparse.y looks like:

    -json_member:	JSON_STRING JSON_COLON json_element
    +json_member:	JSON_STRING JSON_COLON json_element { $$ = *json_conv_member(&$1, &$3); }
		    ;

Removing the debug information from the lexer and parser themselves
here's an example output:

    $ ./jparse -s '{ "test": true, "complete": false, "number": -555 }' 2>&1 |grep debug
    debug[0]: Calling parse_json_block("{ "test": true, "complete": false, "number": -555 }"):
    debug[0]: *** BEGIN PARSE:
    debug[0]: json_conv_member: JSON name type: string
    debug[0]: json_conv_member: JSON value type: boolean
    debug[0]: json_conv_member: JSON name type: string
    debug[0]: json_conv_member: JSON value type: boolean
    debug[0]: json_conv_member: JSON name type: string
    debug[0]: json_conv_member: JSON value type: number
    debug[0]: *** END PARSE

Perhaps the actual value of each type itself should be printed as well
but this is a good start.

NOTE: The concept of the parse_json_() functions from commits cc3173e
and b0b426c084 might actually be invalid which I noted in jparse.y:

    /*
    * XXX The concept of the below parse_json_() functions might be invalid as the
    * value of $$ variables in the bison actions is of type struct json * and these
    * functions take a char const *.
    */

### DIFF
--- a/jparse.tab.ref.c
+++ b/jparse.tab.ref.c
@@ -558,9 +558,9 @@ static const yytype_int8 yytranslate[] =
 /* YYRLINE[YYN] -- Source line where rule number YYN was defined.  */
 static const yytype_uint8 yyrline[] =
 {
-       0,   147,   147,   148,   149,   150,   153,   154,   155,   156,
-     157,   158,   159,   162,   164,   167,   168,   171,   174,   177,
-     178,   181
+       0,   148,   148,   149,   150,   151,   154,   155,   156,   157,
+     158,   159,   160,   163,   166,   169,   170,   173,   176,   179,
+     180,   183
 };
 #endif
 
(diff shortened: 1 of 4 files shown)


xexyl:
I'm afraid this is all I'll be doing today as I'm exhausted (truly knackered in fact) and as I said earlier in the week it'll be a long day (though a great day) so I need to get some rest. I feel really good about this commit though! It's possible I manage to get a bit more in today but not sure I will. If I don't within a couple hours I almost certainly won't.

Hope you finally got some sleep and good day!

lcn2:
This looks like important progress, @xexyl : thank you!

xexyl:
> This looks like important progress, @xexyl : thank you!

Thank you too!

PR #196, opened 2022-05-07
Bug fix and add JSON parser actions + JSON tree

xexyl:
In the parse_json_bool() function I accidentally had checked the
converted bool of 'null' in the union. This has been fixed.

Add several more actions to the parser. In particular:

    json_value:       json_object
		    | json_array
		    | JSON_STRING { $$ = *parse_json_string(ugly_text, &tree); }
		    | json_number
		    | JSON_TRUE { $$ = *parse_json_bool(ugly_text, &tree); }
		    | JSON_FALSE { $$ = *parse_json_bool(ugly_text, &tree); }
		    | JSON_NULL { $$ = *parse_json_null(ugly_text, &tree); }
		    ;

    json_number:    JSON_NUMBER { $$ = *parse_json_number(ugly_text, &tree); }
		    ;

In parse_json_null() check if converted is true. If not it's an error;
else print a debug statement saying it was converted.

In parse_json_number() it still does not print the number or check for
conversion. Part of this is there's more to check and I have to head off
for the day.

In parse_json_bool() there also was no change except for the bug fix as
above. However the function is now called when encountering JSON_TRUE or
JSON_FALSE.

The parse_json_member() function is not used yet and it's probable that
it'll have to change definition as well, taking a struct json *member.

JSON_OBJECT might be in a way mostly accounted for via the json_members
but not all is handled there and nothing is linked into the tree yet.
What this means is I believe (I might be wrong) is that except for
JSON_OBJECT and JSON_ARRAY conversion is done. What's not done is all
the checks on these and also linking it into the parse tree.

Added 'struct json tree' which is zero initialised. Right now the
parse_json_() functions are passed &tree but it's not used except to
check for NULL pointer. At a later commit at a later tree will be used.

There was a typo fix ('the the' by accident in json.c).

I hope that I documented everything here: there was a lot and I have to
stop working on this.

### DIFF
--- a/jparse.h
+++ b/jparse.h
@@ -102,6 +102,7 @@ extern FILE *ugly_in;			/* input file lexer/parser reads from */
 extern unsigned num_errors;		/* > 0 number of errors encountered */
 extern bool output_newline;		/* true ==> -n not specified, output new line after each arg processed */
 extern int token;			/* for braces, brackets etc.: '{', '}', '[', ']', ':' and so on */
+extern struct json tree;		/* the parse tree */
 
 /*
  * function prototypes
@@ -125,7 +126,6 @@ void parse_json_block(char const *string);  /* parse a string as a JSON block */
  *
  * XXX - these are subject to change and are very incomplete as well - XXX
  */
-struct json *parse_json_name(char const *string, struct json *ast);
 struct json *parse_json_string(char const *string, struct json *ast);
 struct json *parse_json_number(char const *string, struct json *ast);
 struct json *parse_json_bool(char const *string, struct json *ast);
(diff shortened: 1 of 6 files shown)


xexyl:
This was done quickly so maybe some things have to change but I made even more progress! I just remembered something I forgot mention: I removed the `parse_json_name()`. 

Have a great day and I hope you can get rest tonight! More from me tomorrow. I'll be able to reply for a bit btw.

lcn2:
Thanks again for the pull request @xexyl 

We will get some sleep and then address the problem that we seem to have created by an earlier commit: assuming it hasn't been fixed.

xexyl:
Just noticed some comments that are no longer valid. The parse functions no longer return a NULL pointer as of yesterday. I'll fix these comments tomorrow or the next day.

In the middle of a film which I hope to finish before my cousin gets here but it's not important if I do.

xexyl:
> Thanks again for the pull request @xexyl
> 
> We will get some sleep and then address the problem that we seem to have created by an earlier commit: assuming it hasn't been fixed.

Sure. Sleep well my friend! Curious what this problem is - but I'll find out in the git log if it hasn't been fixed.

PR #198, opened 2022-05-08
JSON parser updates

xexyl:
The json_member action now uses json_conv_member(). This means that the
parse_json_member() now is:

    struct json *parse_json_member(struct json *name, struct json *value, struct json *ast);

and the action looks like:

    +json_member:	JSON_STRING JSON_COLON json_element { $$ = *parse_json_member(&$1, &$3, &tree); }

which in the generated code is:

    { yyval = *parse_json_member(&yyvsp[-2], &yyvsp[0], &tree); }

Override default debug level in jparse to be DBG_MED and change
JSON_DBG_LEVEL to DBG_MED. I have the idea that a json debug function
could be added that is prefixed with JSON instead of dbg and which might
take a different variable which would mean that the other tools won't be
chatty but jparse can be. However for now this is not done. I made it
DBG_MED as I don't want too much information as there already is a lot
of information.

Updated some comments and remove duplicate XXX comments (some anyway).
There were comment fixes as well - the functions (except of the one for
arrays) do not return NULL pointers as of some commits ago.

TODO:

* The conversion routines should print better debug information. Right
now it only shows the type but it could also display the value - at
least for ones that have a value that's easy to define (strings being
the most obvious example).

* Add safer handling of union checking in struct json in the parse
functions. This will most likely be done in one of the next few commits
but it's possible it won't be today. But if it's not today it'll be in
the next few days.

* Write the parsing of arrays and verify that everything looks good.
After that we can start thinking about the parse tree and linking the
structs into it.

* Decide if the tree should be a global variable or not. I made it a
global variable yesterday to quickly get it in as I had to get going but
I was making use of the parse_json_() functions so I felt I should pass
in a tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ makefile.local
 
 # temporary files
 *.tmp
+
+# parser logs
+jparse.output
--- a/Makefile
+++ b/Makefile
@@ -493,7 +493,7 @@ limit_ioccc.sh: limit_ioccc.h version.h Makefile
 #
 jparse.tab.c jparse.tab.h: jparse.y jparse.h sorry.tm.ca.h run_bison.sh limit_ioccc.sh verge \
 	jparse.tab.ref.c jparse.tab.ref.h Makefile
-	./run_bison.sh -b ${BISON_BASENAME} ${BISON_DIRS} -p jparse -v 1 ${RUN_O_FLAG} -- -d
+	./run_bison.sh -b ${BISON_BASENAME} ${BISON_DIRS} -p jparse -v 1 ${RUN_O_FLAG} -- --report all -d
 
 # How to create jparse.c
 #
@@ -692,7 +692,7 @@ clean_generated_obj:
 prep_clobber:
 	${RM} -f ${TARGETS} ${TEST_TARGETS}
 	${RM} -f ${GENERATED_CSRC} ${GENERATED_HSRC}
-	${RM} -f answers.txt j-test.out j-test2.out json-test.log
+	${RM} -f answers.txt j-test.out j-test2.out json-test.log jparse.output
 	${RM} -rf test-iocccsize test_src test_work tags dbg.out
 	${RM} -f dbg_test.c
 	${RM} -rf dyn_test.dSYM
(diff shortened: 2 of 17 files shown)


xexyl:
I hope to reply to the comments from yesterday today but I'm not sure when that will be. It might be soon but might not be until later on. If not it'll be tomorrow. 

As I noted in one of the todo items:

```
Add safer handling of union checking in struct json in the parse
functions. This will most likely be done in one of the next few commits
but it's possible it won't be today. But if it's not today it'll be in
the next few days.
```

and

```
Decide if the tree should be a global variable or not. I made it a
global variable yesterday to quickly get it in as I had to get going but
I was making use of the parse_json_() functions so I felt I should pass
in a tree.
```

the valid things you brought up will be addressed in an upcoming commit. The union one will be done before the other most likely.

xexyl:
In case you didn't know:

```The way
    this was solved is by making JSON_STRING refer to json_string which
    itself called parse_json_string() called rather than it being directly
    in the JSON_STRING action in json_value.
```

is called the 'change the thought mid-sentence and forget to update the rest of the sentence problem'.  Compare with nonsense. Sorry for that!

xexyl:
Leaving until tomorrow. Good day my friend! I hope to do more tomorrow. I'm not sure if it'll be arrays or something else but it would be nice if I could work on it. However I think it'll be some days before I feel up to looking into that. Anyway enjoy the rest of your day!

xexyl:
Off until tomorrow.. Hope you're all right. Good day!

xexyl:
First of all I want to say I hope you're okay - as I've not heard from you in a couple days.

Second of all I have a commit ready but I've not yet committed it as I need a break. It separates normal debug information from json debug information. 

I will commit this later on today. Hope you're well and having a nice sleep my friend!

lcn2:
Back after some OT activity .. thanks for the git pull.  We will try to get to your questions.

xexyl:
> Back after some OT activity .. thanks for the git pull. We will try to get to your questions.

Thank you. Glad you're okay!

PR #201, opened 2022-05-13
string <-> bool and JSON updates

xexyl:


lcn2:
Good call, @xexyl

xexyl:
> Good call, @xexyl

I guess you mean on all the sanity checks of boolean? I think so too. If not what do you refer to please? I guess other types could have similar checks added though what they should be at this time I don't know - though it's dead early and I also am not looking at the structs.

PR #204, opened 2022-05-16
JSON parser updates

xexyl:
The comments reflect more recent developments with the parser including
no longer stating that there are no actions (in both flex and bison
files).

Although I prefer JSON_WHITESPACE the spec uses ws so we use JSON_WS
(eventually we won't even need to do anything with it so this won't be
an issue at all).


xexyl:
Question for you Landon. I think I know the answer but since I put the code in before it occurred to me (or maybe it did occur to me at the time too) I decided to keep it in. See commit a0602e2 and specifically this part:

```c

    /*
     * additional sanity checks
     */
    length = strlen(ptr);
    if (length < len) {
        /*
         * XXX Is this actually correct? If there is a NUL byte embedded in the
         * string and we're wanting to ignore NULs, the strlen could be < the
         * length.
         */
        err(37, __func__, "total strlen() %zu < length of len %zu", (uintmax_t)length, (uintmax_t)len);
        not_reached();
    }
```

The fact that there might be embedded NUL bytes suggests to me that this test won't always be valid. Of course in my tired enough head it's also possible that I have messed up the logic in the first place. I'm going to take a break now and I hope to return to this function in a while.

lcn2:
> Question for you Landon. I think I know the answer but since I put the code in before it occurred to me (or maybe it did occur to me at the time too) I decided to keep it in. See commit [a0602e2](https://github.com/ioccc-src/mkiocccentry/commit/a0602e279da903b3ae2b5c03a478096f65771a79) and specifically this part:
> 
> ```c
>     /*
>      * additional sanity checks
>      */
>     length = strlen(ptr);
>     if (length < len) {
>         /*
>          * XXX Is this actually correct? If there is a NUL byte embedded in the
>          * string and we're wanting to ignore NULs, the strlen could be < the
>          * length.
>          */
>         err(37, __func__, "total strlen() %zu < length of len %zu", (uintmax_t)length, (uintmax_t)len);
>         not_reached();
>     }
> ```
> 
> The fact that there might be embedded NUL bytes suggests to me that this test won't always be valid. Of course in my tired enough head it's also possible that I have messed up the logic in the first place. I'm going to take a break now and I hope to return to this function in a while.

We don't think the check is correct if we want to allow for `NUL` (i.e., a single byte with the numeric value of 0) in the text.

A careful read of [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) suggests that there is nothing in JSON that prohibits a JSON document contain a single byte NUL byte inside it.

So is appears that the ` if (length < len) { .. }` check should be removed.

lcn2:
Thanks @xexyl and see [comment 1127887442](https://github.com/ioccc-src/mkiocccentry/pull/204#issuecomment-1127887442)

PR #211, opened 2022-05-18
JSON parser now allows NUL bytes - resolve #208

### LINKED ISSUE #208: Bug: Allow NUL bytes in JSON
**Describe the bug**
In some cases, NUL bytes are disallowed in the JSON parser.

**To Reproduce**
Steps to reproduce the behavior (adjust steps as needed):
1. echo -n '"NUL byte"' | ./jparse -J1 -
2. echo -n '"NULXbyte"' | tr 'X' '\0' | hd
3. echo -n '"NULXbyte"' | tr 'X' '\0' | ./jparse -J1 -

**Expected behavior**
While command 1 works, and command 2 shows the NUL byte is being fed to stdin of jparse,
command 3 results in an error and parse exits non-zer0.

**Screenshots**
$ echo -n '"NULXbyte"' | tr 'X' '\0' | ./jparse -J1 -
JSON DEBUG[1]: Calling parse_json_file("-"):
Warning: parse_json_file: found embedded NUL byte in stdin
zsh: done       echo -n '"NULXbyte"' | tr 'X' '\0' |
zsh: exit 1     ./jparse -J1 -

**Environment (please complete the following information, or write n/a where not appliable):**
 - OS: macOS
 - OS distribution: Apple
 - OS Version: 12.31.1
 - Compiler: n/a
 - Compiler version: n/a
 - Bison version: 3.8.2
 - Flex verson: 2.6.4
 - Shell: zsh,
 - Airspeed Velocity of an Unladen Swallow: African or European?

**Additional context**
It seems to be OK to warn about _NUL_ bytes, but a struct reading of [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) suggests that there is no prohibition against having a _NUL_ byte in JSON.

NOTE: When we say _NUL_ byte, we do not refer to the two byte sequence ASCII backslash followed by ASCII zero (i.e., not \0), we refer to a single byte with a numeric value of 0.

BTW: The only place were a NUL byte appears to her allowed is inside a JSON _string_.  A _NUL_ byte outside of a JSON _string_ will be a syntax error.  Nevertheless a check should be made that a NUL byte outside of a JSON _string_ will be detected and marked as a syntax error.

xexyl:
If a string (or a file) has a NUL byte not within a string it is a
syntax error in the parser and parsing will not continue. If however a
NUL byte is in a string it will be accepted.

The way this works is quite simple. The string regex already allowed for
NUL bytes and the default rule (that we define later) in the lexer does
not allow for NUL bytes. Now in the lexer we have the length of each
pattern that matched in ugly_leng. This variable does not exist in bison
though so for each string matched we set ugly_length to ugly_leng
(perhaps this should always be done?). Now the parse_json_string()
function has changed to be:

    /* parse_json_string - parse a json string
     *
     * given:
     *
     *	string	    - the text that triggered the action
     *	len	    - length of the string to convert (important for NUL bytes)
     *
     * Returns a pointer to a struct json with the converted string.
     *
     * NOTE: The len is important for strings that have NUL bytes: without it we
     * would rely on strlen() which would mean that the first NUL byte would be the
     * end of the string. If len <= 0 this function uses strlen() on the string.
     *
     * NOTE: This function does not return if passed a NULL string or if conversion
     * fails.
     */
    struct json *
    parse_json_string(char const *string, size_t len)
    {
    }

Then in the bison grammar rule for string the action looks like:

    json_string: JSON_STRING
		    {
			$$ = parse_json_string(ugly_text, ugly_length);
			json_dbg(JSON_DBG_MED, __func__, "under json_string: after JSON_STRING returning type: %s",
					   json_element_type_name($$));
			json_dbg(JSON_DBG_LOW, __func__, "under json_string: after JSON_STRING before ;");
		    }
		    ;

The important line is the one that calls parse_json_string():

    $$ = parse_json_string(ugly_text, ugly_length);

An important thing to note is that when the debug functions show what is
about to be parsed has a NUL byte it will stop at the first NUL so that
for a file that has a NUL byte between two "s you would see:

    JSON DEBUG[1]: Calling parse_json_block(""", 4):
    JSON DEBUG[1]: *** BEGIN PARSE:
    <
    "
    >

because the NUL byte after the first '"' makes the debug function think
it's reached the end of the string. There are a few ways to go about
solving this which I'm going to bring up on the GitHub issue as this is
already much too long.

In any case this resolves issue #208 for both files with NUL bytes as
well as strings on the command line with NUL bytes.

### DIFF
--- a/jparse.h
+++ b/jparse.h
@@ -110,6 +110,7 @@ extern unsigned num_errors;		/* > 0 number of errors encountered */
 extern bool output_newline;		/* true ==> -n not specified, output new line after each arg processed */
 extern int token;			/* for braces, brackets etc.: '{', '}', '[', ']', ':' and so on */
 extern struct json tree;		/* the parse tree */
+extern int ugly_length;			/* for regex length: needed in some cases in parser (where ugly_leng does not exist) */
 
 /*
  * function prototypes
@@ -133,7 +134,7 @@ void parse_json_block(char const *string, size_t len);  /* parse a string as a J
  *
  * XXX - these are subject to change and some are incomplete as well - XXX
  */
-struct json *parse_json_string(char const *string);
+struct json *parse_json_string(char const *string, size_t len);
 struct json *parse_json_number(char const *string);
 struct json *parse_json_bool(char const *string);
 struct json *parse_json_null(char const *string);
--- a/jparse.l
+++ b/jparse.l
@@ -101,6 +101,8 @@
 /* our header file - #includes what we need */
 #include "jparse.h"
 
+int ugly_length = 0;
+
 /*
  * An exception where the prefix does not change YY to UGLY_ is YY_BUFFER_STATE
  * but because it IS ugly we have done it for them in jparse.h so that where one
@@ -161,6 +163,7 @@ JSON_COMMA		","
 
 {JSON_STRING}		{
 			    /* string */
+			    ugly_length = ugly_leng;
 			    printf("\nstring: <%s>\n", ugly_text);
 			    return JSON_STRING;
 			}
--- a/jparse.ref.c
+++ b/jparse.ref.c
@@ -881,8 +881,8 @@ int yy_flex_debug = 1;
 
 static const flex_int16_t yy_rule_linenum[14] =
     {   0,
-      152,  162,  168,  174,  180,  185,  191,  196,  203,  209,
-      216,  223,  230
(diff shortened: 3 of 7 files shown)


lcn2:
Thanks for this pull request @xexyl!

xexyl:
> Thanks for this pull request @xexyl!

Happy to help!

PR #213, opened 2022-05-19
New utility functions to show unprintable bytes + other updates and fixes

xexyl:
The utility functions are in commit https://github.com/ioccc-src/mkiocccentry/commit/771b959ad77d53f52094ede3e448963e6b9e399e and are based on my idea at issue #208.


lcn2:
Wow, lots of good updates in this pull request, @xexyl.

xexyl:
> Wow, lots of good updates in this pull request, @xexyl.

Thank you! Just had a moment to reply. Going again. Might be able to reply to others later on but if not then I will tomorrow.

xexyl:
Oh btw: it occurred to me. Maybe the functions I moved would better belong in jparse? Not sure. I can see it both ways. 

More from me later or else tomorrow.

lcn2:
It appears that the `extern` function declarations in various `*.h` files are not correct.

For example, `json_conv_bool_str()` is now in `json_util.c` but the extern is in `json.h` and not in `json_util.h`.

lcn2:
> It appears that the `extern` function declarations in various `*.h` files are not correct.
> 
> For example, `json_conv_bool_str()` is now in `json_util.c` but the extern is in `json.h` and not in `json_util.h`.

Please consider checking `*.h` files and where possible, have the `extern` function declarations in `foo.h` be in the same order that they are found in `foo.c`.

xexyl:
> It appears that the `extern` function declarations in various `*.h` files are not correct.
> 
> For example, `json_conv_bool_str()` is now in `json_util.c` but the extern is in `json.h` and not in `json_util.h`.

There’s a good reason for this too. I went to change it but ISO C doesn’t allow for enum forward declarations so I kept it there in `json.h`. If you have a solution I did not think of at the time please let me know. But that’s why it wasn’t moved over. 

I had planned on putting that into the log but maybe I forgot? I don’t remember now.

xexyl:
> > It appears that the `extern` function declarations in various `*.h` files are not correct.
> > For example, `json_conv_bool_str()` is now in `json_util.c` but the extern is in `json.h` and not in `json_util.h`.
> 
> Please consider checking `*.h` files and where possible, have the `extern` function declarations in `foo.h` be in the same order that they are found in `foo.c`.

I sometimes do this as well but today I wanted it just move the functions over (and see above comment). I can look at it more in detail in the coming days no problem.

PR #215, opened 2022-05-20
Various updates

xexyl:
Contrary to what vim thinks the word is 'writable' is not a regional
spelling for 'writeable'. The word is 'writable' full stop. That
/usr/share/dict/words has it in under linux and macOS does not change
the fact it's both incorrect and not even a word so spelt (iOS also has
both, interestingly) even if it might make more sense (to some people).
The thing is spelling often does not make sense. For an amusing and
ironic example: short is longer than long. Possibly it's modelled after
other words like uniteable (alternative spelling for unitable though vim
does not recognise either word and neither do some other dictionaries
but both mean some things/people that can be united). But I've digressed
well beyond the point due to my love of words. The point is writable is
the correct spelling.

### DIFF
--- a/jparse.l
+++ b/jparse.l
@@ -352,7 +352,7 @@ parse_json(char const *ptr, size_t len, bool *is_valid, FILE *dbg_stream)
     ugly_parse();
 
     /*
-     * scan and parse cleanup
+     * scan and parse clean up
      */
     ugly__delete_buffer(bs);
     bs = NULL;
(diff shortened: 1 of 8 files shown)


lcn2:
😄 for the @xexyl the type fixes

xexyl:
> 😄 for the @xexyl the type fixes

The laughter is from my log messages?

xexyl:
Ah yes. I see you reacted to that. I'm glad you appreciate it. I needed that! Having a day where I'm doubting myself.

lcn2:
> > Question @xexyl: Is there a way to add a word a vim dictionary to indicate that it is mis-spelled?
> 
> That's a good question and one I should look into. Actually I tried earlier today to see if `zug` would work but unfortunately it does not (I did not expect it to). I've never had to do this until today so I haven't actually looked into it. Maybe I will but not now. Going to rest. Good luck with the talk!

Thanks!

xexyl:
Happy to help!

PR #217, opened 2022-05-20
Typo fixes in prep.sh

xexyl:
Presumably a yank paste error: the help for -o option only referred to
bison but for prep.sh it's both bison and flex (the same message is in
run_bison.sh and the flex counterpart is in run_flex.sh).

Also:

    -    echo "$0: ERROR: Makefile file not found: $MAKEFILE" 1>&2
    +    echo "$0: ERROR: Makefile not found: $MAKEFILE" 1>&2

### DIFF
--- a/prep.sh
+++ b/prep.sh
@@ -35,7 +35,7 @@ export USAGE="usage: $0 [-h] [-v level] [-V] [-e] [-o] [-m make] [-M Makefile]
     -V              print version and exit 5
 
     -e		    exit in first make action error (def: exit only at end)
-    -o		    do NOT use backup files, fail if bison cannot be used (def: use)
+    -o		    do NOT use backup files, fail if bison or flex cannot be used (def: use)
     -m make	    make command (def: make)
     -M Makefile	    path to Makefile (def: ./Makefile)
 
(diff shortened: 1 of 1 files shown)


xexyl:
That's almost certainly the last I'll be doing today. Off to do something else now.

xexyl:
Leaving for the day. Can maybe reply to some comments later on. I hope I can do more tomorrow. If I can have a better night then it's possible.

Hope the talk you referred to went well and I'll BCNU. Good day!

lcn2:
Good catch @xexyl

xexyl:
> Good catch @xexyl

Happy to help. I know there are many more things like it to resolve. At one point I had thought of going through the files but then a moment later it occurred to me that that’s silly since it would mean I would have to reread everything again (which might actually happen but I am okay with that). The reason it even popped into my head is I didn’t feel up to anything else but I wanted to do something for the repo. 

Fortunately I have a gift in finding errors like typos and such. It will be very helpful as you pointed out before. It will be interesting to see how many changes actually will be needed. Well hopefully we’re getting closer to that point. I will reply to the other comments and then try and sleep.

PR #218, opened 2022-05-21
Release jparse version 0.2 2022-05-22

xexyl:


xexyl:
Another todo item is to identify debug statements that result in things like:

```sh
-> $$ = nterm json_value ()
Entering state 9
Stack now 0 9
Reducing stack by rule 18 (line 435):
   $1 = nterm json_value ()
JSON DEBUG[1]: under json_value: $json_element type: ((node == NULL))
JSON DEBUG[1]: under json_value: $json_value type: ((node == NULL))
JSON DEBUG[1]: under json_element: after json_value before ;
```

and consider if they're needed. I'm not sure if they are. But then it might be that in some cases they won't be NULL which is one of the reasons I said 'consider'. If they can be removed though that would be good as it would get rid of some unnecessary output. Of course eventually none of this will be necessary but for now I think it's quite helpful.

As a comment said in the other thread I'm mostly done for the day - quite tired and still not feeling so well. I feel good about what I got done though. Going to do something else but I can reply throughout the day most likely.

xexyl:
Heading off for the day. Hope you're having a great evening. Enjoy the rest of your day and sleep well when you do!

More from me tomorrow.

xexyl:
Commit https://github.com/ioccc-src/mkiocccentry/pull/218/commits/92c416e8f729273547d7c9e80938bee2a1e300aa seems worthy of a version update and I think you'll agree. Off to do something else now. Good day!

lcn2:
Good updates, @xexyl

xexyl:
> However, see [comment 1133956856](https://github.com/ioccc-src/mkiocccentry/issues/156#issuecomment-1133956856) for a commit that reversed part of this pull request.

That's okay. As I noted in the other place I only did it for the normal debug and warn functions because I was doing it for the json ones as well. But I'm not sure it's needed now anyway since even if I get it working out for strings with NUL bytes the problem is bison and flex will not. So it seems it's not that useful anyway. I don't know. Maybe for json it would be but at this point it's not.

xexyl:
But - I was not aware that it was used in other tools. What tools out of interest ?

PR #219, opened 2022-05-22
Add back bug fix in commit cab3cd304ff

xexyl:
In some functions the check if output was allowed was done prior to
    checking if that bool should be set to false. I believe this was done
    when I moved the check up to make it not do anything at all if it was
    not allowed

### DIFF
--- a/dbg.c
+++ b/dbg.c
@@ -330,6 +330,13 @@ warn(char const *name, char const *fmt, ...)
     va_list ap;		/* variable argument list */
     int saved_errno;	/* errno at function start */
 
+    /*
+     * silence warn() if -q and -v 0
+     */
+    if (msg_warn_silent == true && verbosity_level <= 0) {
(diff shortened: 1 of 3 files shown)


xexyl:
Just remembered that. Going again.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help - even if I hadn't originally caused it in the first place (but all the more reason to do so since I _did_).

xexyl:
Happy you got a laugh out of my comment! I'm going to go get some strawberries for dessert and then go to bed. Hope you have a great night! Will reply to comments tomorrow in the morning if you write any - and maybe you'll have added some code to look at with the parser (I hope so).

Good night (though I might be able to reply briefly in a bit - not certain though as I'm dead tired today)!

PR #220, opened 2022-05-23
Minor updates and fixes in json parser

xexyl:
Maybe can do more in a bit. I hope so though I don't know what yet.


lcn2:
Thanks @xexyl and best wishes on your resting.

xexyl:
> Thanks @xexyl and best wishes on your resting.

Thank you very much! Hopefully the fact I am sick in some way won't greatly affect my ability to work on this. But even if I am and it does ailments usually go through me fairly quickly (except the flu - when I could not get the vaccine due to shortage for example - but I'm asthmatic so that would explain that).

I'll be able to reply to messages as well but maybe not as enthusiastically. But it might be just as enthusiastic. Really depends on how I feel. I guess it's good I took my fever as it might explain why I was feeling hot and then cold and back and forth and also why I'm so exhausted - I probably am sick. But I'll be okay. Zoom meeting starting now. Good day!

PR #221, opened 2022-05-23
Various bug fixes and updates to do with json

xexyl:
Primarily an unhandled field was added to the check function (the no-comment one). There was also a problem in that the Makefile did not have for jinfochk and jauthchk the dependency `json_chk.o` so updating `json_chk.c` would not result in recompilation of those tools. 

Those are the bugs I refer to but perhaps JSON should itself be called a bug or maybe a virus. Probably not as I don't know that I would call it efficient (but I might be a tad bit biased). Well ... just something to think about maybe.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -413,7 +413,7 @@ jauthchk.o: jauthchk.c jauthchk.h json_util.h Makefile
 	${CC} ${CFLAGS} jauthchk.c -c
 
 jauthchk: jauthchk.o rule_count.o json_parse.o json_entry.o dbg.o util.o json_util.o \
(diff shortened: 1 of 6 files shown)


xexyl:
That's the last change for the day. There was also a typo fix in `jparse.y` debug messages. Off to do something else again.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help my friend! Hope you're having a nice sleep. I wish I was sleeping but at least (I think) I'm feeling better today. Hopefully I'm all better but I'll still be careful for a day or two depending on how I feel.

PR #223, opened 2022-05-25
Better JSON error messages for invalid JSON

xexyl:
parse_json() now reports if text/document is not valid JSON at the end
of the function. This function does not yet free the node (or any
children nodes) but yet but this will happen in time. The text is given
in example output below for the other change.

Add to grammar the JSON_INVALID_TOKEN as part of the json_element with a
call to UGLY_ABORT (which is #defined in jparse.h as bison doesn't
honour the prefix for YYABORT) so that the diff looks like:

     json_value:

	 json_object
	 {
    @@ -644,10 +643,15 @@ json_element:
	    json_dbg(JSON_DBG_MED, __func__, "under json_element: returning $json_element type: %s",
					     json_element_type_name($json_element));
	    json_dbg(JSON_DBG_LOW, __func__, "under json_element: ending: "
					     "json_element: json_value");
	 }
    +    |
    +    JSON_INVALID_TOKEN
    +    {
    +       UGLY_ABORT;
    +    }

This means when the lexer encounters an invalid character the parser
won't have to do as much (at least in some cases: I've only tested
simple cases so maybe this doesn't always hold) to do. For example:

    $ ./jparse -s '<'
    JSON DEBUG[1]: Calling parse_json("<", 1, NULL):
    JSON DEBUG[1]: *** BEGIN PARSE:
    <
    <
    >
    Starting parse
    Entering state 0
    Stack now 0
    Reading a token
    --accepting rule at line 215 ("<")
    JSON parser error on line 1:
    invalid token: 0x3c = <<>

    Next token is token JSON_INVALID_TOKEN ()
    Shifting token JSON_INVALID_TOKEN ()
    Entering state 8
    Stack now 0 8
    Reducing stack by rule 19 (line 649):
       $1 = token JSON_INVALID_TOKEN ()
    Stack now 0
    JSON DEBUG[1]: *** END PARSE

    not valid JSON

The last line was added in this commit as well in parse_json()
(see jparse.l).

The previous version had a lot more output. There still is a lot of
debug output but in this case the very first token is invalid so the
parsing is terminated.


lcn2:
Thanks @xexyl.  Lost internet connection in our general region. May be a while before we get back.

xexyl:
> Thanks @xexyl. Lost internet connection in our general region. May be a while before we get back.

Good luck with that! That would be very difficult for me personally. Hope you're having a nice sleep my friend!

PR #225, opened 2022-05-27
Add hack for better json error messages + minor comment updates

xexyl:
An interesting thing I pointed out in commit https://github.com/ioccc-src/mkiocccentry/commit/3c65ca0106c91bcd0b9181ebe4377105eee12e52:

```
    Rename 'JSON_INVALID_TOKEN' to be just 'token' for better error
    messages:
    
        [...]
         * Now if bison has an actual reference to the token value itself this would be
         * ideal to pass to the error message but if it does I'm unaware of it and
         * unfortunately UGLY_ABORT is specific to bison and not in flex. That being
         * said it does have a feature for more custom error functions and this can be
         * looked into.
         */
```

I will look into this functionality more. It doesn't appear to have the token itself but it might have something better than what we have here. I'm not sure. I will look into it in the near future. I'd like to do it now but I'm much too tired for that I'm afraid. Good day!

### DIFF
--- a/jparse.l
+++ b/jparse.l
@@ -154,8 +154,12 @@ JSON_COMMA		","
 			    /*
 			     * Whitespace
 			     *
-			     * Not needed but included for now for debugging
-			     * purposes.
+			     * NOTE: We have to include this action as otherwise
+			     * the action '.' will return an invalid token to
+			     * the parser.
+			     *
+			     * We don't need the below message but for debugging
+			     * purposes we have it printed for now.
 			     */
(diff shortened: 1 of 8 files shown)


lcn2:
nice hack, @xexyl on the token rename.

xexyl:
> nice hack, @xexyl on the token rename.

Thank you! I'll look into the more custom error functions another time. Going afk a bit and then when I'm back I'll probably do something else. Good day and good luck on the Internet!

xexyl:
Heading off for the day now. Will do more tomorrow I'm sure.

Enjoy the rest of your day and good luck with the Internet connectivity!

PR #227, opened 2022-05-29
Various updates and fixes (including errno fixes)

xexyl:
Perhaps the most interesting commit is https://github.com/ioccc-src/mkiocccentry/commit/6d1e2d8f948b0b32e9bef441ff2b116f8da62bcf:

    errno errors in json_entry
    
    There were a number of places in the file where errno was set to 0 prior
    to a write operation but warn() was used instead of warnp().
    
    There were some cases where errno was not set to 0 prior to using
    warnp() for errors.
    
    Also (a more subtle issue) in json_fprintf_str(): if json_putc() fails it
    used warnp() but without setting errno to 0. But json_putc() only
    returns false in two conditions: (0) when a NULL pointer is passed to it
    (which means errno is not relevant unless we set it to EINVAL or
    something like that ahead of time); and (1) when writing to the stream
    fails. In the latter case warnp() is actually used. It's true that
    warnp() saves the errno but the warning message in json_fprintf_str()
    for this case would be misleading as it would actually result in:
    
        called with NULL arg(s)
        json_putc #0 error
    
    The reason is that prior to calling json_putc() we KNOW that both *p !=
    '\0' && stream != NULL which means the only way json_putc() would fail
    is if there was an error with fprintf().
    
    Now one could argue that an additional output operation might be added
    to the function (or the checks for both NULL and NUL were removed) but
    if we set errno to 0 and get an error for a NULL pointer passed in we
    would get 'Success' as errno would still be 0.
    
    So should warnp() actually be called? I'm not sure it should so I've
    changed it to use warn() instead. Nevertheless one could certainly
    change errno to initially be EINVAL or some other error condition but it
    might still result in some unnecessary or confusing warning messages.

I'm going to do something else now and I don't know that I'll do anything else here today. These I noticed but the figuring out bison error locations will take a bit more effort and energy which I simply do not have. But we shall see what I do or don't do later. Good day my friend!

### DIFF
--- a/jauthchk.1
+++ b/jauthchk.1
@@ -1,4 +1,4 @@
-.TH jauthchk 1 "16 March 2022" "jauthchk" "IOCCC tools"
+.TH jauthchk 1 "29 May 2022" "jauthchk" "IOCCC tools"
 .SH NAME
 jauthchk \- IOCCC tool to validate .author.json file found within an entry directory
 .SH SYNOPSIS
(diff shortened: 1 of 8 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help!

PR #233, opened 2022-06-02
Many dbg.{c,md} updates and bug fixes (dbg.c, txzchk.c) + other fixes and updates

xexyl:
The changes are for consistency (but see below) as well as typo fixes
and trying to reduce the length of the text (removing some redundancy).
Some points of interest:

In most cases the texts describing what happens if fmt == NULL or name
== NULL have been moved to the top of the dbg.md to remove duplicate
text. The only difference between each instance was the function name
but I used 'foo' and put below that 'foo' is whichever function is being
used.

Reworded some things to help make said things clearer.

Added missing words in some sentences.

In the example code I changed the VERSION_STRING to be the current date
(or actually 1 June as I made that change yesterday). This is because
the date referred to 2021 and many improvements have been made since
2021. Maybe when everything is in order the date should be updated to
the date the texts are all in order? I don't know.

I got rid of some superfluous commas. More can be done here. However I
have tried to keep the format the same as the original document. If I
encounter any Oxford commas I'll try keeping those in as well even
though it goes against my nature. Important point is I think commas are
overused in much text out there but I am trying to keep them in here to
match the original style.

There's much left to do. I've not yet done a full pass in either file
and this will probably take several complete passes.

As for consistency: some of the things I have updated for this are not
entirely consistent as I have not gone through either file completely
once. I wanted to make this commit to have another clean start.

### DIFF
--- a/dbg.c
+++ b/dbg.c
@@ -2419,7 +2419,7 @@ werr(int error_code, char const *name, char const *fmt, ...)
     /*
      * stage 0: determine if conditions allow function to write, return if not
      */
-    if (err_output_allowed == true) {
+    if (err_output_allowed == false) {
 	return;
     }
 
@@ -2490,7 +2490,7 @@ vwerr(int error_code, char const *name, char const *fmt, va_list ap)
     /*
      * stage 0: determine if conditions allow function to write, return if not
      */
-    if (err_output_allowed == true) {
+    if (err_output_allowed == false) {
 	return;
     }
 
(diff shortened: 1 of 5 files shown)


xexyl:
> Excellent edits and changes, @xexyl and good catches too. Thanks!

Thank you! It's all I'll be doing today I'm afraid. Despite commit https://github.com/ioccc-src/mkiocccentry/pull/233/commits/37bc4fd0d0a578c2c50055b4a8888e8c1aa1a6cf the question I posed in the other thread about behaviour of `warn_or_err` still stands.

Any luck with the Internet access?

xexyl:
...as for that commit I mentioned `werr()` but I actually did mean `warn_or_err`. I'm not sure that there's a problem with `werr()` though perhaps the same question should be asked if that kind of check is there. I don't have it in front of me though to be sure.

PR #236, opened 2022-06-03
Fix warnp call in txzchk where errno is not valid

xexyl:
In this place it was if fnamchk fails but errno is not relevant there:
fnamchk will return an error code but it's not a C function error.
Instead use warn() rather than warnp().


xexyl:
I might end up doing some on `dbg` related things today but not for now ... have other things I need to do. I hope to do some in a bit but this is far from certain I'm afraid. If I don't today I'm sure I will tomorrow.

lcn2:
This is a subtile but valid point.  The call should have been a `warn()` not a `warnp()` call.

xexyl:
> This is a subtile but valid point. The call should have been a `warn()` not a `warnp()` call.

Yes. Probably a typo or brief brain lapse.

Issue #235, opened 2022-06-03
Enhancement: Write a test suite for `txzchk`


xexyl:
I just wrote the following to help start this but I also have some additional thoughts to add that might be considered to make it easier to test the tool. It would require that I make some changes to the tool (and possibly `fnamchk`) but I'm okay with that (though maybe you would want to do the changes to `fnamchk`):

--

# `test_txzchk`: text files for a test suite for `txzchk`

The subdirectories in this directory have text files for a test suite for the
txzchk tool.

The `good/` subdirectory has files that `txzchk` MUST report as valid.

The `bad/` subdirectory has files that `txzchk` MUST report as invalid.

The files MUST end with the extension `.txt` and the filename MUST follow the
rules of the `fnamchk` tool.

The typical format is that which would be generated from `tar -tJvf` but
importantly these files are **NOT** tarballs: they are text files that would
have the output of the tar command. This is to make it much easier to add test
cases and it also prevents having the need to have tarballs in the repo (as well
as constructing said tarballs with invalid input that might be submitted due to
abuse).

Here's an example good file:

```
drwxr-xr-x  0 501    20          0 Jun  3 05:50 test-0/
-rw-r--r--  0 501    20       1854 Jun  3 05:50 test-0/Makefile
-rw-r--r--  0 501    20          4 Jun  3 05:50 test-0/extra2
-rw-r--r--  0 501    20          0 Jun  3 05:50 test-0/prog.c
-rw-r--r--  0 501    20       2793 Jun  3 05:50 test-0/.author.json
-rw-r--r--  0 501    20       3715 Jun  3 05:50 test-0/remarks.md
-rw-r--r--  0 501    20       1429 Jun  3 05:50 test-0/.info.json
-rw-r--r--  0 501    20          4 Jun  3 05:50 test-0/extra1
```

Let's say the file is called `entry.test-0.1654260612.txt`. We can run the
following command to test `txzchk` on it:

```sh
$ ./txzchk -T -E txt entry.test-0.1654260612.txt
```

This will report on whether or not the tarball is valid. If you want to not see
any output at all and only check the exit status one can do:

```sh
$ ./txzchk -T -E txt entry.test-0.1654260612.txt 2>&1 >/dev/null || echo invalid
```

This will not show anything as the tarball is actually valid.

xexyl:
There are some things to consider though. I'll start out with the most important one: `fnamchk` and how `txzchk` uses it.

[...]

PR #240, opened 2022-06-05
Add bad test files in test_txzchk/bad and various updates and fixes

xexyl:
Add the following test files to test_txzchk/bad (txzchk must report at
least one issue for each file - which it does and txzchk_test.sh passes
all tests):

    entry.test-0.1652598666.txt
    entry.test-0.9876543210.txt
    entry.test-1.9876543210.txt
    entry.test-10.9876543210.txt
    entry.test-2.1652598666.txt
    entry.test-2.9876543210.txt
    entry.test-3.1652598666.txt
    entry.test-3.9876543210.txt
    entry.test-4.1652598666.txt
    entry.test-4.9876543210.txt
    entry.test-5.1652598666.txt
    entry.test-5.9876543210.txt
    entry.test-6.1652598666.txt
    entry.test-6.9876543210.txt
    entry.test-7.9876543210.txt
    entry.test-9.9876543210.txt

These files are based on the tarballs that were used during development
of txzchk to make sure all issues that txzchk had to detect were in fact
detected. The only difference is that they're now text files and the
timestamp was updated as the MIN_TIMESTAMP has been updated since the
time of the creation of txzchk.

As for the file:

    entry.test-10.9876543210.txt

A point of interest might be that fnamchk will fail so txzchk will
report that the directory name is indeed wrong. You would see an error
like so:

    ERROR[6]: main: second '-' separated token length: 7 != 6: ./test_txzchk/bad/entry.test-10.9876543210.txt

This is because of:

    #define MAX_ENTRY_CHARS (1)             /* characters that represent the maximum entry number */

which is used in fnamchk as:

        if (len != LITLEN("test-")+MAX_ENTRY_CHARS) {
	    /* ... */
	}

and

        if (len != UUID_LEN+1+MAX_ENTRY_CHARS) {
	    /* ... */
	}

Whether fnamchk should print the directory name at this point and then
return an error is I suppose up to discussion: it would allow for
testing more but it might be unnecessary as well given that a badly
named tarball would presumably be disqualified. Even so it might help
uncover some types of problems with txzchk.

TODO add even more tarball listings. I am pretty sure I have some that I
created as well but more can be contrived as well (one or more of the
above might actually be some I created myself but all the invalid
tarballs that were used for development are also included).

As for the typo fix: in txzchk_test.sh it was in a variable name:
EXIT_CODE_FIlE -> EXIT_CODE_FILE.


xexyl:
An interesting question for you in the log, Landon:

```
Whether fnamchk should print the directory name at this point and then
return an error is I suppose up to discussion: it would allow for
testing more but it might be unnecessary as well given that a badly
named tarball would presumably be disqualified. Even so it might help
uncover some types of problems with txzchk.
```

What do you think?

xexyl:
> An interesting question for you in the log, Landon:
> 
> ```
> Whether fnamchk should print the directory name at this point and then
> return an error is I suppose up to discussion: it would allow for
> testing more but it might be unnecessary as well given that a badly
> named tarball would presumably be disqualified. Even so it might help
> uncover some types of problems with txzchk.
> ```
> 
> What do you think?

Of course depending on where the failure (or failures) is (are) it would be impossible to do this as it uses `strtok`. But in some cases it might be possible? Perhaps it's not really useful though. I will leave that to you.

xexyl:
Heading off for the day. More from me tomorrow. Good day!

lcn2:
> An interesting question for you in the log, Landon:
> 
> ```
> Whether fnamchk should print the directory name at this point and then
> return an error is I suppose up to discussion: it would allow for
> testing more but it might be unnecessary as well given that a badly
> named tarball would presumably be disqualified. Even so it might help
> uncover some types of problems with txzchk.
> ```
> 
> What do you think?

Either way is fine.  The overall `make test` is not rather verbose so having `fnamchk` print the directory name is reasonable.

lcn2:
Thanks for the pull request, @xexyl .

Do you intend to write more tests, or is issue #235 far along along to call it closed?

xexyl:
> Thanks for the pull request, @xexyl .
> 
> Do you intend to write more tests, or is issue #235 far along along to call it closed?

I think I'll be adding more cases but I'm not sure that means that the issue has to be open still. I mean we also have a test suite for the `jinfochk` and `jauthchk` which is closed and we still have added cases so I will go ahead and close it after this.

As for adding test cases I might do the today but not sure. It's a very slow day. I slept in till almost 4 which is great but I'm feeling very fatigued and I also got some sad news that a good friend of mine lost his auntie to Alzheimer's - she died in the last 12 hours or so. 

Most weeks I only have Mondays that are busy at 11 (for about 45 minutes) (and I do this week too) but this week I have three other doctor appointments as well so it might be slower this week. I also have another doctor appointment next week and I'll be discussing surgery that day. I fear it has come to the point where it'll be required and it would remove some stress to just get it over with. 

Anyway I'll go ahead and close that issue. I saw that you were awake a couple hours or maybe few hours ago so I hope you're sleeping by now!

PR #249, opened 2022-06-10
Typo fixes

xexyl:
In run_bison.sh and run_flex.sh in add_sorry it said 'expects 1 args'
instead of 'arg'.

Why not change it to 'one'? It might be up to debate but the reason I
did not is because $# shows digits and not words so I left it as digits
too to keep it matching.

In util.h: I moved a comment above

    -/* for string to int functions */
    -#define LLONG_MAX_BASE10_DIGITS (19)
    +#define LLONG_MAX_BASE10_DIGITS (19) /* for string to int functions */

because there are lines right below that are not related to string to
int but might be confused as being related in some way or another.

### DIFF
--- a/run_bison.sh
+++ b/run_bison.sh
@@ -498,7 +498,7 @@ add_sorry() {
     # parse args
     #
     if [[ $# -ne 1 ]]; then
-	echo "$0: ERROR: add_sorry function expects 1 args, found $#" 1>&2
+	echo "$0: ERROR: add_sorry function expects 1 arg, found $#" 1>&2
 	exit 13
     fi
     FILE="$1"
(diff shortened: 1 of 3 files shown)


xexyl:
This is the last for the day I believe. I can reply to comments but as far as changes go I don't see anything else happening today.

lcn2:
Good catches, @xexyl

xexyl:
> Good catches, @xexyl

Thank you. I just got home. Will reply to the comments in the other threads then doing something else for the day. I'll be going to sleep earlier tonight as I was awake later last night (and the shipment did not include everything that was supposed to be there even which makes it all the more frustrating and so it'll be another night of this in the near future).

PR #251, opened 2022-06-11
Add initial hostchk.sh

xexyl:
Currently it does not parse options and there are only two checks:

Checks that tar with the required options is available.

Test compiles a C program with each of the system header files included
in all the C and header files in the repo. See below for how this is
done.

Rather than have to update the script each time a header file is removed
or added it uses a pipeline of grep and sort in a while read loop. The
pipeline itself is:

    grep -h -o '#include.*<.*>' ./*.c ./*.h|sort -u

and the top and bottom of the loop is:

    while read -r h; do

    ...

    done < <(grep -h -o '#include.*<.*>' ./*.c ./*.h|sort -u)

But instead of creating a C file for each header file used we use some
magic with the compiler and stdin. Thus the part that compiles each test
file is:

    printf "%s\n%s\n" "$h" "int main(void){ return 0; }" | cc -x c - -o "$PROG_FILE"

Thus there's no need to write to the actual disk as it's all stdin and
stdout.

To run this script use make hostchk or make bug-report.

At the risk of stating the obvious this is not complete.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,7 @@ H_FILES= dbg.h jauthchk.h jinfochk.h json_parse.h jstrdecode.h jstrencode.h limi
 DSYMDIRS= $(TARGETS:=.dSYM)
 SH_FILES= iocccsize_test.sh jstr_test.sh limit_ioccc.sh mkiocccentry_test.sh json_test.sh \
 	  jcodechk.sh vermod.sh prep.sh run_bison.sh run_flex.sh reset_tstamp.sh ioccc_test.sh \
-	  jparse_test.sh txzchk_test.sh
+	  jparse_test.sh txzchk_test.sh hostchk.sh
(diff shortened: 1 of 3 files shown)


xexyl:
Very possibly all I do today... I'll discuss it more at the other thread and then leaving for a break.

lcn2:
Thanks @xexyl 

We will try am merge this with our ongoing mods to `json_util.c` and `json_util.h` related to JSON parse tree printing.

xexyl:
> Thanks @xexyl
> 
> We will try am merge this with our ongoing mods to `json_util.c` and `json_util.h` related to JSON parse tree printing.

Well it looks like it worked. I'll be leaving shortly as.

Enjoy the rest of your day! Hopefully tomorrow I can do more.

PR #255, opened 2022-06-17
Change lambda transliterations in utf8_posix_map.c + update .gitignore

xexyl:
According to OED lambda is:

    the eleventh letter of the Greek alphabet (Λ, λ), transliterated as ‘l’.

so it should be capital (or uncial; majuscule) and lower case (or
minuscule) ELL and not y and n. Thus:

    -    { "\xc6\x9b", "y" , -1, -1},	/*  U+019B - ƛ - LATIN SMALL LETTER LAMBDA WITH STROKE */
    +    { "\xc6\x9b", "l" , -1, -1},	/*  U+019B - ƛ - LATIN SMALL LETTER LAMBDA WITH STROKE */

    -    { "\xce\x9b", "n" , -1, -1},	/*  U+039B - Λ - GREEK CAPITAL LETTER LAMDA */
    +    { "\xce\x9b", "L" , -1, -1},	/*  U+039B - Λ - GREEK CAPITAL LETTER LAMBDA */

    -    { "\xce\xbb", "y" , -1, -1},	/*  U+03BB - λ - GREEK SMALL LETTER LAMDA */
    +    { "\xce\xbb", "l" , -1, -1},	/*  U+03BB - λ - GREEK SMALL LETTER LAMBDA */

Now the question of the Latin version. As best I could find this is not
even actually a thing other than in Unicode. I believe the Greek lambda
created the Latin L though so it seems appropriate that the change be
made for the Latin version too.

Also I fixed the spelling 'lamda' to 'lambda'. Apparently Unicode uses
'lamda' for Greek and lambda for Latin which is both ironic and
ridiculous as lambda is the actual letter in the Greek alphabet and it's
spelt lambda not 'lamda' as well.

If there's a 'Latin capital letter lambda with stroke' it probably
should be changed to 'L' as well. However I don't see it in the table so
I'm guessing there isn't anything left to do here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # mkiocccentry specific
 .exit_code.*
 .sorry.*
+.hostchk.*
 answers.txt
 build.log
 chkauth
--- a/ioccc_test.sh
+++ b/ioccc_test.sh
@@ -50,6 +50,7 @@ Exit codes:
     8    ./dyn_test not found or not executable
     9	 ./jparse_test.sh not found or not executable
     10	 ./json_teststr.txt not found or not readable
+    11	 ./txzchk not found or not executable
 
     >=20  some test failed"
 export TEST_VERSION="0.462022-04-23"
@@ -197,6 +198,19 @@ fi
 if [[ ! -r json_teststr.txt ]]; then
     echo "$0: ERROR: json_teststr.txt is not readable" 1>&2
     exit 10
+fi
+# txzchk
+if [[ ! -e ./txzchk ]]; then
+    echo "$0: ERROR: ./txzchk file not found" 1>&2
+    exit 11
+fi
+if [[ ! -f ./txzchk ]]; then
+    echo "$0: ERROR: ./txzchk is not a regular file" 1>&2
+    exit 11
+fi
+if [[ ! -x ./txzchk ]]; then
+    echo "$0: ERROR: ./txzchk is not executable" 1>&2
+    exit 11
 fi
 
 
--- a/utf8_posix_map.c
+++ b/utf8_posix_map.c
@@ -516,7 +516,7 @@ struct utf8_posix_map hmap[] =
     { "\xc6\x98", "k" , -1, -1},	/*  U+0198 - Ƙ - LATIN CAPITAL LETTER K WITH HOOK */
     { "\xc6\x99", "k" , -1, -1},	/*  U+0199 - ƙ - LATIN SMALL LETTER K WITH HOOK */
     { "\xc6\x9a", "l" , -1, -1},	/*  U+019A - ƚ - LATIN SMALL LETTER L WITH BAR */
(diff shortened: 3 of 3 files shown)


xexyl:
I don't expect I'll do anything else today but that's uncertain yet. What is certain is I'm doing something else now (or so I think :) ).

Hope you're having a nice sleep my friend!

lcn2:
Thanks for checking and improving the `utf8_posix_map.c` @xexyl. 

Improving and expanding this map is a good thing to do: all be it a low priority thing at this time. 

This map is a rather unique map that attempts to find a POSIX safe character that "looks somewhat like" a UTF8 character. As such I this a judgement call instead of an official Internationalization.

FYI: a JSON file containing information about a past IOCCC winner (not the `.author.json` file but a new JSON file that will be used to build the new IOCCC web site) needs to use POSIX safe characters in the filename. That file needs to be kind of based on the winner's name, and the winner's name could have UTF8 characters in it. So we need a systematic and approximate mapping from UTF-8 to POSIX safe characters. That's the table will have other uses later on in building the IOCCC website.

xexyl:
> Thanks for checking and improving the `utf8_posix_map.c` @xexyl.
> 
Happy to help!

> Improving and expanding this map is a good thing to do: all be it a low priority thing at this time.
> 

I imagine it would be low priority. It's my love for words that made me discover the problem.

> This map is a rather unique map that attempts to find a POSIX safe character that "looks somewhat like" a UTF8 character. As such I this a judgement call instead of an official Internationalization.

Makes sense to me.

> 
> FYI: a JSON file containing information about a past IOCCC winner (not the `.author.json` file but a new JSON file that will be used to build the new IOCCC web site) needs to use POSIX safe characters in the filename. That file needs to be kind of based on the winner's name, and the winner's name could have UTF8 characters in it. So we need a systematic and approximate mapping from UTF-8 to POSIX safe characters. That's the table will have other uses later on in building the IOCCC website.

Will the same code or map be used or is it something else that has to be constructed as well?

Hope you have a great day!

lcn2:
> Will the same code or map be used or is it something else that has to be constructed as well?

Definitely yes, when it comes time to re-factoring the IOCCC website a tool to build information based upon JSON files (related to win the entries, and related to winning authors) will be needed.

lcn2:
Best wishes on your day, getting sleep, and your upcoming doctors appointments.

xexyl:
> > Will the same code or map be used or is it something else that has to be constructed as well?
> 
> Definitely yes, when it comes time to re-factoring the IOCCC website a tool to build information based upon JSON files (related to win the entries, and related to winning authors) will be needed.

The website is something I'm quite curious about - how it'll look and feel. I'm curious about this but I'll have to wait and see I suppose.

xexyl:
> Best wishes on your day, getting sleep, and your upcoming doctors appointments, @xexyl.

Thank you very much. I'll be leaving for the day shortly. As for doctors (besides online and phone) the next one will likely be pre-op and then surgery (still no date scheduled) and then after that probably post-op check. I will have to have another blood draw somewhere in there as well.

I hope you have a good rest of your day! I'll reply to the other things another time.

PR #258, opened 2022-06-20
JSON parser updates (typos + add string to json_teststr.txt) and hostchk.sh + txzchk_test.sh updates

xexyl:
The more important commits are https://github.com/ioccc-src/mkiocccentry/commit/7a13aa85d6ed9f7455d1609d90cb5edda8c5399c and https://github.com/ioccc-src/mkiocccentry/commit/093cd7320d251573fff70adf1ba9ef0aceb32467:

    Parse args in hostchk.sh
    
    The usage is now:
    
        $ ./hostchk.sh -h
        usage: ./hostchk.sh [-h] [-v level] [-D dbg_level] [-t tar] [-c cc]
    
            -h                          print help and exit 2
            -v level                    set verbosity level for this script: (def level: 0)
            -D dbg_level        set verbosity level for tests (def: level: 0)
            -t tar              path to tar that accepts -J option (def: /usr/bin/tar)
            -c cc               path to compiler (def: /usr/bin/cc)
    
        exit codes:
            0 - all is well
            1 - at least one test failed
            2 - help mode exit
            3 - invalid command line
            >= 30 - internal error
    
    Notice that the default paths to tar and cc are set up by type -P prior
    to setting up to the USAGE variable string. Previously tar was always
    /usr/bin/tar and cc was just cc by itself (no path specified).
    
    Add firewall checks for tar and cc and update references to these
    commands to be "$TAR" and "$CC" instead of tar and cc.
    
    Much more has to be added to this script as well as (the still
    non-existent) bug-report.sh script.

and

    Don't hard code default tar path in txzchk_test.sh
    
    Instead use $(type -P tar) as the default. This variable is used in the
    USAGE string as well.

Still lots to do with `hostchk.sh` (and have to add the `bug-report.sh` script as well) but there might have to be more discussion there and if nothing else I need to be more awake to do that. The same goes for `txzchk_test.sh`.

Will be taking a break now. Not sure if I'll do more later on. Hope you're having a nice sleep my friend!


xexyl:
As for https://github.com/ioccc-src/mkiocccentry/pull/258/commits/42073eab34e20ae48489005ed12c6383a2b5b71f it raises the question: should the fact the json parser can now parse arrays be added (and should the version be updated as well)? If so I'll let you do it though - as worrying about updating the test json file versions is too much for me right now.

This will be the last commit I do almost certainly. In a bit I plan on doing morning things and then I have two appointments today (one zoom, one phone). I'll be doing other things in between.

Good day!

lcn2:
Thanks for the improvements, @xexyl.

We have a number of domestic matters (common, not problematic) this week that are taking away from time to work on this repo.

xexyl:
> Thanks for the improvements, @xexyl.
> 
> We have a number of domestic matters (common, not problematic) this week that are taking away from time to work on this repo.

Thank you for the note. Good luck!

PR #260, opened 2022-06-21
Many typo fixes in util.c

xexyl:
Doing something else now and I don't expect to do any more here today. Hopefully I can be more productive tomorrow.

Good day!

### DIFF
--- a/util.c
+++ b/util.c
@@ -3132,7 +3132,7 @@ clearerr_or_fclose(char const *filename, FILE *file)
  *	line_len = fprint_line_buf(NULL, buf, len, 0, 0);
  *
  * given:
- *	stream	open stread to write to, or NULL ==> just return length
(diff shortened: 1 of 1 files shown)


lcn2:
Thank you @xexyl

xexyl:
> Thank you @xexyl

Sure. Something horrible has happened in the family .. it's not something I'll be wanting to think about as it brings back bad memories so I will just mostly be focusing on other things but it might make some things here difficult. But it might also be that it's a good distraction. 

Will be here a while but watching a film. Soon will get food and a while after that I'll be getting some sleep.

I'll look at the other threads another day. Too tired and overwhelmed to look at them today.

Good day!

xexyl:
Going now .. thank you again. Hopefully I can reply to at least some of the comments tomorrow. Good night.

PR #265, opened 2022-06-25
jparse_test.sh typo fix and behaviour modification

xexyl:
In the usage message the -D was described twice.

Instead of cognitive behavioural modification (CBT) we now have script
behavioural modification (SBT) :) as follows: The script now reads by
default the test suite text file json_teststr.txt. To read from stdin
use the Unix de facto standard -.

There is a reason for the above change in functionality: often I mean to
run the default tests but it blocks as it reads from stdin. Then I
remember that it default reads from stdin. However the test cases we
have used as we were developing the parser are in json_teststr.txt so
this seems a better default file to use. Since - signifies stdin in many
tools we use that for reading of stdin instead of defaulting to stdin.

### DIFF
--- a/jparse_test.sh
+++ b/jparse_test.sh
@@ -4,17 +4,18 @@
 
 # setup
 #
-export USAGE="usage: $0 [-h] [-v level] [-D dbg_level.sh] [-J level] [-q] [-p jparse] [file ..]
+export CHK_TEST_FILE="./json_teststr.txt"
+export USAGE="usage: $0 [-h] [-v level] [-D dbg_level.sh] [-J level] [-q] [-p jparse] [-d] [file ..]
 
     -h			print help and exit 2
     -v level		set verbosity level for this script: (def level: 0)
     -D dbg_level	set verbosity level for tests (def: level: 0)
     -J level		set JSON parser verbosity level (def level: 0)
     -q			quiet mode: silence msg(), warn(), warnp() if -v 0 (def: not quiet)
-    -D dbg_level	set verbosity level for tests (def: level: 0)
     -p /path/to/jparse	path to jparse tool (def: ./jparse)
 
-    [file ...]		read JSON documents, one per line, from these files (def: from stdin)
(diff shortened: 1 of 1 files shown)


xexyl:
I hope you enjoy the punning in the log :)

I don't know if I'll do anything else today .. pretty tired. I'm going to do something else now and I really hope I can rest in a bit. I know I need to get back to the `txzchk` test suite and also the additional scripts (the non-existent `bug-report.sh` as well as the very incomplete `hostchk.sh` script).

Hope you're having a nice sleep my friend and a good morning when you wake up!

lcn2:
Thank you for this useful update, @xexyl

xexyl:
> Thank you for this useful update, @xexyl
> 

Thank you.

> ### UPDATE 1:
> Remove this comment to [comment 1166364752](https://github.com/ioccc-src/mkiocccentry/issues/156#issuecomment-1166364752).

What about update 0 ? :) But thank you for the note about the moving of the comment. Will look.

lcn2:
> > Thank you for this useful update, @xexyl
> 
> > 
> 
> 
> 
> Thank you.
> 
> 
> 
> > ### UPDATE 1:
> 
> > Remove this comment to [comment 1166364752](https://github.com/ioccc-src/mkiocccentry/issues/156#issuecomment-1166364752).
> 
> 
> 
> What about update 0 ? :) But thank you for the note about the moving of the comment. Will look.

"Update 0" was the text of the original comment (that was moved). This is why "Update 1" simply states that the comment was moved. 🤓

xexyl:
> > > Thank you for this useful update, @xexyl
> > 
> > 
> > > 
> > 
> > 
> > Thank you.
> > > ### UPDATE 1:
> > 
> > 
> > > Remove this comment to [comment 1166364752](https://github.com/ioccc-src/mkiocccentry/issues/156#issuecomment-1166364752).
> > 
> > 
> > What about update 0 ? :) But thank you for the note about the moving of the comment. Will look.
> 
> "Update 0" was the text of the original comment (that was moved). This is why "Update 1" simply states that the comment was moved. 🤓

Ah. That makes more sense.

PR #266, opened 2022-06-26
Make jparse.1 consistent with recent developments

xexyl:
Removed from FILES section files that no longer exist.

Make the description of sorry.tm.ca.h more clear (it still could be
improved: it can be argued there is a bit of redundancy but this will
have to do for now).

Update BUGS to note that although jparse is in a sense finished the
parser does not return a tree (even though everything is parsed). I also
noted that better error reporting and various other things (those words
exactly as there are things that are not yet determined) need to be
added.

Added AUTHORS section.


xexyl:
Earlier on today I was hoping to look more at the txzchk test suite but I don't think that's going to happen. It's kind of an off day. Hopefully tomorrow I can do a bit more - on one or more things - but this is not certain either.

Hope you have a great rest of the weekend my friend!

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help!

PR #267, opened 2022-06-27
New test strings for jparse_test and JSON parse functions only spaces checks removed

xexyl:
These two commits might be all I do today .. not really sure yet. Too early literally and figuratively. Later in the morning I have pre-op Covid test and I have a zoom meeting later in the day as well. I've also been awake since just after 0000 so I'm dead tired. Going to try and rest some now actually.

See the individual commit logs for more details. I will mention one of them in the other thread as we discussed it yesterday but then I'm gone for some while I presume.


xexyl:
> Thank you for this effort @xexyl

Happy to help as always my friend!

Zoom meeting momentarily. After that I'll be getting some food and then take it easy the rest of the day. Maybe I'll be able to reply to the other comments (if there are any) at some point. Else tomorrow. Good day!

PR #273, opened 2022-07-06
Various updates and fixes

xexyl:


xexyl:
Well ... commit https://github.com/ioccc-src/mkiocccentry/pull/273/commits/9ded60145ac6fbf3902e31200bc1e1a836cd6d08 actually was the other thing - or more like were the other things as it turned out.

I wonder though. When doing this I came across a possible thing to change in mkiocccentry. Currently for writing to answers file it warns on errors writing to the file. Then at the end it reports how many errors there were and says that it will lead to problems should they make use of the file again if they want to update the entry. But should these errors be fatal instead? I'm not sure. At first when I added it I thought maybe warnings was the right way but now I'm not sure. What do you think ?

And this is really all I'll be doing today I'm almost certain of it. Good changes though but I have to take it easy and I'm having some additional problems with the recovery of the surgery. Good day!

lcn2:
Lots of good stuff in this pull request, thanks @xexyl

PR #274, opened 2022-07-07
Minor fixes and updates

xexyl:
Might end up doing more in a bit ... couldn't sleep but I think I might be able to at least lie down again now so I'll be doing that shortly. Hope you're having a nice sleep my friend!


lcn2:
Thanks @xexyl

PR #279, opened 2022-07-11
More fun with tar and feathers :-)

xexyl:
Instead of saying 'All checks passed.' txzchk now says 'No feathers
found in tarball.' if all tests pass.

Also the top comment now says more about what txzchk actually does (it
did not even state that it verifies that the tarball conforms to the
IOCCC tarball rules as it now does) but having fun with the idea of the
problems checked for are feathers.

### DIFF
--- a/txzchk.c
+++ b/txzchk.c
@@ -1,9 +1,11 @@
 /*
- * txzchk - IOCCC tarball validation check tool
+ * txzchk - IOCCC tarball validation tool
  *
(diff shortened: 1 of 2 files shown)


xexyl:
Perhaps you'll appreciate the comment changes as well.

That's all I'll be doing here until at least tomorrow. Good day my friend!

xexyl:
> LOL @xexyl

Glad you got a kick out of it as well!

PR #280, opened 2022-07-12
Typo fix / clarification in txzchk.1 DESCRIPTION + json parser comments updates

xexyl:
...and I'm afraid this is probably all I'll be able to do today. Busy day and unfortunately I woke up way too early today.

Hopefully tomorrow will be better and I can start making some real changes again.

Good day!

### DIFF
--- a/jnum_chk.c
+++ b/jnum_chk.c
@@ -4,29 +4,17 @@
  *
  * "Because the JSON co-founders minimalism is sub-minimal." :-)
  *
- * Copyright (c) 2022 by Landon Curt Noll.  All Rights Reserved.
+ * This JSON scanner was co-developed by:
  *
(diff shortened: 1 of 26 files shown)


xexyl:
> Hoping your recovery is progressing, @xexyl and that you get some periods of good rest.

Recovery is going well. I'm afraid I can't say the same for rest. Some days I am lucky enough to sleep in a bit but the last two days I was awake before 0100 and if I got back to sleep it was only for a little bit.

Still things feel like they're becoming more normal in function. I have a ways to go for sure but the last two days appear to be a bit better. I'll probably know in the next few days if this is the case or just feels like it but isn't.

I'm heading off for the day now I think. Hope you have a good rest of your day!

PR #283, opened 2022-07-19
Don't hard code patch path in jsemcgen.sh

xexyl:
Instead use type -P patch.

### DIFF
--- a/jsemcgen.sh
+++ b/jsemcgen.sh
@@ -28,6 +28,7 @@
 
 # setup
 #
+PATCH_TOOL="$(type -P patch)"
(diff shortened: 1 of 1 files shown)


xexyl:
Decided to do it quickly now. Going to sleep now. Enjoy the rest of your day!

lcn2:
Ok .. will look at this ...

xexyl:
> Ok .. will look at this ...

Of course one could say that it’s no longer the default default but rather a system default but whether that matters is another matter. 

I do think it’s better this way though in case the user has a strange system where the previous hard coded path is not the same on their system. 

Maybe this script isn’t meant for the normal users but I made this change in some other scripts too and it seems better this way. 

As I said in the other thread I don’t know that I will end up doing anything here today but hopefully I can tomorrow. 

Friday is very unlikely I will as I have the post op appointment. 

Good day my friend!

PR #288, opened 2022-07-28
Typo fixes and chkentry test functions added + bug fix txzchk

xexyl:


xexyl:
As per comment https://github.com/ioccc-src/mkiocccentry/issues/259#issuecomment-1198164241 and looking at the test functions (I looked at these after the comment) I think that I could indeed easily adapt my old code to the new way. I would guess that would be of a big help to you. I might be able to get to some of it today but I'm not sure. Shortly I'll be taking a break from the computer but depending on how I feel after that break I might be able to start.

Hope you're having a good morning (or sleep if you're still asleep)!

xexyl:
Okay well I got another test function in already .. I have to go afk now but when I'm back if I feel like I feel now I will work on more of these. As noted in the log I have not yet added the function that calls the test function but once that's added it can simply call the function and be done with it.

xexyl:
I think that's all I can do for now as I'm having a very hard time focusing my eyes. That should be a great help though I would think. I do have something to bring up that I noticed but it's probably better to have it in the other thread so I'll do that and then I'm off to do other things for the day.

Maybe I can feel more up to other tests to add later on today or tomorrow. I know the dbg readme still has to be sorted but I'm definitely too tired for that one.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help! I'll be shutting down now. Enjoy the rest of your day my friend and sleep well when you do!

PR #290, opened 2022-07-29
Various fixes and updates to txzchk and new test functions for chkentry

xexyl:
The following commits were made but due to conflicts in other commits I
am making a new pull request. Unfortunately it's harder to keep track of
what commit did what as I have it all in one commit now but the
following commits were made except that I have attempted to merge some
of the logs together. The commit ids will no longer be valid of course
but I'm doing it this way to try and keep it more organised (try but
maybe not succeed):

commit 7ceb5e60d15400dc02d4ea8875f1c516efcdd9fc
Date:   Fri Jul 29 07:55:11 2022 -0700

    Minor updates and fixes in txzchk

    Change remaining references of 'issues' to 'feathers' to continue the
    fun of:

        "Because sometimes people add feathers to tar." :-(

    (There are still a few references to issues but this is only to make it
    clear that it's just wordplay: all messages and variables now refer to
    feathers though.)

    I've also improved show_txz_info() in the following ways:

    It now shows the booleans as true or false.

    I have attempted (keyword) to make it all the debug information lined up
    (matching columns for names and values). This might not always work as
    \t is funny like that but for some tarballs it works fine at least. For
    example with the file test_work/entry.test-0.1659097955.txz the
    alignment is still not consistent for every line but for
    test_txzchk/entry.test-0.1658587455.txt (which is only on my system) it
    works fine. It might be a bit better though.

    I thought of making this a new version but that would require updating
    the version in the json files and the changes are only cosmetic anyway
    so it doesn't warrant a new version anyway.

    Minor updates in txzchk.1

commit e0e7c8e578b3ae8cbb7c731cbbc871a95d256aa2
Date:   Fri Jul 29 05:10:30 2022 -0700

    Check for multiple bogus fields in txzchk

    In other words a line like:

        -rw-r--r--  0 501    20          0 Jul 28 08:25 test-0/file bogus data

    will report (for this line specifically):

        Warning: txzchk: test_txzchk/entry.test-0.1658587455.txt: bogus field found after filename: bogus
        Warning: txzchk: test_txzchk/entry.test-0.1658587455.txt: bogus field found after filename: data

    Previously it would have only detected 'bogus'.

    A question that was brought up is whether or not the line should count
    as a filename of 'file bogus data' or if considering 'bogus' and 'data'
    as bogus fields is enough. On the one hand the most likely case is it
    would be a filename with spaces (unless of course the tarball was
    created by mkiocccentry); but on the other hand it might also be someone
    messing with the tarball and since strtok_r() is used (with " \t") it
    would require more code to detect and display the full filename and
    since it's also bogus this works (for now if not indefinitely).

### DIFF
--- a/entry_util.c
+++ b/entry_util.c
@@ -701,6 +701,197 @@ test_title(char *str)
 }
 
 
+
+/*
+ * test_empty_override - test if empty_override is valid
+ *
+ * Determine if empty_override boolean is valid.  :-)
+ * Well this isn't much of a test, but we have to keep
+ * up with the general form of tests!  :-)
+ *
+ * given:
+ *	boolean	boolean to test
+ *
+ * returns:
+ *	true ==> bool is valid,
+ *	false ==> bool is NOT valid, or some internal error
+ */
+bool
+test_empty_override(bool boolean)
+{
+    json_dbg(JSON_DBG_MED, __func__, "empty_override is %s", booltostr(boolean));
+    return true;
+}
+
+/*
+ * test_rule_2a_override - test if rule_2a_override is valid
+ *
+ * Determine if rule_2a_override boolean is valid.  :-)
+ * Well this isn't much of a test, but we have to keep
+ * up with the general form of tests!  :-)
+ *
+ * given:
+ *	boolean	boolean to test
+ *
+ * returns:
+ *	true ==> bool is valid,
+ *	false ==> bool is NOT valid, or some internal error
+ */
+bool
+test_rule_2a_override(bool boolean)
+{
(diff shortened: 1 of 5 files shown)


xexyl:
Hopefully that works. I can try merging the test functions another time. Not sure I have the energy to do that part now but at least the txzchk fixes can be in.

xexyl:
It occurred to me that it's simple to do. I can just yank paste the functions and commit. There wasn't anything else to do.


I might have made other fixes and the commit logs are not as useful as they were maybe but this should work I think.

Will be leaving soon. I'll close the other pull request as this seems the simpler way to go about it.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help as always my friend!

PR #297, opened 2022-08-03
More foo

xexyl:
I do have more quotes but I'm not sure how I want to present them yet so I'm holding off. This is a fun one though!

Unless I do more quotes I don't expect to do anything else here today. I am afraid that that might very well be the case but we shall see.


lcn2:
Foo :) @xexyl

xexyl:
> Foo :) @xexyl

It's a really funny show! That's an Easter egg itself .. the second 'English' subtitles has funny computer dialogues instead. Sadly only the first episode has this feature though.

Going afk shortly. Hope you have a great day!

PR #298, opened 2022-08-03
More foo

xexyl:
You'll get a sense of my humour in reading some of these. I even start to wonder myself If I'm a bit mad. 

Some of the comments are just silly, some are ridiculous, some clever, many funny (not all) and there probably are a few mean ones there too.

That's all I'm doing for the day. I do have another set of things to add but that can be tomorrow or the next day. Doing something else now.

Good day!


xexyl:
BTW: After merging this try enabling it to see a funny one (the thing is ridiculous but the comment I think is funny and it has some wordplay there too).

lcn2:
Gunaxf @krkly

xexyl:
> Gunaxf @krkly

Welcome. I would reply in like form but not at the laptop and more about to get ready to sleep so I won’t do it in my head either. 

Hope you have a great night! 

Hopefully tomorrow I can do more though there will be some activity here unfortunately which might or might not make that harder to do. 

Good night my friend!

PR #305, opened 2022-08-07
Make empty comments more empty in foo and foo more foo in foo

xexyl:
I hope that these make you laugh or if nothing else smile, Landon.

Have a great safe holiday my friend! I'll be adding to that strange array and maybe I can get to other things too.


lcn2:
Fun edit, @xexyl, thanks!

**IMPORTANT NOTE**:  We will very soon go into vacation mode, limited much of our work to offline activity, with an infrequent return to the repo and with very limited ability to answer questions / engage in dialogue for about a month.

With that in mind, consider the following challenge:

The following command line:

```c
$ tr '[a-m][n-z][A-M][N-Z]' '[n-z][a-m][N-Z][A-M]' < foo.c | less
```

is an easy way "spoil" the easter eggs on the `oebxergfB` array.

While the _ROT13_ decoder should remain in `foo.c`, consider applying an additional non-_ROT13_ [substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher) to further scramble the text in the  `oebxergfB` array.

As an example, consider the following keyword (where each letter appears only once and most of the UPPER and lower case letters are used) for a [simple substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher#Simple_substitution):

```
blacksmithgunpowderySTUMPEDFLYBACKWHORING
```

You might with to write a simple tool that takes input and outputs a C string that goes into a new `oebxergfB` array.  Keep the tool to yourself, don't add it to any repo.  Your tool would take text on stdin and:

1. perform _ROT13_
2. perform a [simple substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher#Simple_substitution) on a long keyword (where upper and lower case letters appear only once)
3. write to stdout, "foo\n" style lines.

Next add code to `foo.c` to apply the reverse [simple substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher#Simple_substitution) using the same keyword BEFORE the existing _ROT13_ code.

Finally, replace the `oebxergfB` array with the scrambled array using your private tool.

We do NOT recommend going overboard on the encryption.  Keep the screaming simple. 

If you do that, then the above shell command line won't allow a simple _ROT13_ to spoil the easter eggs.  Force someone who wants a spoiler to have to write code themselves.

xexyl:
> Fun edit, @xexyl, thanks!

Thank you as well!

> 
> **IMPORTANT NOTE**: We will very soon go into vacation mode, limited much of our work to offline activity, with an infrequent return to the repo and with very limited ability to answer questions / engage in dialogue for about a month.

Enjoy my friend - but be safe!

> 
> With that in mind, consider the following challenge:
> 
> The following command line:
> 
> ```c
> $ tr '[a-m][n-z][A-M][N-Z]' '[n-z][a-m][N-Z][A-M]' < foo.c | less
> ```
> 
> is an easy way "spoil" the easter eggs on the `oebxergfB` array.
> 

Good point though doing `less -r` would be better since there are some characters that won't be transcribed properly otherwise. I don't see that as giving an extra spoiler though. I am not sure if I ever realised you could have more than one range in `tr` like you did at least in the way you did it.


> While the _ROT13_ decoder should remain in `foo.c`, consider applying an additional non-_ROT13_ [substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher) to further scramble the text in the `oebxergfB` array.

Good idea! I guess it could be some other encryption too. 



> 
> As an example, consider the following keyword (where each letter appears only once and most of the UPPER and lower case letters are used) for a [simple substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher#Simple_substitution):
> 
> ```
> blacksmithgunpowderySTUMPEDFLYBACKWHORING
> ```
> 
> You might with to write a simple tool that takes input and outputs a C string that goes into a new `oebxergfB` array. Keep the tool to yourself, don't add it to any repo. Your tool would take text on stdin and:
> 
> 1. perform _ROT13_
> 2. perform a [simple substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher#Simple_substitution) on a long keyword (where upper and lower case letters appear only once)
> 3. write to stdout, "foo\n" style lines, ending with a final "...\n", line (to make it easy to add new messages to the `oebxergfB` array)
> 
> Next add code to `foo.c` to apply the reverse [simple substitution cipher](https://en.wikipedia.org/wiki/Substitution_cipher#Simple_substitution) using the same keyword BEFORE the existing _ROT13_ code.
> 
> Finally, replace the `oebxergfB` array with the scrambled array using your private tool.
> 
> We do NOT recommend going overboard on the encryption. Keep the screaming simple.
> 

Damn .. there goes my idea of bringing in my Enigma machine! :( (No .. definitely did not want to do that but the idea amuses me.) I agree to keep it simple.


> If you do that, then the above shell command line won't allow a simple _ROT13_ to spoil the easter eggs. Force someone who wants a spoiler to have to write code themselves.

Great idea. I'll see what I can come up with. I admit I'm having a lot of fun adding the quotes now so I might want to do that later - or maybe this earlier on. I do have a tool I'm using that does similar to what you say but without the extra cipher. 

This won't be done today or tomorrow though as I have people working in the house and the dogs are going barking mad. Well they should be.

Hopefully I can get the motivation to work on this idea sometime soon though. I'm making a reminder with the link to the comment so I don't forget.

We're hoping my cousin can visit sometime soon and if so that'll take some days to recover (though totally worth it) but we're not sure when that will be just yet. It might or might not happen before you're back.

Enjoy and stay safe my friend!

lcn2:
Going now @xexyl .. see the above command the we just modified after your remarks.

xexyl:
> Going now @xexyl .. see the above command the we just modified after your remarks.

Thank you.

Good day and month and stay safe! Have a great holiday and BCNU! :) Until we meet again!

xexyl:
> Good point though doing `less -r` would be better since there are some characters that won't be transcribed properly otherwise. I don't see that as giving an extra spoiler though. I am not sure if I ever realised you could have more than one range in `tr` like you did at least in the way you did it.

I know it doesn't matter but I see what you did with `tr` now. It's quite simple. Funny thing is I'm quite sleepy and I plan on trying to sleep again shortly. But it bothered me that I didn't put it together then so I have to state it now. It actually wasn't more than one range - it just looked like it at a tired and quick glance.

Hope you're having a great time! I'm hoping to look into this sometime soon. Actually I did start but I'm too sleepy to continue so I'm going to stop now and take a rest - hopefully a nap.

Until we meet again.

xexyl:
... Substitution cipher done. Just have to do the other parts now. The hardest part will be reconstructing the array but that should be doable too I just have to think about it when more awake. The backup drive is cool enough now where I can put it away and then I'm going to try going back to sleep.

Even though you won't see this anytime soon I thought you'd like to have it noted .. even t hough it'll probably be in before you even see it.

xexyl:
I'm kind of taking it easy whilst you're gone but I just made some fixes to the cipher. It now does everything properly. However it's not yet in the foo.c file. This will probably not happen today but it might happen over the weekend.

Meanwhile I hope you're having a great holiday!

xexyl:
I did it today after all. See: https://github.com/ioccc-src/mkiocccentry/pull/306 and https://github.com/ioccc-src/mkiocccentry/pull/306/commits/bb361300388f6bf094f23fd3b3b2121cad4ff08b.

I now have to extract the old strings so I have a copy of them always (though I do have a way to decrypt them all it is true).

Going to take it easy the rest of the day and I might not do anything over the weekend after all. WE shall see. I do have some thoughts on how to improve it slightly but I'm not sure if I'll bother with that.

BTW: do you understand the question I posed at comment https://github.com/ioccc-src/mkiocccentry/pull/306#issuecomment-1213340967 ?

PR #313, opened 2022-08-18
Fix 16beee5 for CentOS compilation

xexyl:
Because chkentry.c uses a function that is defined in foo.c chkentry.h
included foo.h. However this resulted in a compilation error in CentOS
as supposedly an array that was defined was not being used (this is
because it's used in foo.c but not chkentry.c but chkentry.c #included
foo.h for above mentioned function) i.e. it was used in one place but
not the other.  Removing static would not work because then it would
result in multiple definitions of the same symbol.

There were three fixes I could think of. One was discarded before as
undesirable (inclusion guards to enable certain variables) and another
would be a crime against foo so I opted with moving the function
declaration into chkentry.h since that's where it's used anyway.

Run make depend and fix typos in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,8 @@ TRUE= true
 # legacy targets #
 ##################
 
-# In a few cases, legacy systems such as Centos 7 that are pre-c11 cannot compile
-# this repo without extraordinary work-a-rounds / gross hacks.
+# In a few cases, legacy systems such as CentOS 7 that are pre-c11 cannot compile
+# this repo without extraordinary workarounds / gross hacks.
 #
 # While this repo does not officially support pre-c11 systems that have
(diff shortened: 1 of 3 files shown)


lcn2:
Thanks 🙏 @xexyl

xexyl:
> Thanks 🙏 @xexyl

I'm happy I can help. Afraid it's all I can do today though :( I'm so tired I can barely think. I doubt I'll manage to sleep but I might have a lie down at some point.

Hope you're having a nice day!

lcn2:
So far, things look good for macOS after this pull request.

xexyl:
> So far, things look good for macOS after this pull request.

I should hope so as I tested it and committed from it! :)

Going for the other idea I had as you thought it might be better .. will do another pull request if all is okay.

PR #315, opened 2022-08-18
More foo

xexyl:
This was done due to a merge conflict and I wanted to make sure everything works okay.

I hope tomorrow I can work on txzchk.

### DIFF
--- a/foo.c
+++ b/foo.c
@@ -1214,6 +1214,10 @@ static char const *oebxergfB[] =
     "\n"
     "-- Ywz Lqaknf",
 
+    "Fubasql Dzudw: Tnw eyarsm ismszudw osa tyqpnump nfwmyl; bnwf'tt tubwayttf\n"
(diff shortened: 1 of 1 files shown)


lcn2:
Thanks 🙏 @xexyl and hoping we are near the end of foo.

xexyl:
> Thanks 🙏 @xexyl and hoping we are near the end of foo.

Well I'd say let's say we're close to limiting foo to few fufus! 

I still want to make it so that it can be automatically generated and I still do have some quotes to add but I'll try limiting it now. I'm awake now but I'm hoping to go back to sleep in a bit as I've not had enough sleep.

I do have a doctor appointment today - surgery follow-up. Given that I've not had enough sleep as well I'm not sure (did not think of this yesterday) if I will get anything in today as far as say txzchk. We'll see however.

As for foo I have had an idea on how to help with my tool: the issues I mentioned in the other thread (as an OT point).

Hope you're having a nice day.

PR #316, opened 2022-08-19
Add three json test_ functions

xexyl:
    
        test_github()
        test_twitter()
        test_url()
    
    and the #if 0 .. #endif was removed for each one.
    
    As well since test_affiliation() already existed (which I only
    discovered after writing it) I removed the #if 0 .. #endif so that it's
    validated too.

...there were also some comment fixes in foo. 

This is probably all I'll do today: very tired and going to try sleeping again. Later like I said I have a surgery follow-up and so I am probably done here.

### DIFF
--- a/entry_util.c
+++ b/entry_util.c
@@ -940,7 +940,7 @@ object2author(struct json *node, unsigned int depth, struct json_sem *sem,
 	}
 	return false;
     }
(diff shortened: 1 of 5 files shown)


lcn2:
Thanks 🙏 @xexyl

xexyl:
> Thanks 🙏 @xexyl

Happy to help!

PR #317, opened 2022-08-20
Several fixes and enhancements to txzchk + formatting fix in foo

xexyl:
The following fixes were made (but note that if a test is not mentioned
below it does not mean it's not already in place and it's entirely
possible I miss below one or more tests that has or have been added).

After each file size is parsed if the new total size is < the previous
size it is an issue.

After each file size is parsed if the new total size is > MAX_DIR_KSIZE
it is an issue (this was only done at the end after all files parsed).

After each file size is parsed if the new total size is < 0 it is an
issue (the current file size was tested for < 0 and also 0 but the total
was not until the very end).

After each file size is parsed if the new rounded file size (up to 1024
multiple) is < the previous it is an issue.

Each test is only displayed once. That is there are new booleans that
disable showing a warning and also increasing the number of feathers
(issues) in the tarball but perhaps the total feathers should be
increased and only the message should not be. The reason I did it this
way is because it's simpler to keep track of the values and if for
example a 0 file size comes in between what should the message be?
Another reason I opted for this is it's less verbose as the tool is
already quite verbose without appropriate options (-q or -e but the
latter should be removed and will be eventually when the test script is
updated to show more details).

Perhaps all the checks made on the tarball should be added to the man
page but if so that's for another time.

I have not added any of these to the final report but perhaps they
should be.

Perhaps the booleans should be an unsigned int being the number of times
it occurs?

### DIFF
--- a/foo.c
+++ b/foo.c
@@ -61,9 +61,8 @@ static char const *oebxergfB[] =
     "ul twzuespayknueyttf twll bnym '>'. Usiwsmw leawxwd qk.\n"
     "-- Xwzft",
 
-    "C dsm'b hmsx nyto so fsq nyto yl xwtt yl C lnsqtd tuhw;\n"
-    "ymd C tuhw twll bnym"
-    " nyto so fsq nyto yl xwtt yl fsq dwlwavw.\n"
+    "C dsm'b hmsx nyto so fsq nyto yl xwtt yl C lnsqtd tuhw; ymd C tuhw twll\n"
+    "bnym nyto so fsq nyto yl xwtt yl fsq dwlwavw.\n"
     "-- Nutrs Nyppuml",
 
     "I lwebusm: '{}'\n"
--- a/txzchk.c
+++ b/txzchk.c
@@ -614,6 +614,140 @@ check_empty_file(char const *txzpath, off_t size, struct txz_file *file)
     }
 }
 
+/* convert_file_size	- converts number string to off_t
+ *
+ * This function is to remove duplicate code as this code is used in more than
+ * one function.
+ *
+ * given:
+ *	current_file_size   - pointer to off_t to store the file size in
+ *	p		    - the token extracted from strtok_r() (number
+ *			      string)
+ *
+ * Returns: true if conversion succeeded; false if it fails.
+ *
+ * NOTE: This function does not return on NULL pointers (which should actually
+ * never happen).
+ */
+static bool
+convert_file_size(off_t *current_file_size, char *p)
+{
+    /*
+     * firewall
+     */
+    if (current_file_size == NULL)
+    {
+	err(16, __func__, "passed NULL current_file_size");
+	not_reached();
+    }
+    else if (p == NULL)
+    {
+	err(17, __func__, "passed NULL number string");
+	not_reached();
+    }
+
+    errno = 0;
+    *current_file_size = strtoimax(p, NULL, 10);
+    if (errno != 0)
+	return false;
+
+    else if (*current_file_size < 0)
+    {
+	warn("txzchk", "%s: file size < 0: %jd", txzpath, (intmax_t)*current_file_size);
+	++txz_info.total_feathers;
+    }
+    txz_info.file_sizes += *current_file_size;
+
+    /* false: check new totals but don't print rounded file size if OK */
+    check_txz_files_size(false);
+
+    return true;
+}
+
+/* check_txz_files_size	    - check current total file size sum
+ *
+ * This function is used after each line is parsed. It's important because if a
+ * file takes one or both of the sizes (file_sizes or rounded_file_sizes) to >
+ * the max or < 0 or 0 then it's an issue that needs to be flagged. We have
+ * booleans so that we only warn of this once.
+ *
+ * given:
+ *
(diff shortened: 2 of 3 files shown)


xexyl:
What do you think of the questions/thoughts I posed, Landon? Which would you like? Anything else? I'll be trying to sleep again sometime soon I think. Hope you're having a great day my friend!

lcn2:
> What do you think of the questions/thoughts I posed, Landon? Which would you like? 

Sorry without context we don't know what you referred to.

On a remote setup with a cell phone, so we cannot easily scan nor easily go into detailed dialog for a few weeks.

xexyl:
> > What do you think of the questions/thoughts I posed, Landon? Which would you like?
> 
> Sorry without context we don't know what you referred to.
> 
> On a remote setup with a cell phone, so we cannot easily scan nor easily go into detailed dialog for a few weeks.

Ah. I thought you had your laptop too. Well the questions were in the log but maybe the most important one I can easily shorten without context (which only increases the length and might not be necessary here).

When the tool detects a problem - let's say that the current size < the previous size. I warn about this and increase the number of issues (feathers). But then I set a bool to say it's happened and don't trigger it again. Should it be triggered again though and warn each time?

Even if does increase the number of problems should it be reported? Alternatively if it does not increase the number of problems should it be reported?

Hopefully that helps?

lcn2:
> Ah. I thought you had your laptop too. 

Not for a few days

lcn2:
> But then I set a bool to say it's happened and don't trigger it again. Should it be triggered again though and warn each time?

Perhaps reporting it again isn't so bad sometimes?

It probably depends on the error. Reporting each time the sum of file sizes exceeds the limit isn't helpful. Reporting each time a negative file size is found is helpful. 

Use your judgment, @xexyl

xexyl:
> > Ah. I thought you had your laptop too.
> 
> Not for a few days. We wrote weeks but we should have written days.

Thank you for the note.

Hope you're having a great day!

xexyl:
> > But then I set a bool to say it's happened and don't trigger it again. Should it be triggered again though and warn each time?
> 
> Perhaps reporting it again isn't so bad sometimes?
> 
> It probably depends on the error. Reporting each time the sum of file sizes exceeds the limit isn't helpful. Reporting each time a negative file size is found is helpful.
> 
> Use your judgment, @xexyl

Well another idea I had is to report it once but still count it and then give a final report if verbosity level is high enough. That might be good. 

I should make it clear though: every time a negative file size is encountered it already is reported. The context for this question/thought was for the size of all files after each file was added. Does that change your answer?

lcn2:
> The context for this question/thought was for the size of all files after each file was added. Does that change your answer?

Perhaps once the total size has exceeded the total size limit, it can be reported once?  Not that reporting it more than is all that bad: if someone submits something too big, then multiple error messages is the least of their problem.  Nevertheless reporting total size has exceeded the total size limit and setting and flag seems reasonable.

xexyl:
> > The context for this question/thought was for the size of all files after each file was added. Does that change your answer?
> 
> Perhaps once the total size has exceeded the total size limit, it can be reported once? Not that reporting it more than is all that bad: if someone submits something too big, then multiple error messages is the least of their problem. Nevertheless reporting total size has exceeded the total size limit and setting and flag seems reasonable.

I think that's what I made it do yesterday. It could of course count it so that if a new file size drops the size down below the max - well of course that would all be abuse and would be disqualified but I like being complete. What do you think?

lcn2:
> > > The context for this question/thought was for the size of all files after each file was added. Does that change your answer?
> > 
> > 
> > Perhaps once the total size has exceeded the total size limit, it can be reported once? Not that reporting it more than is all that bad: if someone submits something too big, then multiple error messages is the least of their problem. Nevertheless reporting total size has exceeded the total size limit and setting and flag seems reasonable.
> 
> I think that's what I made it do yesterday. It could of course count it and report number of times so that if a new file size drops the size down below the max - well of course that would all be abuse and would be disqualified but I like being complete. What do you think?

Sure.

xexyl:
> > > > The context for this question/thought was for the size of all files after each file was added. Does that change your answer?
> > > 
> > > 
> > > Perhaps once the total size has exceeded the total size limit, it can be reported once? Not that reporting it more than is all that bad: if someone submits something too big, then multiple error messages is the least of their problem. Nevertheless reporting total size has exceeded the total size limit and setting and flag seems reasonable.
> > 
> > 
> > I think that's what I made it do yesterday. It could of course count it and report number of times so that if a new file size drops the size down below the max - well of course that would all be abuse and would be disqualified but I like being complete. What do you think?
> 
> Sure.

Meaning I should count the number of times and then report it depending on verbosity?

And unrelated I just did a pull request with some changes to foo including the key. But you'll see that in time.

If this is what you mean with txzchk though I'll get to it another time. Will make a note of it if that's what you mean?

lcn2:
> Meaning I should count the number of times and then report it depending on verbosity?

yes

xexyl:
> > Meaning I should count the number of times and then report it depending on verbosity?
> 
> yes

Just updated txzchk.h (the struct) so I'll do this in the coming days.

Doing something else now. Good day!

PR #318, opened 2022-08-21
Move oebxergfB to oebxergfB and expand oebxergfB

xexyl:
This required changing foo and foo plus the Makefile (make depend).


xexyl:
I did in fact change the keys so I'll have to send you an updated version which as I said elsewhere is also a bug fixed version. Should I wait until you're home?

And I'm afraid that's all I'll be doing for the rest of the day here. Going to take it easy. Enjoy the rest of your day!

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Thank you as well.

Please let me know when you want me to send the updated tool and I'll do that.

xexyl:
Oh another thing. I excluded the new header file from picky. Reason being is that there are diacritics in some of the strings.

Is there a way past this or is this okay to not have it pass picky?

xexyl:
Ah ... must be the option `-8` which means I can put it in the less picky files list. Right?

lcn2:
> Ah ... must be the option `-8` which means I can put it in the less picky files list. Right?

Yes

xexyl:
> > Ah ... must be the option `-8` which means I can put it in the less picky files list. Right?
> 
> Yes

I ended up doing it as I looked at the help. ...which I see you just noticed. Thanks.

PR #320, opened 2022-08-22
Add more test_ functions to entry_util + fix typos

xexyl:

    +extern bool test_remarks(char *str);
    +extern bool test_min_timestamp(time_t tstamp);
    +extern bool test_timestamp_epoch(char *str);

To clarify the difference between test_min_timestamp() with the already
existing test_formed_timestamp() (since it also refers to
MIN_TIMESTAMP): the test_formed_timestamp() function verifies that the
formed timestamp >= MIN_TIMESTAMP whereas test_min_timestamp() tests
that the entry has the same minimum timestamp required.

I also fixed some typos.

As well I added some defines:

    #define PROG_C_FILENAME "prog.c"
    #define REMARKS_FILENAME "remarks.md"
    #define MAKEFILE_FILENAME "Makefile"

and made use of these in the entry_util.c file. Perhaps these can also
be used in mkiocccentry.c but I only just thought of this and I must
leave for now (there was a typo fix in mkiocccentry.c however).

### DIFF
--- a/entry_util.c
+++ b/entry_util.c
@@ -1407,7 +1407,7 @@ test_Makefile_override(bool boolean)
 /*
  * test_abstract - test if abstract is valid
  *
- * Determine if abstract length is <= MAX_ABSTRACT_LEN.
+ * Determine if abstract length is > 0 && <= MAX_ABSTRACT_LEN.
  *
  * given:
  *	str	string to test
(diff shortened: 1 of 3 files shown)


xexyl:
I was going to do the test_no_comment but I started to have questions that probably should not be questions (besides the fact that the defines have since had changed values so maybe the function should be renamed?) which tells me that I am too tired to do anything more so I'm going to have a lie down. Maybe I can do more later .. though I think if I do it probably will be other things.

lcn2:
Thanks @xexyl and good timing and well merged!

As of this pull request, the following test functions need to be completed:

- test_name
- test_no_comment
- test_past_winner
- test_rule_2a_size
- test_rule_2b_size
- test_tarball

In addition to whatever test functions need to be applied to `mkiocccentry.c` code.

xexyl:
> Thanks @xexyl and good timing and well merged!
> 

Good timing how? Anyway thank you and I am happy to help!

> As of this pull request, the following test functions need to be completed:
> 
> * test_name
> * test_no_comment
> * test_past_winner
> * test_rule_2a_size
> * test_rule_2b_size
> * test_tarball
> 

Thank you for the list!

> In addition to whatever test functions need to be applied to `mkiocccentry.c` code.
> 

Of course. 

> ### UPDATE 0
> We will work on writing `test_tarball()` as this is a bit more complex.
> 

Yes. You might (or might not) find the old code of mine of use for this one. But then a lot has changed so maybe it’s not as useful as it might otherwise be. 

> All of the other test functions above should be fairly straightforward value testing.

Yes.

xexyl:
As for the test size functions. Some code from mkiocccentry might be useful. 

But also I thought I had added these functions. If I did however - ah it has to be the override bools. I just remembered that and that might be why I haven’t done those yet. I doubt I will do any more test functions today though but hopefully tomorrow I can do more. We shall see.

xexyl:
I'm currently finishing these off except the one you said you're working on. Just a FYI.

PR #321, opened 2022-08-22
More test_ functions added + typo fixes

xexyl:

    extern bool test_name(char *str);
    extern bool test_no_comment(char *str);
    extern bool test_past_winner(bool boolean);

I was going to add the two rule size test functions but I'm unclear
about something so I'll be posting a comment instead to discuss it once
a pull request has been formed.

Typo fixes in entry_util.c and mkiocccentry.c. This includes changing
pseudo-name to pseudonym - this fix could possibly cause a logomachy but
I deem that risk worth it as it should be correct! :-)


xexyl:
I was going to do the test size functions but something occurred to me: how to decide what is valid? I mean if the function returns false it'll indicate the entry is invalid but under what grounds? After all there's the option to override the warning so how is this necessarily invalid?

Please advise and I'll be happy to do these two functions. Oh and should they take a bool of the override? I guess not unless these are done in the correct order: but we have to test that the booleans do not mismatch with the sizes.

lcn2:
> I was going to do the test size functions but something occurred to me: how to decide what is valid? I mean if the function returns false it'll indicate the entry is invalid but under what grounds? After all there's the option to override the warning so how is this necessarily invalid?
> 
> Please advise and I'll be happy to do these two functions. Oh and should they take a bool of the override? I guess not unless these are done in the correct order: but we have to test that the booleans do not mismatch with the sizes.

We will have sleep on this .. perhaps the test function needs to take 2 args, one is the size and the 2nd arg is a override boolean .. not sure.  Going to sleep on it.

xexyl:
> > I was going to do the test size functions but something occurred to me: how to decide what is valid? I mean if the function returns false it'll indicate the entry is invalid but under what grounds? After all there's the option to override the warning so how is this necessarily invalid?
> > Please advise and I'll be happy to do these two functions. Oh and should they take a bool of the override? I guess not unless these are done in the correct order: but we have to test that the booleans do not mismatch with the sizes.
> 
> We will have sleep on this .. perhaps the test function needs to take 2 args, one is the size and the 2nd arg is a override boolean .. not sure. Going to sleep on it.

Sleep well! About to get some sleep too.

lcn2:
Actually it was easy .. short nap .. :) ...

The `test_rule_2a_size()` function should check for >= 0 and <= `RULE_2A_SIZE`.
The `test_rule_2b_size()` function should check for >= 0 and <= `RULE_2B_SIZE`.

The issue of `rule_2a_override` and `rule_2b_override` will be handled by the check functions that we will write.  The `mkiocccentry` code does the asking of the over if they want to try and override the Rule 2 size limits already.

xexyl:
> Actually it was easy .. short nap .. :) ...
> 

That was no nap! 😀

> The `test_rule_2a_size()` function should check for >= 0 and <= `RULE_2A_SIZE`. The `test_rule_2b_size()` function should check for >= 0 and <= `RULE_2B_SIZE`.
> 

Okay but won’t this prevent some clever abuse of rules with size? I mean okay most likely it would be disqualified but what if?

> The issue of `rule_2a_override` and `rule_2b_override` will be handled by the check functions that we will write. The `mkiocccentry` code already does the asking of the user (if they want to try and override the Rule 2 size limits) already.

I already wrote some code handling the inconsistencies in the old code files. Not sure if any of it will be useful though. 

As for the mkiocccentry: one of the reasons I don’t necessarily update it to use the new functions is because it doesn’t easily allow for individual tests. It might work in some cases but not all. 

If this doesn’t make sense I will try clarifying at the laptop tomorrow. Please advise on the other question with sizes being too big. 

Going to sleep now. Cheers.

xexyl:
Also if it checks for >= 0 it means that it might not trigger one of the override tests?

Anyway really going to sleep now.

lcn2:
> Also if it checks for >= 0 it means that it might not trigger one of the override tests?
> 
> 
> 
> Anyway really going to sleep now. 

You are correct, the lower bound test should be > 0 (not >= 0) so that a zero size (as well as over the rule limit size) will fail the test.

Failing the `test_rule_2a_size()` test does NOT disqualify the entry.  It just triggers, in the case of `mkiocccentry` to ask the user if they really want to try an break the rule.  In the case of the chk function we will write, the presence of a `true` value in the appropriate boolean in the parse tree will allow a 0 size or an over the limit size to still return true.

xexyl:
> > Also if it checks for >= 0 it means that it might not trigger one of the override tests?
> > Anyway really going to sleep now.
> 
> You are correct, the lower bound test should be > 0 (not >= 0) so that a zero size (as well as over the rule limit size) will fail the test.
> 

Thank you.

> Failing the `test_rule_2a_size()` test does NOT disqualify the entry. It just triggers, in the case of `mkiocccentry` to ask the user if they really want to try an break the rule. In the case of the chk function we will write, the presence of a `true` value in the appropriate boolean in the parse tree will allow a 0 size or an over the limit size to still return true.

Ah. I see. Well that helps. I was going to add the boolean anyway like you suggested but the mail preview (from iOS Mail preview in Notification Centre) didn't show this much of the mail. In this case I'll add the functions without the boolean. Once I've done this I'll do a pull request though it might take longer than it should as I'm both awake and unawake: I want to be asleep but cannot sleep for an hour now so I'm sitting up at the laptop. I will try going back to sleep in bit.

I hope later on I can tackle the txzchk stuff but that might already mostly be done anyway other than printing the results depending on debug level.

I'll say here instead of elsewhere as it's simpler: for commit edb3bd70bb5715caf5e392339a4b0b885eecec89 thank you for this part in particular:

    Changed a number of test_foo() debug messages to be consistent and
    to not leak potential entry info at JSON_DBG_MED level.
    
That was indeed my doing. I try including more information in log messages but I did not think of this here. Or .. at least i'm pretty sure that was me!

Hope you had a good sleep my friend and I hope you're having a good day!

xexyl:
Finished those functions; see https://github.com/ioccc-src/mkiocccentry/pull/322/commits/f95e092de85146d6c095cd67f86c96cd63fc728f.

I believe this completes the test_ functions.

PR #324, opened 2022-08-30
Verify .info.json, .author.json and fnamchk exist

xexyl:
There were some other minor fixes and changes too of course. This is all I'll be doing for the day. Off to do other things.

Note that I did not remove that duplicate define as I don't know if `chkentry.h` should include the header file or not.

Good day! I hope this helps. I'll quote the XXX marks in the other thread and then will be doing something else.


xexyl:
Please see https://github.com/ioccc-src/mkiocccentry/issues/259#issuecomment-1231960822 as well.

lcn2:
> Please see [#259 (comment)](https://github.com/ioccc-src/mkiocccentry/issues/259#issuecomment-1231960822) as well.

We have to look at this pull request is more detail as well as the above mentioned comment before we merge this pull request.

xexyl:
> > Please see [#259 (comment)](https://github.com/ioccc-src/mkiocccentry/issues/259#issuecomment-1231960822) as well.
> 
> We have to look at this pull request is more detail as well as the above mentioned comment before we merge this pull request.

Sure. I can rename the function if that helps. But I think it's fine for now. Some of it would have to change later on maybe but the checks could simply be yank pasted. The code is there and that's good it's just a matter of maybe needing to move it about.

xexyl:
Another thought of taking care of the problem - though it seems a bit much - is having a callback that will process the files or check that the files exist. That way the same function could be used.

But if you want me to make some changes I will be happy to do that but it'll have to be tomorrow as I'm heading off now. Anyway it shouldn't be a problem to move the code about. I just wanted the actual checks in but that's why I noted that it might need to be moved here and there.

xexyl:
I think I have a solution to the problem. See https://github.com/ioccc-src/mkiocccentry/issues/259#issuecomment-1232084167. It would be a natural progression of this commit and I can start it tomorrow.

lcn2:
We need to think about this pull request a bit more.

xexyl:
> We need to think about this pull request a bit more.

What about it? If you're quick I can maybe fix it.

xexyl:
Making some quick fixes ... stay tuned.

xexyl:
I hope that helps. Stub functions added which will parse the json. If not invalid json it'll get a `struct json *` which can be used to validate the file's contents. This should remove any concern of code duplication.

Of course there might be that some other things have to be fixed too but I'll have to do that tomorrow as I need to go now. Will make a comment in the other thread and then shutting down.

xexyl:
> I hope that helps. Stub functions added which will parse the json. If not invalid json it'll get a `struct json *` which can be used to validate the file's contents. This should remove any concern of code duplication.
> 
> Of course there might be that some other things have to be fixed too but I'll have to do that tomorrow as I need to go now. Will make a comment in the other thread and then shutting down.

See also https://github.com/ioccc-src/mkiocccentry/issues/259#issuecomment-1232109271 for some example output.

Will fix anything needed tomorrow if anything needs to be done. Must go now.

xexyl:
...okay leaving now for real. Will reply tomorrow to anything. 

Cheers.

lcn2:
Thanks @xexyl we figured it was better to accept the pull and look at the result and if needed modify.

The direction of `chkentry` is not fully laid out .. so things might change.

lcn2:
We will need to make some major surgery to this pull request in terms of JSON semantics.

You might wish to focus on other areas of code ...

xexyl:
> Thanks @xexyl we figured it was better to accept the pull and look at the result and if needed modify.
> 
> The direction of `chkentry` is not fully laid out .. so things might change.

That makes sense then.

> We will need to make some major surgery to this pull request in terms of JSON semantics.
> 

Well first I had major surgery and now chkentry will have some major surgery so I feel less alone! :-)

> You might wish to focus on other areas of code ...

Until you have figured out chkentry more, right? What about consistencies and other things like you suggested the other day? Should I think on those still?

xexyl:
I just noticed a problem with this commit ... will fix when I am back from a lie down.

PR #325, opened 2022-08-31
Minor bug fix in chkentry.c

xexyl:
Don't assume failure to validate files because when the file is ignored
it would make the test fail when it shouldn't. This fixes make test.


xexyl:
https://github.com/ioccc-src/mkiocccentry/pull/325/commits/e4c3d2b7d4f7d8c160f499395ce831da7e1354b9 is because something could have been a bit too revealing .. leaving it at that as I don't want to reveal any more! :-)

Going to rest now ... might do more today but not sure. Cheers.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Sure .. and thanks to you too!

PR #328, opened 2022-09-01
Typo and comment consistency fixes

xexyl:


xexyl:
> Thanks @xexyl
> 
> We will look at the other questions in other comments when we are back at a laptop.

Likewise I will read your replies when at the laptop which won’t be until tomorrow.

PR #333, opened 2022-09-04
Update man pages + add script to HELP find inconsistencies in SYNOPSIS and usage

xexyl:
Removed AUTHORS and added NOTES. In the case of jparse.1 I also removed
the FILES section. There's some humour here and there as well.

There was a typo fix in foo.c that I noticed as well which I had to fix.


xexyl:
Enjoy the NOTES in the txzchk man page in particular! I'm going to do something else now. Good day!

xexyl:
As for https://github.com/ioccc-src/mkiocccentry/pull/333/commits/92c51a273fdc09b944bead46937741aea99585e4: I forgot to say in the log that it also checks for man pages and you can of course specify the man section as well.

There are some limitations but it's helpful as a guide. Obviously it does not try editing the man page at all. It even warns that one has to know what they're doing and to not do it if there are any doubts. I think however it'll be useful to have in the repo. Right now it only works on one tool at a time; I did that to simplify it.

Some things need to be fixed and that includes adding it so that it can check its own man page (which doesn't exist yet for that reason). Have to go very soon but this will be of help as the tools are updated.

Hope you're well and in case we do not interact before your trip home I hope it goes well and safe! I also hope you're well in general. Until next time.

lcn2:
Thanks @xexyl.. good progress on documentation 👍

xexyl:
> Thanks @xexyl.. good progress on documentation 👍

Thank you for that! I have some improvements to the script that will be committed soonish. It still won't detect some problems but then it's not meant to replace editing the man page anyway.

I hope you're doing well!

I might have to take a break soon but I should have the changes in soon. And yes there will be a number of quotes added to that special array. I added them when I wasn't able to work on anything else.

PR #335, opened 2022-09-06
Bug fixes and minor improvements to run_usage.sh + typo fixes

xexyl:
Better handling of missing usage in man page. It's still not perfect but
it at least now checks the number of matches in the man page. This can
still not work properly but fixing it would not be worth it when one has
to still edit the man page and verify it renders okay. This change is
worth it though.

If there's no usage string (either because the tool does not have a -h
option or the -h option does not have any line with 'usage:') the script
no longer shows empty (and thus invalid) possible SYNOPSIS to add.

Moved the limitations comments to the man page (some of which were
already there and some which were not).

Script now says it's possible that there's a missing or inconsistent
SYNOPSIS.

Updated the man page.


xexyl:
.. I found a mistake in the jparse man page. Will fix that another time. Or I should say I'll commit it another time. It happened because the meaning of an option changed slightly.

xexyl:
Question for you Landon:

Should the script attempt different man page sections? For example if it fails to find it in second 1 it could check additional sections? One could still specify the man section or man page directly but it would make it easier to add a Makefile rule that runs the script on all the targets to see what's missing / what's inconsistent / etc.

What do you think?

xexyl:
With commit https://github.com/ioccc-src/mkiocccentry/pull/335/commits/8614df9294dcb689ede30e9be812ded76a8ae22e the holiday thing is no longer true .. but I put a joke in there too that you'll appreciate (more once you get back to where you can see email).

I'm hoping to add a man page for the timegm script later on today but right now I need to rest. I'm thinking for now (this could change) I'm going to try focusing on documentation.

lcn2:
That is a good chunk of work, @xexyl !

xexyl:
> That is a good chunk of work, @xexyl !

Thank you! I wish I had realised I did not update the title to say man pages too. Oh well. Anyway I'm happy I got so much done.

BTW you should try activating foo right now. I did improve the current quote slightly. I can tell you how since it won't spoil anything: I changed the '5' to '0.5'.

I also am eager to share the tool with you as I think you'll really appreciate the joke I put in the header file. It's good enough with the comment at the top but it's even better. Well when you're home I can do that. I hope you have a good and safe trip home - if you're not home yet.

lcn2:
On our back home ..

xexyl:
> On our back home ..

Safe travels! Please let me know when you're able to receive email again though I probably won't be able to get it to you till tomorrow.

PR #337, opened 2022-09-10
Silence warning in foo

xexyl:
Supposedly it was a buffer overflow but this was triggered by -Weverything and previous tests on my part showed it's fine. It's using an obscure feature (as foo does) so it probably just confused the compiler or it's being extra cautious or something like that.


xexyl:
I had a really rough night so I don't think I'll manage to get anything done today but we'll see.

Good day! Hope you're sleeping well. I'm going to try going back to sleep myself.

xexyl:
BTW: can you receive emails now ?

lcn2:
Good foo @xexyl 😀

xexyl:
> Good foo @xexyl 😀

Thank you. Do you refer to the row of crosses and how it probably crossed you (the reader of the code) or something else? I enjoyed putting that in. 

I also enjoy (and it’s there though I think only two so not actually there in this way) doing `- - - - -` … 

When I was rewriting the script system in the rewrite of the MUD I told you about when I discovered that that was legal it was a bloody nuisance I can tell you!

But I was also trying to learn the ghastly C++. I tried really hard to like it but I found it abhorrent and ended up doing what I call C-ifying. Make it more and more like C - you know things like ignoring the stl library and using direct pointers and so on. But other things took over and I stopped working on it. 

I mean it was more or less done but they are also kind of history though not totally. Personally I feel like Java is more like C than C++ and at least that has a pretty impressive graphics API. 

I still don’t know why clang triggered that warning. It seems useless to me. The calculation doesn’t go beyond the array size so I am not sure why it thinks it did. Oh well. It is resolved now for whatever it is or not worth!

Tomorrow I will send you the new version of the tool so that you can have some fun! Unfortunately I haven’t made it possible for extracting the plaintext strings so that we can each individually add strings but we can discuss that over email later on. Anyway it’s lower on the priority queue. 

I hope I replied to everything. I see you just made another comment about a commit but I will have to check it tomorrow so I might as well reply at that point too after I have seen it!

Good night my friend! Sleep well when you do. I am sorry I have to go to sleep now. It’s a real nuisance but anyway hopefully tomorrow we can discuss more and I can get some things in. 

Until next time! BCNU!

PR #340, opened 2022-09-12
Typo fixes + man page additions and fixes

xexyl:


xexyl:
Doing something else now .. probably won't do anything else here today as I'm too tired and am forgetting little details here and there.

I can probably do more discussion depending on when you reply but it'll probably be an early night today.

xexyl:
> Thanks @xexyl

Happy to help! 

I will check the commit on the private tool tomorrow morning hopefully. There’s a possibility that I can do something here but it would have to be me being awake enough earlier on. 

Otherwise it won’t be until Wednesday. 

Going to sleep now. 

One thing: should I focus on the pending issues with txzchk or should I focus more on the man pages? (Assuming of course that I am alert enough to do either!)

Good night!

PR #342, opened 2022-09-15
Fix commit d9018f5505c2eb0b1fa77baa1ee9f73af70b3b7a

xexyl:
When I changed jparse_test.sh to use -j instead of -p I did update the case statement but I forgot to update the getopts which would mean that the -j option would never have been seen. Thank you Landon for pointing this out to me!

### DIFF
--- a/jparse_test.sh
+++ b/jparse_test.sh
@@ -33,7 +33,7 @@ export EXIT_CODE=0
 
 # parse args
 #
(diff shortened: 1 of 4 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help my friend!

Hope you're having a nice sleep too.

PR #344, opened 2022-09-20
Fix potential invocation errors in Makefile

xexyl:
Just like with commit 3d1c7f12be1853d01e182ff97ccdad985b274616 I have changed the command -v to type -P due to the possible problem I described in that commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -828,7 +828,7 @@ chk_sem_info.h: jsemtblgen jsemcgen.sh test_JSON/info.json/good/info.reference.j
 # sequence exit codes
 #
 seqcexit: Makefile
-	@HAVE_SEQCEXIT="`command -v ${SEQCEXIT}`"; if [[ -z "$$HAVE_SEQCEXIT" ]]; then \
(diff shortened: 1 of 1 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Sure. 
> 
> We confirmed that this change works well on several platforms.

That is good! Makes sense that it would but I am glad you tested it because I was concerned that I could have made a typo (even with the simple regular expression used … not even really a regexp but just a search and replace). 

Like I said earlier I have no idea why I even thought to test the problem. I just did. Maybe I suspected something might not work right. I have been known to do that many times before. I kind of have a knack for finding problems. 

Which is a funny thing I noted years ago: if you’re really good at debugging or troubleshooting or solving problems more generally - if you’re good enough you actually understand how problems work so you can more easily cause them. Which I guess one might expect from an IOCCC winner. 

Well anyway glad it works well. Bedtime for me. Good night!

xexyl:
Actually there’s a funny childhood story about solving problems. I will tell you in the other thread. I will have a read of your reply to my other question tomorrow (I hope!).

PR #347, opened 2022-09-23
Man page updates + addition of mkiocccentry_test.8

xexyl:
As for the files not named I simply fixed the name so that they all have .sh in the name as they're all shell scripts.

The usage of make install will now install mkiocccentry_test.8 as well.


xexyl:
.. not sure if I'll do anything else today. If I don't do anything soon I probably won't at all and I don't feel up to doing anything right now so it might be that this is all I end up doing. Will be shutting the laptop down soon again. I will have it up at least one more time but it's very possible I won't be doing anything here at that point because as I mentioned elsewhere today is rather shot.

If I do anything it'll possibly be more man pages but we'll see. Hope you're having a nice sleep.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help. I'm afraid I won't be doing anything else today here but maybe tomorrow I can. Have a great rest of your day!

PR #348, opened 2022-09-24
Header comment updates in txzchk*.{c,h,sh} + fix make picky + bug fix in 4d0b1b

xexyl:


lcn2:
Thanks @xexyl 

Power is being restored after hurricane in Canada.

xexyl:
> Thanks @xexyl
> 

Welcome. I presume that the way I handled `mkiocccentry` with the `MAX_DIR_KSIZE` macro is fine ?

> Power is being restored after hurricane in Canada.

Crikey! You went from the US into a hurricane! I didn't know you were supposed to be there yet though; I thought it was the other things you have going on. Glad you're okay! I almost asked but been kind of hectic here too though clearly not as hectic as it has been for you!

Stay safe!

lcn2:
Limited power restored

lcn2:
> Welcome. I presume that the way I handled mkiocccentry with the MAX_DIR_KSIZE macro is fine ?

Due to limited access, we only were able to browse via GitHub (no code search function in GitHub browser).  Did not see `MAX_DIR_SIZE` being used, so we cannot tell.  Sorry (tm Canada 🇨🇦) 🥸

xexyl:
> > Welcome. I presume that the way I handled mkiocccentry with the MAX_DIR_KSIZE macro is fine ?
> 
> Due to limited access, we only were able to browse via GitHub (no code search function in GitHub browser). Did not see `MAX_DIR_SIZE` being used, so we cannot tell. Sorry (tm Canada 🇨🇦) 🥸

Let me try rewording it better. The diff looks like:

```diff
diff --git a/mkiocccentry.c b/mkiocccentry.c
index 29a58e0..e2ee0e9 100644
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -4403,26 +4403,10 @@ verify_entry_dir(char const *entry_dir, char const *ls)
     }
     if (kdirsize <= 0) {
        err(132, __func__, "ls k block value: %d <= 0", kdirsize);
        not_reached();
     }
-    if (kdirsize > MAX_DIR_KSIZE) {
-       fpara(stderr,
-             "",
-             "The entry is too large.",
-             "",
-             NULL);
-       errno = 0;                      /* pre-clear errno for errp() */
-       ret = fprintf(stderr, "The entry directory %s is %d kibibyte (1024 byte blocks) in length.\n"
-                             "It must be <= %d kibibyte (1024 byte blocks).\n\n", entry_dir, kdirsize, MAX_DIR_KSIZE);
-       if (ret <= 0) {
-           errp(133, __func__, "fprintf error while printing entry directory kibibyte sizes");
-           not_reached();
-       }
-       err(134, __func__, "The entry directory is too large: %s", entry_dir);
-       not_reached();
-    }
```

Thus the macro is removed and we let `txzchk` handle the actual summing and counting. This is because to make it so mkiocccentry do it I would have to take the code I wrote for txzchk but make a number of modifications - with certain problems to be addressed (some of which would be complicated with some decisions to be made being unclear). Since txzchk has to be used for every entry anyway I don't see the harm in this but I do see gains in it. If however this is not true and the function that runs `ls` on the entry directory in `mkiocccentry` then it will add some complications to `mkiocccentry` including having to do similar parsing of the lines like `txzchk` does.

Does that help clarify? If not feel free to just delay this until another time. That's okay.

PR #350, opened 2022-09-30
Typo fix in jsemcgen.sh

xexyl:


xexyl:
Probably all I'll do today here. It's a particularly frustrating day - unrelated of course - on top of being tired (as I always am of course).

Maybe I'll be up to doing more later on but for now I'll be afk. Stay safe! I mean including the disaster of the hurricane.

lcn2:
Good catch @xexyl

xexyl:
> Good catch @xexyl

Thank you and glad to help!

I don't think I'll get anything done today after all. I can't rest but I'm too tired to focus so I'm going to just do something that doesn't take any thinking.

xexyl:
vim tip: to put the name of the file into the vim buffer try (in insert mode):

```vim
^r%
```

where of course ^r means ctrl + r. It always has irritated me that vim has that kind of usage but it is what it is (that kind meaning ctrl/etc.).

See also:

https://vim.fandom.com/wiki/Get_the_name_of_the_current_file

and

https://vim.fandom.com/wiki/Insert_current_filename (the comments under this one is quite interesting and useful - not things I would have thought of probably though who knows)

BTW: I use one of the things mentioned in the first article for setting the title of the shell (tab) when editing a file in vim like this in `.vimrc`:

```
    autocmd! BufEnter * let &titlestring = "vim " . strpart(hostname(), 0, stridx(hostname(), ".")) . ":" . expand("%:p")

```

You can look at how else to change it by `:help titlestring` and `:help statusline`.

One more tip: if you have in `.vimrc`:

```
"always show file name etc. at bottom of screen (next two lines):
set ruler
set laststatus=2
```

it can help you see what file you're editing. It will update if you have more than one window open (see `:help new` ... you can do this at the command line with options `-o` and `-O` depending on which orientation you want).

Going to do something else now ... have a great day!

BTW: how are things after the hurricane that you ended up heading right into ? Stay safe either way!

PR #351, opened 2022-10-03
Initial (incomplete) locations and pure parser + make make picky run picky on shell scripts

xexyl:
The first step is the removal of the prefix sorry_. This started out as a joke between Landon and me, first as ugly_ but changed to sorry_. As Landon explained:

    Avoided the appearance of attacking any particular individual.  It
    was not our intention to disrespect anyone, even though we disagree
    with some of the technical decisions.  Where we have fundamental
    technical disagreements, we attempted to express those technical
    disagreements with humor in hopefully a more fun way.  As also now
    apologize for how `bison` and `flex` generated code may look, instead
    of simply calling it ugly.  As such, we hopefully improved some of
    the humor in our code while trying to be nice and friendly to others.

He elaborated that:

    FYI: One reason to avoid the appearance of an attack is out of a
    courtesy to Steven Johnson, the original architect of yacc. Brian
    Kernighan credits Steven Johnson with use of his original yacc
    allowing him to quickly produce the awk interpreter. While neither
    is what you would call "humor impaired", out of respect for their
    work we decided to "tone down" language and make it appear to be
    more "fun" and less "caustic".

The reason to remove the prefix is because it causes confusion between documentation and what is generated i.e. between what is usually expected and what we actually have. This is important because there are a number of complications including that the interaction of bison and flex when it comes to re-entrancy is a proper mess so we need to have less confusion. Another fun prefix can be added another time once everything is complete (finishing locations and making it re-entrant which was also kind of started) should it be desired.

Now as for the changes:

Make bison parser pure. This does not mean it's yet re-entrant but it's the first step to do so. This was done by adding:

    %define api.pure full

to jparse.y.

Add %locations to jparse.y.

Removed all references of sorry_ and SORRY_ as well as all the commentary on the prefix.

Changed sorry.tm.ca.h to work with the removal of the prefix but still be sorry (see the diff if you need to know as this log is already too long).

Several functions have had a change in prototype. These include yyerror() which now takes a YYLTYPE *yyltype (currently unused but provides location information - or will) and yylex() which is now:

    extern int yylex(YYSTYPE *yylval_param, YYLTYPE *yylloc_param);

the YYSTYPE * is our struct json.

These changes do not seem to break any parsing as the jparse_test.sh script reports no problems (nor does make test or make prep) and running the parser on a file manually returns the expected results. Even so this is not complete. I want it in so that Landon can also work on it.

The Makefile did have to change slightly as (because these changes are not completed) one parameter of a function is not yet used (i.e. a warning was disabled).

We can say with this commit that the JSON parser is still code complete but the locations and the re-entrancy is not.

There was a typo fix in jsemtblgen.c as well.


xexyl:
... doing something else now and I don't expect to return to the repo today...

If you'd rather first discuss the experimenting with the location and pure parser I'm quite okay with that. You can just merge the other commit (or I can recommit it in another pull request). But this way you can also see the changes live without having to extract the file I uploaded yesterday (which is slightly different from what the commit above is as I improved the sorry file).

lcn2:
Thanks you @xexyl

xexyl:
> Thanks you @xexyl

Welcome. I was going to say you’re up early but maybe not since you might currently be in a different time zone. 

Hope you have a great day!

We can talk about these changes in the other thread I guess?

xexyl:
> Thanks you @xexyl

Happy to help! We can discuss it in the other thread.

PR #352, opened 2022-10-03
`make parser' now compiles chkentry + update readme for txzchk_test.sh

xexyl:
The question of whether or not it should just do `make all' is certainly something to consider but for now this covers everything that will need to be recompiled after the parser is regenerated.

There was also a typo fix in jparse.l - kind of. I had removed 'demo' from 'demo tool' but I meant to add it back as maybe jparse itself is kind of a demo as it doesn't actually do anything with it - it just parses it but does nothing with it other than report if it's valid or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -700,6 +700,7 @@ parser: jparse.y jparse.l Makefile
 	${CP} -f -v jparse.c jparse.ref.c
 	${MAKE} jparse
 	${MAKE} jsemtblgen
+	${MAKE} chkentry
 
 #
 # make parser-o: Force the rebuild of the JSON parser.
--- a/jparse.l
+++ b/jparse.l
@@ -37,7 +37,7 @@
 
 
 /*
- * jparse - JSON parser tool
+ * jparse - JSON parser demo tool
  */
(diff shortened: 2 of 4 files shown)


xexyl:
... off to do other things for the rest of the day now.

xexyl:
Shutting down for the day. Maybe I can do more tomorrow.

Good day!

lcn2:
Thanks @xexyl 

... still in Nova Scotia

xexyl:
> Thanks @xexyl
> 

Welcome. 
> ... still in Nova Scotia

Stay safe!

I haven’t felt up to much lately but I hope that in some days I will be up to doing more. I will be going to sleep soon. 

Got another booster dose today. Flu vaccine in a couple weeks. Very tired from Saturday. But still glad I got my next dose. It will never end sadly. There was a chance at the very beginning that it wouldn’t become seasonal but I don’t think there’s a chance of it not being seasonal now. But right now it hasn’t even got to that point yet. 

Who knows when it will. Seems it won’t be in the near future anyway sadly. 

Enjoy the rest of your day my friend!

PR #353, opened 2022-10-05
Improve make depend + add some (bad) test files (txzchk and `.info.json` and `.author.json`) and minor rewording in txzchk

xexyl:


xexyl:
Doing something else now and I don't expect to resume here until another day.

lcn2:
Still in the far corner of Canada.  Interesting take on the depend rule.  We are not sure that one needs to be that careful for systems without `grep(1)` and `cut(1)`, although it doesn't hurt.

Given the complexity of the depend rule, the code is better moved into a `depend.sh` shell script.

There is an issue with:

```make
     ${CC} ${CFLAGS} -MM ${ALL_CSRC} >> Makefile; \
```

command in that different C compilers will output the dependency rules in slightly different ways.  For example, some will output:

```make
mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
  location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h \
  iocccsize.h json_util.h json_parse.h entry_util.h json_sem.h
```

while another will output:

```make
mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
 location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h iocccsize.h \
 json_util.h json_parse.h entry_util.h json_sem.h
```

Notice that because the one compiler will output two spaces for the intended line
instead of a single space intent, the wrap of a long line can current at a place.
In the above example, `iocccsize.h` is at the end of a line in one case whereas
the other cause it was sifted to the start of the next line (because of the 2 space intent).

The result of this problem is that the `make depend` rule all will result in unnecessary
changes to `Makefile` when run on different compiler platforms.

What needs to happen in the depend.sh` shell script is to pipe the output
of the `${CC} ${CFLAGS} -MM ${ALL_CSRC}` through an `awk(1)` command
that will canonicalize the output in a C compiler independent way.

lcn2:
Thanks @xexyl 

Consider [comment 1269076754](https://github.com/ioccc-src/mkiocccentry/pull/353#issuecomment-1269076754) is you will.

xexyl:
> Still in the far corner of Canada. Interesting take on the depend rule. We are not sure that one needs to be that careful for systems without `grep(1)` and `cut(1)`, although it doesn't hurt.
> 
> Given the complexity of the depend rule, the code is better moved into a `depend.sh` shell script.
> 
> There is an issue with:
> 
> ```makefile
>      ${CC} ${CFLAGS} -MM ${ALL_CSRC} >> Makefile; \
> ```
> 
> command in that different C compilers will output the dependency rules in slightly different ways. For example, some will output:
> 
> ```makefile
> mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
>   location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h \
>   iocccsize.h json_util.h json_parse.h entry_util.h json_sem.h
> ```
> 
> while another will output:
> 
> ```makefile
> mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
>  location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h iocccsize.h \
>  json_util.h json_parse.h entry_util.h json_sem.h
> ```
> 
> Notice that because the one compiler will output two spaces for the intended line instead of a single space intent, the wrap of a long line can current at a place. In the above example, `iocccsize.h` is at the end of a line in one case whereas the other cause it was sifted to the start of the next line (because of the 2 space intent).
> 
> The result of this problem is that the `make depend` rule all will result in unnecessary changes to `Makefile` when run on different compiler platforms.
> 
> What needs to happen in the depend.sh`shell script is to pipe the output of the`CC{CFLAGS} -MM ${ALL_CSRC}`through an`awk(1)` command that will canonicalize the output in a C compiler independent way.

Well that issue has already been discussed. We (iirc) decided it’s not really an issue or at least it won’t be later on since once everything is finished we won’t really need to do make depend so the changes in indentation won’t matter. 

It might not be necessary to have checks on the cut and grep tools. Not sure. Originally I did it because awk was used and that’s not on every system. But then I improved it to use cut. 

So I propose that I could get rid of the check for those tools and be done with it. Or we can keep them there and be done with it. 

But maybe you have some ideas in the above that will be useful and easier to read and consider at the laptop. I am typing this on my phone and am about to go get ready to sleep. 

Let me know what you think of this and whatever the case I will look at the comment in more detail tomorrow or the next day - probably tomorrow unless I have a bad night. 

Enjoy the rest of your day and stay safe!

xexyl:
> Thanks @xexyl
> 
> Please consider [comment 1269076754](https://github.com/ioccc-src/mkiocccentry/pull/353#issuecomment-1269076754) is you will.

Just replied but perhaps incompletely. Will look at it more at the laptop. I can say that the indentation issue we already discussed and came to a consensus. But I will look at it more tomorrow or the next day. 

Good night!

xexyl:
> > Thanks @xexyl
> > Please consider [comment 1269076754](https://github.com/ioccc-src/mkiocccentry/pull/353#issuecomment-1269076754) is you will.
> 
> Just replied but perhaps incompletely. Will look at it more at the laptop. I can say that the indentation issue we already discussed and came to a consensus. But I will look at it more tomorrow or the next day.
> 
> Good night!

Btw: there’s another reason to not worry about the indentation. Namely that adding another script - another file - is probably not necessary for a minor detail especially since we both primarily develop under macOS so the same compiler. Add that once the repo is ready there won’t be a need for make depend that adding another file (which we’re going to try and reduce or at least reorganise the directory) seems like unnecessary clutter. 

What do you think? Well anyway I am going now but I will leave this page open on my phone so hopefully I will look at it the next time I have the laptop up - since I will definitely be looking at the phone too. 

Good night!

lcn2:
> > Still in the far corner of Canada. Interesting take on the depend rule. We are not sure that one needs to be that careful for systems without `grep(1)` and `cut(1)`, although it doesn't hurt.
> > Given the complexity of the depend rule, the code is better moved into a `depend.sh` shell script.
> 
> For the reasons I cited above I still don't think this is actually true. It's not that complex, really, and I works perfectly in the Makefile as well. It also prevents overwriting the Makefile with nonsense which I feel a shell script could more easily do.
> 
> > There is an issue with:
> > ```makefile
> >      ${CC} ${CFLAGS} -MM ${ALL_CSRC} >> Makefile; \
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > command in that different C compilers will output the dependency rules in slightly different ways. For example, some will output:
> > ```makefile
> > mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
> >   location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h \
> >   iocccsize.h json_util.h json_parse.h entry_util.h json_sem.h
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > while another will output:
> > ```makefile
> > mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
> >  location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h iocccsize.h \
> >  json_util.h json_parse.h entry_util.h json_sem.h
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Notice that because the one compiler will output two spaces for the intended line instead of a single space intent, the wrap of a long line can current at a place. In the above example, `iocccsize.h` is at the end of a line in one case whereas the other cause it was sifted to the start of the next line (because of the 2 space intent).
> 
> Which as I said earlier (and which we discussed a long while back) this should not be a problem in the end as there won't be any need for `make depend`. However I did not say the below:
> 
> > The result of this problem is that the `make depend` rule all will result in unnecessary changes to `Makefile` when run on different compiler platforms.
> 
> This will only happen if someone runs `make depend` though and on a different machine. And the odds are if they do run it they run it for a good reason - a dependency change. Right?
> 
> > What needs to happen in the depend.sh`shell script is to pipe the output of the`CC{CFLAGS} -MM ${ALL_CSRC}`through an`awk(1)` command that will canonicalize the output in a C compiler independent way.
> 
> I'm not even sure how to go about this with the above in mind as well as the fact that it's only on some lines (for targets that have more than one line worth of rules). Given that this would only clutter the directory even more as well (which I mentioned yesterday) I'm also not sure it's needed?

Well when you start getting a _pile of backslashed lines_, it might be time to put them into a shell script.  The question is if

- make limit_ioccc.sh
- make all_ref
- make seqcexit
- make picky
- make checknr
- make man2html
- make depend

are at the _pile of backslashed lines_ level?   Some of them are not _pile of backslashed lines_, others might be, still others seem to be a _pile of backslashed lines_.

Remember that with issue #338 such shell scripts will be moved down into a sub-directory, hiding the clutter.  So perhaps converting the rules that have become _pile of backslashed lines_ into invocations of a shell script is a reasonable option and keeps the Makefile for manageable?  Maybe?

xexyl:
> > > Still in the far corner of Canada. Interesting take on the depend rule. We are not sure that one needs to be that careful for systems without `grep(1)` and `cut(1)`, although it doesn't hurt.
> > > Given the complexity of the depend rule, the code is better moved into a `depend.sh` shell script.
> > 
> > 
> > For the reasons I cited above I still don't think this is actually true. It's not that complex, really, and I works perfectly in the Makefile as well. It also prevents overwriting the Makefile with nonsense which I feel a shell script could more easily do.
> > > There is an issue with:
> > > ```makefile
> > >      ${CC} ${CFLAGS} -MM ${ALL_CSRC} >> Makefile; \
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > command in that different C compilers will output the dependency rules in slightly different ways. For example, some will output:
> > > ```makefile
> > > mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
> > >   location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h \
> > >   iocccsize.h json_util.h json_parse.h entry_util.h json_sem.h
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > while another will output:
> > > ```makefile
> > > mkiocccentry.o: mkiocccentry.c mkiocccentry.h util.h dyn_array.h dbg.h \
> > >  location.h utf8_posix_map.h limit_ioccc.h version.h sanity.h iocccsize.h \
> > >  json_util.h json_parse.h entry_util.h json_sem.h
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Notice that because the one compiler will output two spaces for the intended line instead of a single space intent, the wrap of a long line can current at a place. In the above example, `iocccsize.h` is at the end of a line in one case whereas the other cause it was sifted to the start of the next line (because of the 2 space intent).
> > 
> > 
> > Which as I said earlier (and which we discussed a long while back) this should not be a problem in the end as there won't be any need for `make depend`. However I did not say the below:
> > > The result of this problem is that the `make depend` rule all will result in unnecessary changes to `Makefile` when run on different compiler platforms.
> > 
> > 
> > This will only happen if someone runs `make depend` though and on a different machine. And the odds are if they do run it they run it for a good reason - a dependency change. Right?
> > > What needs to happen in the depend.sh`shell script is to pipe the output of the`CC{CFLAGS} -MM ${ALL_CSRC}`through an`awk(1)` command that will canonicalize the output in a C compiler independent way.
> > 
> > 
> > I'm not even sure how to go about this with the above in mind as well as the fact that it's only on some lines (for targets that have more than one line worth of rules). Given that this would only clutter the directory even more as well (which I mentioned yesterday) I'm also not sure it's needed?
> 
> Well when you start getting a _pile of backslashed lines_, it might be time to put them into a shell script. The question is if
> 
> * make limit_ioccc.sh
> * make all_ref
> * make seqcexit
> * make picky
> * make checknr
> * make man2html
> * make depend
> 
> are at the _pile of backslashed lines_ level? Some of them are not _pile of backslashed lines_, others might be, still others seem to be a _pile of backslashed lines_.

I'm not sure. I think it's probably worrying about minor details that don't matter so much. There were examples where this was true. The reset timestamp script comes to mind as a good example. But several (if not all) of the ones listed are not all that long and I think are good to be in the Makefile. Makes it so one doesn't have to open another script to see what it does and leaves scripts to more complicated issues. I mean I don't think running `picky` or `seqcexit` is all that complicated (for example) and it works well as part of the Makefile.

> 
> Remember that with issue #338 such shell scripts will be moved down into a sub-directory, hiding the clutter. So perhaps converting the rules that have become _pile of backslashed lines_ into invocations of a shell script is a reasonable option and keeps the Makefile for manageable? Maybe?

The part about moving scripts is true but for most of those listed I'm not sure it's necessary. I mean is a script for `depend` which is so few lines really necessary? Not counting the check for the tools it's only nine lines and three of those lines are messages. Is a six or nine line rule really significant enough to move to a script?

Even if it is once the repo is ready for use it'll not be something that's used much, make depend. Either way I'm not sure how to go about how to resolve the problem you describe because it only happens for objects that have more than one line worth of dependencies and only the lines beyond the first. I'm not an expert with awk though I feel like I should be better than I actually am (this comes down to priorities of course).

The only other thing I bring up is the `depend` rule protects against overwriting the Makefile and I feel like a script is more likely to mess that up - though not necessarily. 

Anyway if you wish to do this you can but my personal opinion is it's probably unnecessary for the reasons I listed. I would look at it except for the part above where I am not sure of how to determine the lines that need indentation and those that don't. I mean the rule runs directly on all the source files at one go so it would be difficult to determine I would think? On the other hand I would be curious how you would go about it simply as I might learn a new trick or two - if you think it's worth it instead of working on something else (or you only have time / energy for it maybe).

lcn2:
Perhaps this issue can be revisited when the work for issue #338 is being done.

xexyl:
> Perhaps this issue can be revisited when the work for issue #338 is being done.

That's a thought. It might be that it turns out that it's like our previous consensus - that it won't matter. But maybe something else will be thought of before then that will change this. I guess we can wait and see.

Have a great day! I'm going to do something else.

PR #354, opened 2022-10-07
Bug fix jstr_test.sh and improve jstr_test.8 + utf8_test.8 clarification + new man pages

xexyl:
There was a reference (or two counting the echo) in jstr_test.sh to ./jstrdecode when it should have been $JSTRDECODE.

The man page is more detailed on what the script does and why it's important.

This update was actually accidental as I had forgotten I already added it. This happened as follows: I looked at the GitHub page but after this man page had been added a new comment apparently was not made so I looked at the old list. I did not check if the man page was there and instead copied another script man page to this one and then edited it as if it did not exist. This turned out for the better though because I noticed the bug in jstr_test.sh and the man page is better now.


lcn2:
Thanks @xexyl

And we will review the remarks marked with 👀 when we return in a few days.

xexyl:
> Thanks @xexyl
> 

You’re most welcome!

> And we will review the remarks marked with 👀 when we return in a few days.

No rush or worries on my behalf. Just have a safe rest of your trip including your travelling home!

xexyl:
… I will soon have the laptop up again but I will not be able to focus on the repo for a while. However I do plan to make some commits or at least one commit with the changes I described in the txzchk thread. 

Hope you have a great Saturday!

PR #355, opened 2022-10-09
Complete txzchk_test.sh

xexyl:
This is a large update and perhaps worthy of a new version of txzchk but I've not done this as this would require even more changes and I need a rest.

The test script txzchk_test.sh now checks for expected errors for the bad files and if there is a difference it is flagged as an error. If there is an error file for files that are supposed to pass (good files) it is also an error.

Although the script does not check the exit code of txzchk it does check that if there's an error file (which as noted is enforced for bad files) that it matches exactly (yes this makes it inflexible but this is necessary). It also reports an issue if there was output to stderr but there's no error file (which would suggest that a good file failed when it should not have).

Removed from fnamchk and txzchk the -e option.

New option to txzchk -w always shows warnings. This is because the script requires warnings for bad test files but using -q would previously disable this (and we need -q for other reasons as well).

The script always passes -w -v 0 -q to txzchk and this cannot be changed without the error files being updated. These options were selected after careful thinking and experimenting with the different kinds of output. This means the script -v option is really only for the script itself. The -D option was removed.

In ioccc_test.sh specify ./ before the test_txzchk for the -d option. This is to make it consistent with the other options that specify a path and this is the way that the error files also have it. If one is to use an inconsistent path then this will cause an error. Perhaps the find(1) invocations should change slightly so that it acts at the directory it found so we don't have to worry about this but this can come at another time and not today (I only just thought of this but as noted I am too tired to continue working on this).

Updated txzchk.1, txzchk_test.8 and fnamchk.1.

Updated test_txzchk/README.md with information on how to generate the error files (this information is also in the man page for txzchk_test.sh). It looks like:

    for i in ./test_txzchk/bad/*.txt; do ./txzchk -q -v 0 -w -T -E txt "${i}" 2>"${i}".err; done

Note that this will overwrite all error files so this has to be done carefully (please see the man page for more information on this).

I'm not sure if I've added any test tarball files but all the bad files now have an associated error file.

Two exit codes that the script checks for from fnamchk(1) have had in fnamchk.c the comment /*ooo*/ added to make sure the error code remains the same. If this is the wrong comment to add I'm happy to change it to the correct one.

There might be other changes that I'm missing but hopefully this is a sufficient changelog.


xexyl:
And now for a much needed break that I think is also well deserved. Please feel free to add more test files. See the documentation on how to generate the error files. Please also let me know if you think of anything else that should be done or can be improved. As I noted in the changelog I'm very tired so I cannot do it today but this is great progress I think.

I'm not sure this will allow closing the txzchk issue thread but it's much closer to it now.

Good day!

lcn2:
There is a lot of good work here, @xexyl.  We look forward to reviewing this code in more detail when we return.

Hope your eyes and your backup drive are better.

We will be returning from Canada 🇨🇦 soon, recovering from the travel and getting unpacked and dealing with domestic stuff that was on hold while we were away.

xexyl:
> There is a lot of good work here, @xexyl. We look forward to reviewing this code in more detail when we return.
> 

Thank you! We can continue any thoughts about it here or the txzchk thread. I don't mind which.

> Hope your eyes and your backup drive are better.
> 

Both are better. Well the drive I did have to format but that's okay in this case. I'm not sure why this keeps happening though.

> We will be returning from Canada 🇨🇦 soon, recovering from the travel and getting unpacked and dealing with domestic stuff that was on hold while we were away.

Have a safe trip home and good luck with the at home stuff!

PR #359, opened 2022-10-11
Fix txzchk_test.sh under linux

xexyl:
A problem occurred in string_to_intmax2() in the order of checks for failures so that under linux the errno was set to non-zero but this was not triggered under macOS. This fix saves errno first (as an abundance of caution) and then checks the other conditions. Then after those are passed the saved errno is checked and if not 0 then it's used (setting errno to the saved value).

I also updated the test_txzchk/bad error file. This should solve the problem but if not I'll have to look at it tomorrow.

### DIFF
--- a/hostchk.sh
+++ b/hostchk.sh
@@ -136,6 +136,7 @@ if [[ $status -ne 0 ]]; then
     RUN_TAR_TEST=
 fi
 date > "$TEST_FILE"
+status="$?"
 if [[ $status -ne 0 ]]; then
     echo "$0: ERROR: date > $TEST_FILE exit code: $status" 1>&2
     EXIT_CODE=31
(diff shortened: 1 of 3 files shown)


xexyl:
Bedtime for me. Please let me know if this solves the problem for you too. Seems to on my end.

lcn2:
Thanks, @xexyl, hope to test soon.

xexyl:
Does it work okay for you ? It works for me and I tested again this morning.

xexyl:
> Thanks, @xexyl, hope to test soon.

Thanks. If not please let me know here or in the other thread - whichever you prefer - and I'll look at it more.

lcn2:
> Does it work okay for you ? It works for me and I tested again this morning.

See [comment 1275962334](https://github.com/ioccc-src/mkiocccentry/pull/359#issuecomment-1275962334).

xexyl:
> > Does it work okay for you ? It works for me and I tested again this morning.
> 
> See [comment 1275962334](https://github.com/ioccc-src/mkiocccentry/pull/359#issuecomment-1275962334).

Yep - I already saw it ... had started to reply before that came through. Thanks.

And sorry for the double commit on this. Unfortunately I was not on a different (non master) branch yesterday when fixing this and I relied on git commit --amend and unfortunately today when doing a pull earlier (just in case I missed an email about this) it pushed the old unfinished commit. So I decided I didn't like that so I instead amended that. That's what happens when you're in a rush and tired. Probably doesn't matter but it does bother me as it's an unnecessary commit.

Hope you had a good sleep. I had a rotten sleep again. Might try resting soon.

PR #360, opened 2022-10-12
Man page fixes + improvements to hostchk.sh and have_timegm.sh

xexyl:
I hope to get to the two threads later on but I'm fearing that I might not get much more done today. Still these are good changes with the caveat that I'll want to look again at the man page of the scripts to make sure the exit codes are consistent with the actual script.

Or perhaps we don't need to have all the exit codes? It would simplify the man pages. What do you think of that?

The other caveat is I am tired so it's possible I made a mistake with exit code in `have_timegm.sh` script - specifically being consistent with the `have_timegm.c` file. I thought I had it but now I'm not certain of it. Still if not it should be fixed but I think it's the right idea as the script should have consistent exit codes with the tool itself I think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -410,7 +410,7 @@ all: ${TARGETS} ${TEST_TARGETS}
 # rules, not file targets
 #
 .PHONY: all all_ref bug-report build checknr clean clean_generated_obj \
-        clean_mkchk_sem clobber configure depend hostchk install ioccc_test \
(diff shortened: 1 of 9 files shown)


xexyl:
Ugh. I see a mess here with previous commits. I'll convert to a draft maybe and recreate the repo again. Sigh. I won't be able to do that now though.

PR #361, opened 2022-10-12
Man page fixes + improvements to hostchk.sh and have_timegm.sh

xexyl:
I hope to get to the two threads later on but I'm fearing that I might not get much more done today. Still these are good changes with the caveat that I'll want to look again at the man page of the scripts to make sure the exit codes are consistent with the actual script.

Or perhaps we don't need to have all the exit codes? It would simplify the man pages. What do you think of that?

The other caveat is I am tired so it's possible I made a mistake with exit code in have_timegm.sh script - specifically being consistent with the have_timegm.c file. I thought I had it but now I'm not certain of it. Still if not it should be fixed but I think it's the right idea as the script should have consistent exit codes with the tool itself I think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -410,7 +410,7 @@ all: ${TARGETS} ${TEST_TARGETS}
 # rules, not file targets
 #
 .PHONY: all all_ref bug-report build checknr clean clean_generated_obj \
-        clean_mkchk_sem clobber configure depend hostchk install ioccc_test \
+        clean_mkchk_sem clobber configure depend hostchk bug-report install ioccc_test \
         legacy_clobber man2html mkchk_sem parser parser-o picky prep prep_clobber \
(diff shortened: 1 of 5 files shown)


lcn2:
Thanks for the good update @xexyl

xexyl:
> Thanks for the good update @xexyl

Thank you - happy to help! I guess you'll get to testing the txzchk test script under linux later on. I'll be leaving soon so I won't be able to do any fixes at that point but I can look at it tomorrow if necessary. Similar for replying to comments in the two threads - the hostchk and bug report ones.

lcn2:
Test code works under Red Hat Linux.

xexyl:
> Test code works under Red Hat Linux.

That’s great to hear - though I suspected it. 

I am just unwinding for a short bit and then I will get some sleep. I am exhausted today. Hope you have a great night!

PR #364, opened 2022-10-14
Additional man pages, man page fixes, shell script updates and fixes and typo fixes

xexyl:
For the updates to the scripts I'll let the logs speak for themselves. There is an amusing set of notes in the scripts though that I'll share here:

```sh
# NOTE: This script is called from make all and if it exits non-zero it will
# issue a warning, suggesting you report it via the bug_report.sh script. That
# script itself will call hostchk.sh which will reissue that warning (more than
# once in fact since make all calls hostchk.sh and bug_report.sh calls it via
# make all and running hostchk.sh directly) and each time it will suggest using
# bug_report.sh.
#
# Obviously one needn't run this script a second or third time just because the
# Makefile and bug_report.sh suggest you do! :-) Sorry in advance if this (to
# make use of an American English phrase which I will happily do for a pun for a
# pun not made is a wasted opportunity :-) ) throws you for a loop! :-) But now
# that you're in the loop it shouldn't even matter. :-)
#
```

A similar note is there in `bug_report.sh`. I noted this in one of the other threads.

There's also a reference to the old bug fixed:

```sh
# The below references to PIPESTATUS (instead of $?) fix the bug introduced in commit
# 8343c4b8cb97e52df64fe8973e68f0d83c6090e1 where the exit status of each command
# was not checked properly which meant that even if a test failed it would not
# be reported as an issue which rather defeated the purpose of this script.
#
# As amusing as the thought is that there's a bug in a script to help report
# bugs and issues, this bug was not in fact intentional. :-) I had thought of it
# earlier on but I neglected to address it before the commit as I moved on with
# adding the tests. However since the world definitely needs more reasons to
# laugh I'm keeping this comment here for the sake of humour and the irony that
# I caused. You're welcome! :-)
```

I'll be taking a break now. Not sure if I'll do anything more today here. Pretty tired but I feel good about what I got done at least.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,7 @@ SH_TARGETS=limit_ioccc.sh
 #     well if we directly referred to TARGETS.
 #
 MAN1_TARGETS= mkiocccentry txzchk fnamchk iocccsize chkentry jstrdecode jstrencode \
-	      jparse
+	      jparse bug_report hostchk
 MAN3_TARGETS= dbg
 MAN8_TARGETS= reset_tstamp verge limit_ioccc iocccsize_test ioccc_test run_usage utf8_test \
 	      have_timegm jparse_test txzchk_test run_flex run_bison vermod mkiocccentry_test \
--- a/bug_report.1
+++ b/bug_report.1
@@ -26,6 +26,10 @@ If it exits non-zero and it's not because of the \fB\-h\fP or \fB\-V\fP option i
 Even if the script exits 0 it does not necessarily mean there is no problem.
 If you feel there is a problem that was not detected you may still open an issue on the GitHub page (or email the Judges).
 The script will tell you how to do this whether there are issues detected or not.
+.PP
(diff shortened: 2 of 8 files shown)


lcn2:
Thanks @xexyl 

BTW: This commit was modified in order to resolve conflicts with commit a2cb8c529e1caa340c361281f93705370472fb4e.

xexyl:
> Thanks @xexyl
> 
> BTW: This commit was modified in order to resolve conflicts with commit [a2cb8c5](https://github.com/ioccc-src/mkiocccentry/commit/a2cb8c529e1caa340c361281f93705370472fb4e).

It’s funny you would say that because I had just opened the diff and was concerned that maybe it would have caused a conflict with what I did earlier today! So thank you for addressing that! It will save me the trouble tomorrow morning. 

Speaking of such looking at your commit and it looks very useful so thank you for that. Should help with consistency in exit code descriptions in scripts.

PR #368, opened 2022-10-17
Mass update man pages

xexyl:
This commit was in part an experiment with fixing formatting in some places in man pages. Possibly for some of the changes some files need to be updated still and there also definitely has to be more consistency updates but these can wait until another time. I went ahead and did certain things for a number of man pages as follows:

In the case of specific exit codes being in a list use .TP and .TQ macros. This makes it much cleaner and it aligns the two columns correctly. The .TQ macro might be used in one other man page in one place too though not sure which one that might be now.

In some cases I have used .I and .B instead of \fI..\fP and \fB..\fP but this is not consistent and in some cases it cannot be done this way (thinking of a full stop or dot as the very next character - it might be possible but I'm not sure how yet if it is).

Move URLs to their own line (one exception in txzchk.1). This is because sometimes (depending on the size of the screen) the url might be spread across more than one line and this can cause a problem. Thus before URLs I have added a line with the .br macro by itself for a line break.

The options should all use .TP now instead of .PP. Again I might have missed some files but most of them should have this now. This is cleaner.

Additionally for the files edited if an option arg is needed in the OPTIONS section I have made it in italic and refer to it in the description.

I've updated the dates on the man pages and hopefully all the ones I have edited now have the date of 17 October 2022.

Again much more has to be done but this is a good update on the man pages. Probably more needs to be discussed but this can wait. I did this commit as I did not feel up to anything else and it was something I could do that is needed (even if more can be done). One reason I did not try and finish it all is because it's almost certain I'll have to do it again and so better to go through things once in full than more than once.

### DIFF
--- a/bug_report.1
+++ b/bug_report.1
@@ -12,8 +12,8 @@ If no step fails it will tell you that everything is okay but still informs you 
 \fB\-h\fP
 Show help and exit.
 .TP
-\fB\-v\fP
-Set verbosity level.
+\fB\-v \fIlevel\fP\fP
+Set verbosity level to level.
 .TP
 \fB\-V\fP
 Show version and exit.
--- a/fnamchk.1
+++ b/fnamchk.1
@@ -1,4 +1,4 @@
-.TH fnamchk 1 "15 October 2022" "fnamchk" "IOCCC tools"
+.TH fnamchk 1 "16 October 2022" "fnamchk" "IOCCC tools"
 .SH NAME
 fnamchk \- IOCCC compressed tarball filename sanity check tool
 .SH SYNOPSIS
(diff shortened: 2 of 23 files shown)


xexyl:
.. and off to do other things. Not sure if I'll work on anything else today but this is a good update with lots of consistencies and format clean up.

lcn2:
Lots of good work here, @xexyl, thanks!

xexyl:
> Lots of good work here, @xexyl, thanks!

Thank you!

I hope to do more tomorrow preferably on `bug_report.sh` but we'll see.

Enjoy your afternoon and evening!

PR #369, opened 2022-10-18
Improve comment in dyn_array.h + mass fix man pages so all references to '-' have a '\' before them

xexyl:
For the latter commit I also did this with the `dbg` repo. This is the commit log for this repo:

    Fix - to \- in man pages
    
    Because I could not think of a way at this hour with such poor sleep to
    do it directly I just removed \ before all '-'s and then added them back
    like:
    
        sed -i -e 's/\\-/-/g' -e 's/-/\\-/g' *.[138]
    
    I then updated the date of the man page. This was started because of a
    mkiocccentry commit ce061e0e160004d2de7d8e5fd145bcb79ab7eb48:
    
        Made -h, -V and other -option in man pages consistent
    
        We are NOT sure of this is the best way, but
        when a -option is mentioned in the EXIT STATUS section
        of the man page, we put a \ in front of it.
    
        Example:
    
        \-h and help string printed or \-V and version string printed
    
    That is indeed the correct way. I fix them as I notice them but this
    time I just did a mass replace as above.
    
    In all likelihood this will have to be done again but for now they're
    all correct.

Although I'm probably not going to be able to I'm going to try resting now. Maybe I can do more later on.

### DIFF
--- a/bug_report.1
+++ b/bug_report.1
@@ -1,4 +1,4 @@
-.TH bug_report 1 "14 October 2022" "bug_report.sh" "IOCCC tools"
+.TH bug_report 1 "18 October 2022" "bug_report.sh" "IOCCC tools"
 .SH NAME
 bug_report.sh \- run a series of tests, collecting system information in the process, to help report bugs and issues
 .SH SYNOPSIS
@@ -20,7 +20,7 @@ Show version and exit.
 .SH EXIT STATUS
 If no issues are detected the script will exit 0.
 In this case it will tell you that all should be okay and you can safely delete the log file.
-If it exits non-zero and it's not because of the \fB\-h\fP or \fB\-V\fP option it will recommend you to report it, reminding you to attach the log in your report.
+If it exits non\-zero and it's not because of the \fB\-h\fP or \fB\-V\fP option it will recommend you to report it, reminding you to attach the log in your report.
 .SH NOTES
 .PP
 Even if the script exits 0 it does not necessarily mean there is no problem.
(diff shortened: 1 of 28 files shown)


xexyl:
I think you'll be particularly pleased with commit https://github.com/ioccc-src/mkiocccentry/pull/369/commits/f7e24a436f1fbbb179c8d9dba64799b41d07a696.

However I made a mistake with a git commit --amend and I'm having a merge problem so I might just fix the issue after you merge this into the master branch and then I can delete my branch and fix the changes. I might do otherwise but I'm really tired and need a break - and might not want to work more on it later on. Super tired.

xexyl:
...there, solved it. `git` can be really annoying sometimes. I would say it can be a git itself, if it was a person.

xexyl:
Doing something else now and I don't expect to return to this today.

xexyl:
... just fixed a typo that I will commit tomorrow hopefully with the other things discussed in the bug_report.sh thread.

I'll be shutting down now. I hope you have a great afternoon!

lcn2:
As previously mentioned, you might wish to view a few select man pages produced by the Berkeley folk (such as `/usr/share/man/man1` in macOS) such as `id(1)`, `ls(1)`, `dd(1)`, `true(1)`,`grep(1)`, and other section 1 man pages (both small and large in size).

Try to match the style and man macro method of such man pages.

BTW: We do not see using `\-` in those man pages.

xexyl:
> As previously mentioned, you might wish to view a few select man pages produced by the Berkeley folk (such as `/usr/share/man/man1` in macOS) such as `id(1)`, `ls(1)`, `dd(1)`, `true(1)`,`grep(1)`, and other section 1 man pages (both small and large in size).
> 
> Try to match the style and man macro method of such man pages.
> 
> BTW: We do not see using `\-` in those man pages.

I specifically read this in documentation about roff itself. It might also be in `man-pages(7)` under linux. Not sure about that.

xexyl:
Check https://man7.org/linux/man-pages/man7/groff_man_style.7.html under Portability:


       Similar caveats apply to escapes.  Some escape sequences are
       however required for correct typesetting even in man pages and
       usually do not cause portability problems.  Several of these
       render glyphs corresponding to punctuation code points in the
       Unicode basic Latin range (U+0000–U+007F) that are handled
       specially in roff input; the escapes below must be used to render
       them correctly and portably when documenting material that uses
       them syntactically—namely, any of the set ' - \ ^ ` ~
       (apostrophe, dash or minus, backslash, caret, grave accent,
       tilde).


I've seen it in other documentation as well.

lcn2:
See [comment 1282880789](https://github.com/ioccc-src/mkiocccentry/issues/275#issuecomment-1282880789) in regards to considering the Linux [groff_man_style.7.html](https://man7.org/linux/man-pages/man7/groff_man_style.7.html) as a possible replacement to BSD style man pages.

xexyl:
> See [comment 1282880789](https://github.com/ioccc-src/mkiocccentry/issues/275#issuecomment-1282880789) in regards to considering the Linux [groff_man_style.7.html](https://man7.org/linux/man-pages/man7/groff_man_style.7.html) as a possible replacement to BSD style man pages.

And please see my reply at https://github.com/ioccc-src/mkiocccentry/issues/275#issuecomment-1283092210.

PR #372, opened 2022-10-21
Scripts clean up wrt shellcheck + typo fixes

xexyl:


xexyl:
I'm hoping to do more in a bit but I'm taking a break now.

xexyl:
> Interesting updates, thanks @xexyl

Interesting how, please? I guess just that some of it's not necessary?

But I think it was worth at least fixing the checks that could be fixed. I don't like having to disable warnings and I know you're the same. I figured I would document a bit more the rationale behind the ones we cannot change too when I was working on this.

Perhaps the `-S` option is more useful though. Or then again perhaps not. I think it's probably good to have even if just to be complete. We won't use it but maybe it'll be useful to others. I guess time will tell.

Unrelated I have given some thought on how I might make it easier to update the man pages. However I won't be doing that today. Going to just take it easy for a bit. Then I'll get my dinner and figure out the evening.

Hope you have a good rest of your day!

lcn2:
> Interesting how, please?

Fixes and adjustments to please `shellcheck` are often interesting.  The `-S` is a nice idea for uses other than this repo too.

xexyl:
> > Interesting how, please?
> 
> Fixes and adjustments to please `shellcheck` are often interesting. The `-S` is a nice idea for uses other than this repo too.

Ah I see. Yes I can see what you mean by the fixes being interesting. And I agree that the `-S` option is a good idea for uses outside this repo. That is indeed what made me think of it a while back. By chance I had the thought and time and energy to do it today. I wasn't sure about the option letter but if this ever needs to change (or the semantics more generally) it can be done then. For now `-S` works.

Thank you for clarifying.

PR #374, opened 2022-10-22
Updates

xexyl:
Just a couple minor updates.

The bug_report.sh is obviously the only useful one (well the other one matters to me but as far as the development goes it's not important). 

I'll not be doing anything else today but I hope tomorrow I can do more.  Have a great night!

### DIFF
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -191,8 +191,10 @@ run_check 15 "make -v"
 # which tar: find the path to tar
 #
 # NOTE: we don't need to check if tar accepts the correct options in this script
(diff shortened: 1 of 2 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help! Hope you're having a nice sleep. Been awake a while but have been distracted so only now looking at the repo. In a bit I probably will take a break to shower and such. That might be very soon and in fact I rather plan it to be. Anyway the hope is after that I will be able to do some work here.

PR #376, opened 2022-10-26
New version of bug_report.sh

xexyl:
This is a big update which adds checks for all the tools we need as well as all the optional tools. This does which(1) on each tool but depending on if it's optional or not it will be reported differently if it's not found.

Add more thorough version checks as well. The comments in the script explain the process which I will repeat here but first an important point is that for some tools we do not try getting the version. For example echo supposedly has a --version option in linux but from quick tests it doesn't work and in macOS (and presumably BSD) it does not even have that option in the man page. Another example is cat because some options namely -v are valid and cat would block whether or not you consider -v harmful (as a certain person (in?)famously suggested (perhaps controversially and perhaps not) in USENIX Summer Conference Proceedings in 1983. Another example is 'yes' because trying `yes --version' would just repeatedly show --version on a line by itself as that's entirely the point of the command. What's worse with these is that since it's first redirected to stderr, only showing the output if it succeeds, it would appear to hang to the user.

As for what steps are taken for version it goes like this:

    # A question is how do to determine the version of a tool when there's no
    # universal option to get the version of _all_ tools (and in some tools we
    # cannot detect the version as it will block instead or in the case of echo or
    # yes just print the args).
    #
    # This is a good question and there's no guarantee we can obtain a version. In
    # that case we will report it as an unknown version. Nevertheless we go through
    # a series of steps as follows. First (step 0) we attempt to get the path to the
    # command and then we run the following on the path (if it appears to be a
    # built-in we cannot use the path of course).
    #
    #   1)  use --version (note: under macOS and BSD this will fail on several
    #       tools and in some tools it will fail under linux as well)
    #   2)  use -V (note: this might fail due to multiple reasons)
    #   3)  use -v (note: this might fail due to multiple reasons)
    #   4)  try what(1) (note: this is a macOS and seems available under some BSDs
    #       as well - but despite the man page saying it conforms to to IEEE Std
    #       1003.1-2001 ("POSIX.1") it is not available under linux)
    #   5)  try ident(1) (note: this appears to be a BSD command that's not
    #       available under macOS either)
    #   6)  try strings(1) with showing just the first 10 lines
    #
    # As soon as one of these returns a zero exit code (except for what(1) which is
    # described in the function below) we will stop. If we get through all steps
    # without any results we will mark it as an unknown version.  The first step to
    # succeed we will run a second time to record the output.
    #
    # Unfortunately this is far from perfect but we hope that it will help in a lot
    # of cases. A note about what(1) and ident(1) is that we will only check for it
    # once and if it does not exist we won't try it again. We also use the command
    # command -p in hopes to get the actual path. This also might not be perfect.
    #

As for optional tools it does not warn it is a problem except that run_bison.sh and run_flex.sh have to succeed as they use backup files if the tools cannot be used and not being able to use the backup tools would be a problem (the -o option is for the maintainers / developers of jparse - Landon and me).

For more details on optional tools missing and anything else read the comments. I added a lot for this purpose. Is bug_report.sh finished? I'm not sure. I think it's close to being finished but I'm not sure if it is or isn't finished entirely. Certainly hostchk.sh is not finished.

### DIFF
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -23,7 +23,7 @@ export TOOLS="./run_bison.sh ./run_flex.sh ./hostchk.sh ./dbg ./dyn_test ./fnamc
 	      ./reset_tstamp.sh ./txzchk ./txzchk_test.sh ./utf8_test ./verge
 	      ./vermod.sh ./jsemcgen.sh ./jsemtblgen"
 
-export BUG_REPORT_VERSION="0.3 2022-10-18"
+export BUG_REPORT_VERSION="0.4 2022-10-26"
 export FAILURE_SUMMARY=
 export WARNING_SUMMARY=
 export DBG_LEVEL="0"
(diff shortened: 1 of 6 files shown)


xexyl:
BTW: try the following:

```sh
grep 42 bug_report.sh
```

It was entirely coincidental but it works out perfectly. Actually here you go .. save you the effort so you can see it before you even merge and pull:

```sh
$ grep 42 bug_report.sh 
run_check 42 "make all" # the answer to life, the universe and everything conveniently makes all :-)
```

Really doing other things now. Good day!

xexyl:
(There's also some other humour in other comments of course .. not in any messages but comments.)

xexyl:
Enjoy https://github.com/ioccc-src/mkiocccentry/pull/376/commits/5663af80637bf8c75b9ee3c1aed5014777433a84!

:-)

lcn2:
We will look at the overall status, thanks @xexyl and best wishes on a better rest!

xexyl:
> We will look at the overall status, thanks @xexyl and best wishes on a better rest!

Thank you my friend! Who knows how tonight will go. But that's hours from now. Anyway I'm glad I got as much done today as I did. Not the only thing I got done today either so I feel pretty good considering that I woke up shortly after midnight.

Do make sure to look at the commit I mentioned above. It's just fun and it plays along with one of your tips for contestants - specifically about chocolate. It's in the comments only of course. 

Hope you have a great day!

xexyl:
> > We will look at the overall status, thanks @xexyl and best wishes on a better rest!
> 
> Thank you my friend! Who knows how tonight will go. But that's hours from now. Anyway I'm glad I got as much done today as I did. Not the only thing I got done today either so I feel pretty good considering that I woke up shortly after midnight.
> 
> Do make sure to look at the commit I mentioned above. It's just fun and it plays along with one of your tips for contestants - specifically about chocolate. It's in the comments only of course.
> 
> Hope you have a great day!

Going to get ready to sleep now but I am still curious about the status of the script. If you feel like it is finished feel free to close the issue. Or if you feel it’s finished but you think we should discuss it once more don’t close it. If you don’t feel it’s ready to be closed please let me know what else should be done (maybe you thought of something new even) but please do so in that thread. 

Of course we could also keep it open until everything else is resolved as maybe something new will come up. But we can discuss these things too. 

Good night my friend!

PR #378, opened 2022-10-27
New version of hostchk.sh + comments consistency update

xexyl:
Added test source files that must return specific values (and in the case of one print a certain string).

It's not modular yet. Whether this should be done or not I don't know. One benefit from that might be that we be that it would be easier to run all of these tests even if one fails. Currently when one fails no further tests are run.

There are several kinds of tests. First we obtain the value of ENOENT and then verify that setting errno = ENOENT will indeed be the value previously obtained. We then check for the string: "errno: $ENOENT" (whatever ENOENT might be: keep it in a variable for consistency check but also in case some system has a different value). We also print "Hello, world" and check that that was indeed printed. There are checks for return values 1 and 0 and maybe another one or two I've forgotten.

I'm not sure if this is all that needs to be done but it might be (other than possibly making it modular).

Updated .gitignore: added hostchk/. Even though I explicitly remove it when the script finishes and also via trap it's now in the .gitignore just in case something goes wrong.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,8 @@ verge.html
 oebxergfB.txt
 # bug report logs
 bug-report*.txt
+# hostchk.sh related glob
+hostchk/
 
 # Prerequisites
 *.d
--- a/chk.auth.head.c
+++ b/chk.auth.head.c
@@ -3,6 +3,16 @@
  *
  * "Because grammar and syntax alone do not make a complete language." :-)
  *
+ * This JSON parser was co-developed by:
+ *
+ *	@xexyl
+ *	https://xexyl.net		Cody Boone Ferguson
+ *	https://ioccc.xexyl.net
+ * and:
+ *	chongo (Landon Curt Noll, http://www.isthe.com/chongo/index.html) /\oo/\
+ *
+ * "Because sometimes even the IOCCC Judges need some help." :-)
+ *
  * The concept of this file was developed by:
  *
  *	chongo (Landon Curt Noll, http://www.isthe.com/chongo/index.html) /\oo/\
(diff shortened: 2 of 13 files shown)


xexyl:
Does this make `hostchk.sh` finished or should any other tests be added? If more should be done we can discuss it in that thread. If nothing else I guess I should make it modular unless it's okay that if one test compile fails (of the new tests I mean) that the script simply exits. I'm not sure if that's okay or not. I think ideally it would be better if it could run all the tests but if it's not strictly necessary maybe we can not worry about it: not sure.

I'm happy to do it but another time and probably not today - though it's possible that depending on how I feel I might go for it.

Been awake since really early again and now taking a break. I can't rest but I'm going to try relaxing anyway.

xexyl:
And just modularised it so that it will in fact continue if a test fails. Maybe this is all it needs then? I don't know. Going to do other things now ... good day!

lcn2:
Thank you very much for your work on consistency in this repo, @xexyl, that is very much approached!

See also [comment 1294147205](https://github.com/ioccc-src/mkiocccentry/issues/171#issuecomment-1294147205).

xexyl:
> Thank you very much for your work on consistency in this repo, @xexyl, that is very much approached!
> 
> See also [comment 1294147205](https://github.com/ioccc-src/mkiocccentry/issues/171#issuecomment-1294147205).

Thank you for being so supportive and appreciative! I really appreciate it! I know you also appreciate it so let’s not let it go so far as: it is turtles all the way down! 😀

I hope to reply to the other messages and then I will get ready to sleep. Been awake too many hours.

xexyl:
.. I have more consistency updates but I'm going to try resting now. Will continue later. I don't know if I'll manage anything else today but I expect at least these changes in some hours from now.

PR #380, opened 2022-10-28
Change cnt -> count

xexyl:
There are two reasons for this. First is so that it spells a word and does not trigger vim spellcheck (a bonus is it also is more descriptive). The other is it's too lexicographically close or similar to another word that one might accidentally type (or read) if one is tired or distracted. :-) This is probably not good.

Of course it might be wrong for me to write that in the log but I'm not sure where else to put it so there it is.

### DIFF
--- a/chkentry.c
+++ b/chkentry.c
@@ -247,23 +247,23 @@ main(int argc, char *argv[])
     struct json *auth_tree = NULL;	/* JSON parse tree for .author.json, or NULL ==> not parsed */
     bool info_valid = false;		/* .info.json is valid JSON */
     bool auth_valid = false;		/* .author.json is valid JSON */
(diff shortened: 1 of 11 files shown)


xexyl:
Of course part of it was also to make you laugh on your birthday but it's still a valid point that it could be a problem. 

Doing other things for a bit and then will be going to sleep. Either way enjoy the rest of your birthday! :-)

xexyl:
...right. I should have known that you'd be working on it. I kind of wondered about that. I'll do this later on. I do think it's a good idea though. Makes it clearer and it won't be as much of a risk at offending some people. I'd do it now but I'm not sure what else you have in the working so I'll close this and go for it another day.

Off now - well will be doing something else for a bit and then leaving. Enjoy the rest of your birthday my friend!

PR #382, opened 2022-10-30
Fix ioccc_test.sh and update chkentry.sh + other minor updates

xexyl:
Something I forgot to note is that in chkentry.c I removed extraneous spaces in messages.

A point of interest in that same commit is what I noted in comments: if rule 2a or rule 2b size is (are) out of range we only warn but we do not yet check for the override boolean. This is because I'm not as clear on the json semantics tests. Nevertheless now `ioccc_test.sh` passes so I changed `chkentry` to exit 1 in case there are validation errors.

I believe this makes `chkentry` itself complete but the `entry_util.c` functions should maybe (not sure) be updated. I'll let you work on that if you think that's necessary. Perhaps it's not though: not sure. It might be that simply warning will be enough since you'd also check the boolean. But I rather gather that this is not enough so you can change that. Perhaps doing so will also give me a better idea of the json semantics checks for those remaining man pages.


xexyl:
> Thanks @xexyl

Sure. I feel great about the progress I made today! All the more since I'm so drained from waking up so early this week. I really hope that changes sometime soon. But still at least it has given me some extra time to do things as I tend to leave early for other reasons (besides having to go to sleep earlier - as I said before no matter what time I go to sleep this tends to happen).

lcn2:
> Sure. I feel great about the progress I made today! 

Yes, you did!

xexyl:
> > Sure. I feel great about the progress I made today!
> 
> Yes, you did!

Thank you!

PR #383, opened 2022-10-30
Undo need for uudecode

xexyl:
Per discussions on GitHub I've rolled back the need for uudecode as the need for it is no longer valid after all. Thus the *.uu files are removed and the earlier files are back.

I also undid the man page changes as the -u option no longer is there.

Updated the good and bad directories (for general json files) to be under test_JSON/general.json to be clearer what it's for: the info.json is for .info.json files, the author.json is for .author.json file (which will be, it seems, changed to be .auth.json) and the general.json is for any json file.


xexyl:
If you would like me to look at the info and author json bad files I can also do that but I'm not sure I'll do it today. I'll be leaving soon probably.

lcn2:
Thank you, @xexyl.  Very good progress today!

We hope you can use this accomplishment to have a good rest.

xexyl:
> Thank you, @xexyl. Very good progress today!
> 

Thank you!

> We hope you can use this accomplishment to have a good rest.

Well I hope so. Time will tell I suppose. I'll be heading off soon probably though I'll be awake a while yet.

xexyl:
BTW with your update to .gitignore:


```sh
$ git --no-pager  log --merges |grep -c '^commit'
272
$ git --no-pager log |grep -c '^commit' # not sure if there's an option to exclude merges
2089
```

That's a lot of commits that aren't merges! I know a few of mine early on got put into one commit too so there are actually more than what it seems.

PR #386, opened 2022-11-01
Fix jparse_test.sh

xexyl:
The problem happened because of a series of incidents as follows. Originally the idea for the {auth,info}/bad was for both invalid data because it was invalid for chkentry (or originally jauthchk and jinfochk) but the other day it was decided that the files should be valid JSON but fail the semantics tests (only).

A while back I had the idea (to test the general json parser) of having other files that would not be tested by (what became) chkentry_test.sh but only jparse_test.sh. Now yesterday when I deleted the invalid JSON files it was decided they should be moved to the general.json subdirectory.

However as I had to hurry off in my head I decided to copy all the bad files to the bad directory since clearly they were bad. That's the problem. But some of the files are actually valid JSON so they should actually be in the good subdirectory!

Additionally I made a slight change to jparse_test.sh: add a summary (for string failures and file failures) so I could easily extract a list of files. To do this more easily I prefixed each line with either 'FILE:' or 'STRING:'. However this prefix could probably be removed now. Given that it helped with fixing the problem I have kept it in for now.


xexyl:
As for man pages commit I just realised two problems. Looking at them now. Namely the `checknr` and `man2html` rules now fail. I remembered install as that's the most important one but the others slipped my mind. Will fix that and commit.

xexyl:
I just ended up undoing it - for now. I have a backup of the Makefile so I can do it properly again when I have the time. If you feel like doing it before then that's okay too of course. Perhaps you have another way of going about it but I feel like making the rules work is a better option than just letting them be unusable for now.

xexyl:
Going to reform this pull request I think. Closing and will make a new one soon.

PR #390, opened 2022-11-05
Add check for IOCCC submit server readiness

xexyl:
If the submit server is not ready the user of mkiocccentry is told that
    because the submit server is not ready they may not enter the contest.
    It also asks them to please not email the Judges. If however the server
    is ready it tells them how to register as well as (after forming the
    tarball that's not a test entry) where they can upload their entry
    (again assuming they've registered).
    
    This commit makes these checks a bit more modular (well - the checks
    weren't actually in before except for test mode but the display of these
    things were there and it is this that is modular but the check for the
    submit server being ready was added to it in the two new functions).

That commit also fixed some typos and in commit d4e9f088c9aefe1be679927842a8e45b8de8bf34 there was also a typo fix (jparse.y).

### DIFF
--- a/jparse.l
+++ b/jparse.l
@@ -224,17 +224,17 @@ JSON_COMMA		","
  * and report their presence if we find any NUL bytes in data.
  *
  * We scan data for both NUL bytes.  When a NUL byte is detected, count
- * we increment the num_errors external varaible.
+ * we increment the num_errors external variable.
  *
  * We also keep track of newlines.  For the first MAX_NUL_REPORTED occurrences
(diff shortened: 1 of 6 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Sure. Let me know about the other thing. I imagine that'll take more time to look at! I'm sorry you're awake though.

I'll be leaving for now though I'm not sure why since I can already tell I won't be able to rest :( still I have to try.

PR #401, opened 2022-11-08
Make json lexer/scanner re-entrant

xexyl:
With this commit the lexer is now also re-entrant (with two caveats and
one todo below)!

This required quite a few changes with both the scanner and the parser
including use of getter and setter functions as well as a local variable
of yyscan_t (actually a void *). In the case of the YY_USER_ACTION macro
it seems the use of the getter macros are not needed but everywhere else
is.

The first caveat is that the filename is still a global. This should be
fixed. If we want to be strictly technical the num_errors variable also
is global and probably should be fixed.

The second is that it appears that yydebug is always global and cannot
be enabled or disabled per scanner/parser instance.

The todo (and right now a caveat) is that now run_flex.sh has to
generate a new file: jparse.lex.ref.h and then copy it to jparse.lex.h
with sorry.tm.ca.h prepended to it just like the run_bison.sh script
does for jparse.tab.ref.h and jparse.tab.h. Once that is done the
Makefile can be updated for this rule. Right now jparse.y includes
jparse.lex.ref.h but once the script has been updated this needs to be
changed to jparse.lex.h. It cannot be included in jparse.h.

Due to the nature of this commit I have it in a different branch so that
any problems can be sorted. However something I just remembered is that
earlier today I made some commits of some of these files on the updates
branch so perhaps this is a problem. If it is I will correct it.


xexyl:
Landon would you please take care of the todo I noted ? That would be appreciated. Maybe I can get to it later on if you don't and maybe that would be before you even get to it but it would be appreciated at this time anyway. Going to take a break but I feel good about it. I'll write about it in the other thread and then it's break time.

lcn2:
Best wishes on your rest.  Thank you @xexyl

xexyl:
> Best wishes on your rest. Thank you @xexyl

Thank you also! I'm just taking it easy ... can't rest but I don't want to rest right now - not lie down anyway.

Hope you have a great day!

PR #402, opened 2022-11-09
Fix segfault in chkentry and memory leak in jparse

xexyl:
Under macOS this problem did not show up but under linux it did. The problem is that chkentry also called clearerr_or_fclose() on the streams of the .info.json and .auth.json files but because the scanner also uses this it resulted in a double free. I'm not sure exactly why this did not happen before since it was always the same stream in the end but possibly because yyin was used it was in a different state. As I noted in either a commit log or a GitHub comment we can no longer use yyin due to complexities introduced with the re-entrant scanner wrt setting/getting yyin needing the scanner context (which we don't always have): specifically the function that opens the json stream (or stream that is supposed to be json) does not have the scanner. Instead the function that that function calls has the scanner context so we can't use yyin. We obviously have to close it in the scanner since at that point it'll always be reached but we don't always use chkentry so there would be a file descriptor leak.

### DIFF
--- a/chkentry.c
+++ b/chkentry.c
@@ -659,16 +659,13 @@ main(int argc, char *argv[])
     }
 
     /*
-     * cleanup - except for info_path and auth_path
+     * cleanup - except for info_path, auth_path, info_stream and auth_stream.
+     *
+     * We don't try closing info_stream or auth_stream because the json parser
+     * already does that at this point. This is because we no longer use yyin
+     * due to complications introduced when making the lexer re-entrant. This is
+     * not a problem under macOS but it is under linux.
      */
-    if (info_stream != NULL) {
-	clearerr_or_fclose(info_stream);
-	info_stream = NULL;
-    }
-    if (auth_stream != NULL) {
-	clearerr_or_fclose(auth_stream);
-	auth_stream = NULL;
-    }
     if (info_tree != NULL) {
 	json_tree_free(info_tree, JSON_DEFAULT_MAX_DEPTH);
 	info_tree = NULL;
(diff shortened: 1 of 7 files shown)


lcn2:
While this is a good catch, and the fix looks correct @xexyl,
we request that you submit a pull request to order the argument `filename` from:

```c
extern struct json *parse_json(char const *filename, char const *ptr, size_t len, bool *is_valid);
extern struct json *parse_json_stream(char const *filename, FILE *stream, bool *is_valid);
```

to:

```c
extern struct json *parse_json(char const *ptr, size_t len, char const *filename, bool *is_valid);
extern struct json *parse_json_stream(FILE *stream, char const *filename, bool *is_valid);
```

The `filename`, while important for error messages, is not the primary object that is being parsed.
The `is_valid` at the end if also a useful convention for _returning a boolean_ as a final argument.
Please reorder so that the lead arguments of those functions focus on the key object being parsed.

xexyl:
> While this is a good catch, and the fix looks correct @xexyl, we request that you submit a pull request to order the argument `filename` from:
> 
> ```c
> extern struct json *parse_json(char const *filename, char const *ptr, size_t len, bool *is_valid);
> extern struct json *parse_json_stream(char const *filename, FILE *stream, bool *is_valid);
> ```
> 
> to:
> 
> ```c
> extern struct json *parse_json(char const *ptr, size_t len, char const *filename, bool *is_valid);
> extern struct json *parse_json_stream(FILE *stream, char const *filename, bool *is_valid);
> ```
> 
> The `filename`, while important for error messages, is not the primary object that is being parsed. The `is_valid` at the end if also a useful convention for _returning a boolean_ as a final argument. Please reorder so that the lead arguments of those functions focus on the key object being parsed.

That's fair enough. I can do that certainly - but not right this moment.

xexyl:
... actually I will do it now ... won't be long. I will be going afk briefly and then will do it.

xexyl:
Done.

PR #403, opened 2022-11-09
Reorder parse_json and parse_json_stream args

xexyl:
The diff for the parameter changes:

    -extern struct json *parse_json(char const *filename, char const *ptr, size_t len, bool *is_valid);
    -extern struct json *parse_json_stream(char const *filename, FILE *stream, bool *is_valid);
    +extern struct json *parse_json(char const *ptr, size_t len, char const *filename, bool *is_valid);
    +extern struct json *parse_json_stream(FILE *stream, char const *filename, bool *is_valid);

### DIFF
--- a/chkentry.c
+++ b/chkentry.c
@@ -423,7 +423,7 @@ main(int argc, char *argv[])
      * parse .info.json if it is open
      */
     if (info_stream != NULL) {
-	info_tree = parse_json_stream(info_path, info_stream, &info_valid);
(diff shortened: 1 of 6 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

No problem at all. I have a quick comment in the man page issue and then I'll be doing something else entirely.

PR #405, opened 2022-11-09
Fix bug_report.sh

xexyl:
Since have_timegm.sh no longer exists bug_report.sh failed because it tried to run it along with all the other tools.

### DIFF
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -27,7 +27,6 @@ export TOOLS="
     ./chkentry
     ./dbg
     ./fnamchk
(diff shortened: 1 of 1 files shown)


lcn2:
good catch @xexyl

xexyl:
> good catch @xexyl

I thought of it earlier and then I forgot about it ... but decided to try running it. I'm finding it's very useful to try before doing any commits! It's helped me uncover a number of problems with changes of different kinds (some platform specific and some not).

Issue #407, opened 2022-11-10
Bug report: jstr_test.sh failure due to test/ subdirectory


xexyl:
NOTE: I have a fix for the main problem and I'll be committing it but I'm going to document everything anyway. I'll say what is fixed at the end.

## Describe the bug

There are a number problems actually depending on how it's invoked.

I am deferring it to the relevant output as it says it all.

## To Reproduce

0.  See below.

## Expected behaviour

If it cannot find the files needed it should fail. This has not been fixed.

But it should also not fail in the first place as the script should be able to find the script itself. See below for fix which I'll be committing in a bit. This HAS been fixed though just like some of the other scripts under `test/` if one runs it directly under `test` the script will fail probably because of the block:

```sh
elif [[ -f ../mkiocccentry.c ]]; then
    cd ..
    status="$?"
    if [[ $status -ne 0 ]]; then
        echo "$0: ERROR: cd .. exit code: $status" 1>&2
        exit 3
    fi
    TOPDIR="$PWD"
    if [[ $V_FLAG -ge 3 ]]; then
        echo "$0: debug[3]: assume TOPDIR is ..: $TOPDIR" 1>&2
    fi

```

which will change to the parent directory which means that `./jstr_test.sh` no longer is the right name.

## Relevant Output

For example try running it from the root directory. THIS HAS BEEN FIXED! I am only including it so you can see what is fixed in an upcoming commit.

```sh
test/jstr_test.sh -Z /Users/cody/code/xexyl-mkiocccentry
rm -f test/jstr_test.out test/jstr_test2.out
test/jstr_test.sh: about to run test #0
./jstrencode: Beginning jencchk test on code JSON encode/decode functions ...
./jstrencode: ... passed JSON encode/decode test
test/jstr_test.sh: test #0 passed
test/jstr_test.sh: about to run test #1
./jstrencode -v 0 -n < ./jstrencode | ./jstrdecode -v 0 -n > test/jstr_test.out
test/jstr_test.sh: test #1 passed
test/jstr_test.sh: about to run test #2
./jstrencode -v 0 -n < ./jstrdecode | ./jstrdecode -v 0 -n > test/jstr_test.out
test/jstr_test.sh: test #2 passed
test/jstr_test.sh: about to run test #3
cat $SRC_SET | ./jstrencode -v 0 -n | ./jstrdecode -v 0 -n > test/jstr_test.out
cat: jstr_test.sh: No such file or directory
[...]

PR #408, opened 2022-11-10
Bug fixes and enhancements

xexyl:
Several scripts under test did not show the version right in the case of bug_report.sh. Specifically I removed the 1>&2 from the echo of the version as with that there it showed a different and confusing output. For example:

    ## Checking if ./test/txzchk_test.sh is executable
    ## ./test/txzchk_test.sh IS EXECUTABLE
    ## RUNNING: ./test/txzchk_test.sh -V
    0.5 2022-11-04
    ./test/txzchk_test.sh version

Notice the last two lines where the version by itself is printed on a line and then below it it just says the name of the script followed by version followed by a space and then newline. It now looks like:

    ## Checking if ./test/txzchk_test.sh is executable
    ## ./test/txzchk_test.sh IS EXECUTABLE
    ## RUNNING: ./test/txzchk_test.sh -V
    ./test/txzchk_test.sh version 0.5 2022-11-04

which is cleaner and clearer.

Updated bug_report.sh version. It now has a new function get_version_minimal which tries to get the version of a tool in a minimal way. Consider cat(1). Using -v is problematic because it blocks (it's an option and then expects input from stdin). -V might or might not exist. --version also might or might not exist. What we try on it is --version and then also (if available) what(1) and (if available) ident(1). If it does not find anything beyond that it returns and does not even warn that there's an unknown version as it's almost certainly not important and would just clutter the log (we don't use strings(1) on it for that reason as well).

The bug_report.sh script also has had some clarifying comments (and some humour in comments as well) added. There possibly were other things with the script.

The ioccc_test.sh did not use -Z option for jstr_test.sh which caused some problems. More importantly jstr_test.sh itself was fixed so that raw references (for when needing the filename itself to cat) of jstr_test.sh were changed to $0. This fixes a problem I reported (#407) but there still is another problem with that script.

There possibly were other changes but these are the important ones if any others were in fact made.

### DIFF
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -57,7 +57,7 @@ export TOOLS="
     ./vermod.sh
     "
 
-export BUG_REPORT_VERSION="0.6 2022-11-05"
+export BUG_REPORT_VERSION="0.7 2022-11-10"
 export FAILURE_SUMMARY=
 export WARNING_SUMMARY=
 export DBG_LEVEL="0"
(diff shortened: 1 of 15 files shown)


lcn2:
Good mods @xexyl thanks

xexyl:
> Good mods @xexyl thanks

Happy to help! Hope you can go back and get some sleep.

PR #409, opened 2022-11-10
Fix bug in jstr_test.sh - resolves #407

### LINKED ISSUE #407: Bug report: jstr_test.sh failure due to test/ subdirectory

xexyl:
Now the script checks for failure of test 3. I'm not sure if there are any conditions that might not be picked up. For example if another part of the pipeline fails it might not be detected but at least if one of the files does not exist (or cannot be found) the test will fail.

### DIFF
--- a/entry_util.c
+++ b/entry_util.c
@@ -1680,7 +1680,7 @@ object2manifest(struct json *node, unsigned int depth, struct json_sem *sem,
 
 
 /*
- * form_tar_filename - return an malloced tarball filename
+ * form_tar_filename - return a malloced tarball filename
  *
  * given:
  *	IOCCC_contest_id	IOCCC contest UUID or test
--- a/json_sem.c
+++ b/json_sem.c
@@ -1287,7 +1287,7 @@ sem_member_value_int(struct json const *node, unsigned int depth, struct json_se
  * and if it is a JTYPE_NUMBER (JSON number), return a pointer to a time_t
  * or return NULL on error or invalid input.
  *
- * If the JTYPE_NUMBER (JSON number) cannot be returned as an time_t, NULL is returned.
+ * If the JTYPE_NUMBER (JSON number) cannot be returned as a time_t, NULL is returned.
  *
  * given:
  *	node	JSON parse node being checked
@@ -1298,7 +1298,7 @@ sem_member_value_int(struct json const *node, unsigned int depth, struct json_se
  *		NULL ==> do not report a JSON semantic validation error
  *
  * returns:
- *	!= NULL ==> decoded JTYPE_NUMBER as an time_t from the value part of JTYPE_MEMBER
+ *	!= NULL ==> decoded JTYPE_NUMBER as a time_t from the value part of JTYPE_MEMBER
  *	    The val_err arg is ignored
  *	NULL ==> invalid arguments or JSON conversion error
  *	    If val_err != NULL then *val_err is JSON semantic validation error (struct json_sem_val_err)
--- a/test/jstr_test.sh
+++ b/test/jstr_test.sh
@@ -204,16 +204,26 @@ echo "cat \$SRC_SET | $JSTRENCODE -v $V_FLAG -n | $JSTRDECODE -v $V_FLAG -n > $T
 # note: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead. [SC2002]
 # shellcheck disable=SC2002
 cat $SRC_SET | "$JSTRENCODE" -v "$V_FLAG" -n | "$JSTRDECODE" -v "$V_FLAG" -n > "$TEST_FILE"
-# We cannot double quote "$SRC_SET" because it would make the shell think it's a
-# single file which of course does not exist by that name as it's actually a
-# list of files. Thus we disable shellcheck check SC2086.
-# shellcheck disable=SC2086
-cat $SRC_SET > "$TEST_FILE2"
-if cmp -s "$TEST_FILE2" "$TEST_FILE"; then
-    echo "$0: test #3 passed"
-else
+STATUS="$?"
+if [[ "$STATUS" -ne 0 ]]; then
     echo "$0: test #3 failed" 1>&2
     EXIT_CODE=45
+else
+    # We cannot double quote "$SRC_SET" because it would make the shell think it's a
+    # single file which of course does not exist by that name as it's actually a
+    # list of files. Thus we disable shellcheck check SC2086.
+    # shellcheck disable=SC2086
+    cat $SRC_SET > "$TEST_FILE2"
+    STATUS="$?"
+    if [[ "$STATUS" -ne 0 ]]; then
+	echo "$0: test #3 failed" 1>&2
+	EXIT_CODE=45
+    elif cmp -s "$TEST_FILE2" "$TEST_FILE"; then
+	echo "$0: test #3 passed"
+    else
+	echo "$0: test #3 failed" 1>&2
+	EXIT_CODE=45
+    fi
 fi
 
 # All Done!!! -- Jessica Noll, Age 2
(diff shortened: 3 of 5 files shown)


lcn2:
Thanks for this test update, @xexyl   

We have an improvement suggestion for that test you modified. Instead of checking the value of just `$?`, which tests just the status of the final command in the pipe, you may test the exit status of several both commands in the pipe. 

See [this tutorial](https://www.cyberciti.biz/faq/unix-linux-bash-find-out-the-exit-codes-of-all-piped-commands/) for the basic concepts.

xexyl:
> Thanks for this test update, @xexyl
> 
> We have an improvement suggestion for that test you modified. Instead of checking the value of just `$?`, which tests just the status of the final command in the pipe, you may test the exit status of several both commands in the pipe.
> 
> See [this tutorial](https://www.cyberciti.biz/faq/unix-linux-bash-find-out-the-exit-codes-of-all-piped-commands/) for the basic concepts.

It's funny you'd say that as I actually thought I might end up doing that but I had other things to do. I'll take a look at it tomorrow. I have an idea how to do it but maybe the document will be something interesting too.

xexyl:
.. ah yes, PIPESTATUS. I was thinking of that of course. Tomorrow.

lcn2:
> .. ah yes, PIPESTATUS. I was thinking of that of course. Tomorrow. 

Best wishes on a good sleep 💤

xexyl:
> > .. ah yes, PIPESTATUS. I was thinking of that of course. Tomorrow.
> 
> Best wishes on a good sleep 💤

Thank you. For now just doing other things. Will get food in a bit. Hope you have a great rest of your day!

xexyl:
> 
> ### UPDATE 0
> We merged this pull request. But you might wish to use the above method to test the exit status of both `jstrencode` and `jstrdecode` to submit another pull request.

Sure. As I said I thought of it earlier on. It does seem to work as is but it would be better to test all parts of the pipeline. It is what I referred to in the log even. I can look at it tomorrow. I am just too tired to do it now.

lcn2:
> > ### UPDATE 0
> > We merged this pull request. But you might wish to use the above method to test the exit status of both `jstrencode` and `jstrdecode` to submit another pull request.
> 
> Sure. As I said I thought of it earlier on. It does seem to work as is but it would be better to test all parts of the pipeline. It is what I referred to in the log even. I can look at it tomorrow. I am just too tired to do it now.

Best wises on your rest.

xexyl:
> > > ### UPDATE 0
> > > We merged this pull request. But you might wish to use the above method to test the exit status of both `jstrencode` and `jstrdecode` to submit another pull request.
> > 
> > 
> > Sure. As I said I thought of it earlier on. It does seem to work as is but it would be better to test all parts of the pipeline. It is what I referred to in the log even. I can look at it tomorrow. I am just too tired to do it now.
> 
> Best wises on your rest.

Thank you! I slept in a little bit at least. Been up a while. My dogs need to go out and I hope after that I'll get some things done .. though if I feel like I can rest I'll do that first. Hope you're sleeping well!

PR #411, opened 2022-11-15
Add dyn_array man page

xexyl:
The dyn_array.md file has yet to be written. There are at least two other things that I will briefly mention on GitHub as they should also be done before creating a separate repo but this is a good start.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ SH_TARGETS=limit_ioccc.sh
 #
 MAN1_TARGETS= man/mkiocccentry man/txzchk man/fnamchk man/iocccsize man/chkentry man/jstrdecode man/jstrencode \
 	      man/jparse man/bug_report man/hostchk man/run_flex man/run_bison
-MAN3_TARGETS= man/dbg
+MAN3_TARGETS= man/dbg man/dyn_array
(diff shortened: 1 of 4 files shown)


xexyl:
Okay quickly as I'm very tired and I need to take a break. There are some things to note besides the fact there's no readme for the facility yet.

* Towards the end I was getting very tired and so in trying to hurry up you might notice that not all the macros (in description) refer to the parameters. I'm not sure if any should mention them or if they should all mention them but either way something has to change.
* For example I only referred to the two test files: the header file and the source file. Part of this is as I said I need to leave but another part is because right now it's dependent on the dbg repo also so I'm not even sure what to do about that. Perhaps it's better an example is not made until that has been resolved.

I'm almost certain there was at least one more note but I don't know what it is now .. much too tired. I need to take a break. I'm not sure if I'll return to this today. Hope you're sleeping well! I'm hoping that I can rest a bit very soon myself.

lcn2:
Thank you @xexyl

xexyl:
> Thank you @xexyl

You're very welcome. We can talk about the other issues I brought up in the other threads later. I'll be shutting down the laptop soonish. Just taking it easy a bit more. Hope you have a great night!

PR #414, opened 2022-11-17
Rename ioccc_test/ to test_ioccc/ + bug fixes

xexyl:
As suggested on GitHub I have renamed the subdirectory. Because of the rename there are a lot of changes in this commit.

I also fixed some failures of some scripts that occurred under some systems and conditions.


xexyl:
Sorry for the delay .. I mean I know it's not a big deal but I did indeed end up going afk a short bit. This should be good now but I'm happy to fix any problems that might show up due to the nature of the changes.

Doing something else now.

lcn2:
Thanks 🙏 @xexyl

xexyl:
> Thanks 🙏 @xexyl

Sure .. let me know if you encounter any issues. I'm pretty tired right now but I should be here a while yet. Was looking at man2html related things. Will post a link in that thread and then going to do something else entirely.

PR #415, opened 2022-11-18
Make make silent if nothing to be done + add option nounput to scanner

xexyl:


lcn2:
good idea @xexyl

xexyl:
> good idea @xexyl

Yes. Makes it a bit cleaner in output.

Hope you had a nice sleep. I'm unfortunately having a difficult day. Was getting up (or standing up) and had a spasm so my back is a mess right now. Important to be sitting up but it's not so easy. Hopefully tomorrow will be better.

That's not why I've not done anything else today mind you but it's not pleasant either. Hope you have a great day!

Issue #417, opened 2022-11-20
Bug report: make test fails on RedHat Linux 7.9


lcn2:
## Describe the bug

On Linux (RH7.9) `make clobber all test` fails.

## To Reproduce

Steps to reproduce the behavior (adjust steps as needed):

0. make clobber all
1. make test

## Expected behavior

After `make clobber all`, the `make test` should pass and exit 0.

## Relevant Output

After a successful `make clobber all`, the `make test` fails:

```
$ make test
make[1]: Entering directory `/var/tmp/mkiocccentry/dbg'
make[1]: Leaving directory `/var/tmp/mkiocccentry/dbg'
make[1]: Entering directory `/var/tmp/mkiocccentry/soup'
make[1]: Leaving directory `/var/tmp/mkiocccentry/soup'
make -C test_ioccc test
make[1]: Entering directory `/var/tmp/mkiocccentry/test_ioccc'
cc  -I.. utf8_test.c -c
cc  -I.. utf8_test.o ../utf8_posix_map.o ../dbg/dbg.o ../util.o ../dyn_array.o -o utf8_test
cc  -I.. jnum_chk.c -c
cc  -I.. jnum_test.c -c
In file included from jnum_test.c:26:0:
../dbg/dbg.h:55:23: error: two or more data types in declaration specifiers
 typedef unsigned char bool;
                       ^
In file included from ../dyn_array.h:52:0,
                 from ../util.h:42,
                 from ../json_parse.h:26,
                 from jnum_test.c:33:
../dbg/dbg.h:55:1: warning: useless type name in empty declaration [enabled by default]
 typedef unsigned char bool;
 ^
In file included from jnum_test.c:26:0:
../util.h:53:23: error: two or more data types in declaration specifiers
 typedef unsigned char bool;
                       ^
In file included from ../json_parse.h:26:0,
                 from jnum_test.c:33:
../util.h:53:1: warning: useless type name in empty declaration [enabled by default]
 typedef unsigned char bool;
 ^
make[1]: *** [jnum_test.o] Error 1
make[1]: Leaving directory `/var/tmp/mkiocccentry/test_ioccc'
make: *** [test] Error 2
```

## bug report attachment

vvvvvvvvvvvvvvvvvvvvvvvvv --- Drag the file into the area below, please!

[bug-report.20221119.214019.txt](https://github.com/ioccc-src/mkiocccentry/files/10048954/bug-report.20221119.214019.txt)

[...]

PR #418, opened 2022-11-20
Fix obscure bug with bool typedef (#417) + more restructuring

xexyl:
The problem as reported is that under some systems running make clobber all test would result in a compilation error:

    In file included from jnum_test.c:26:0:
    ../dbg/dbg.h:55:23: error: two or more data types in declaration specifiers
     typedef unsigned char bool;
			   ^
    In file included from ../dyn_array.h:52:0,
		     from ../util.h:42,
		     from ../json_parse.h:26,
		     from jnum_test.c:33:
    ../dbg/dbg.h:55:1: warning: useless type name in empty declaration [enabled by default]
     typedef unsigned char bool;
     ^
    In file included from jnum_test.c:26:0:
    ../util.h:53:23: error: two or more data types in declaration specifiers
     typedef unsigned char bool;
			   ^
    In file included from ../json_parse.h:26:0,
		     from jnum_test.c:33:
    ../util.h:53:1: warning: useless type name in empty declaration [enabled by default]
     typedef unsigned char bool;
     ^
    make[1]: *** [jnum_test.o] Error 1
    make[1]: Leaving directory `/var/tmp/mkiocccentry/test_ioccc'
    make: *** [test] Error 2

This is because bool was already defined and so the 'typedef unsigned char bool;' was a problem. The solution is in a number of files add an ifdef guard so that bool is only defined once - either from #including stdbool.h or from the typedef. Also as for util.h we have to move the inclusion of dyn_array.h so that it's below the boolean define as otherwise we will get another issue in some systems.

This commit involves changes to some files that will possibly require changes elsewhere - iocccsize.h (which I won't touch normally) and dbg.h (which I'll do a pull request in the dbg repo).


xexyl:
There you go. Now I really am going to rest!

xexyl:
... making progress on jparse but I have to take a break - have some other things I need to do first. I hope to return to it after I'm back at the laptop. It does work it appears but some clean up is definitely required and more thorough testing.

xexyl:
> ... making progress on jparse but I have to take a break - have some other things I need to do first. I hope to return to it after I'm back at the laptop. It does work it appears but some clean up is definitely required and more thorough testing.

I'm very frustrated with this. I make progress on one part but then it causes a problem elsewhere. I then fix that and it's another problem. I think I might end up starting over with it. However right now I don't feel up to that - and maybe not it at all. I'm contemplating the possibilities and one of those possibilities is not worrying about this until tomorrow or the next day. The fact I got the bug sorted and the dynamic array facility moved to a subdirectory is good progress for the day.

xexyl:
..starting over was it seems a good idea. I just have to test it under linux and if all is good I can commit. I didn't move jstrencode or jstrdecode to jparse though. That can happen another time. I'm too scattered right now and need to finish up for the day (at least for the day here - for now or for the rest of the day).

xexyl:
> ..starting over was it seems a good idea. I just have to test it under linux and if all is good I can commit. I didn't move jstrencode or jstrdecode to jparse though. That can happen another time. I'm too scattered right now and need to finish up for the day (at least for the day here - for now or for the rest of the day).

A minor problem I've run into. I'm going to take a short break and then I hope to tackle it.

lcn2:
> > ..starting over was it seems a good idea. I just have to test it under linux and if all is good I can commit. I didn't move jstrencode or jstrdecode to jparse though. That can happen another time. I'm too scattered right now and need to finish up for the day (at least for the day here - for now or for the rest of the day).
> 
> 
> 
> A minor problem I've run into. I'm going to take a short break and then I hope to tackle it. 

We merged the pull request as it stood now. Thanks @xexyl

xexyl:
> > > ..starting over was it seems a good idea. I just have to test it under linux and if all is good I can commit. I didn't move jstrencode or jstrdecode to jparse though. That can happen another time. I'm too scattered right now and need to finish up for the day (at least for the day here - for now or for the rest of the day).
> > 
> > 
> > A minor problem I've run into. I'm going to take a short break and then I hope to tackle it.
> 
> We merged the pull request as it stood now. Thanks @xexyl

Sure. It might take a bit of work to get my current updates (when they're ready) put in place but I'll work it out.

PR #420, opened 2022-11-20
Makefiles clean up and fixes

xexyl:
Fix 'No rule to make target `jparse/Makefile', needed by `jparse'' in Makefile.

Fix strange issue where running make clobber would pass to rm -f -f instead of just -f. I'm not even sure why this happened - I don't have the time or energy to try figuring it out. It seems odd because there's no reference to -f -f in any file. It did not occur with the dbg subdirectory but it does with soup and test_ioccc/. The fix was removing the single -f from the rm invocation in those Makefiles. It's very funny because even after removing the -f from the invocation in e.g. soup/Makefile if one does `cd soup; make clobber' you'll see that it does in fact run rm -f.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -417,7 +417,7 @@ jparse.tab.o: jparse.tab.c Makefile
 jparse_main.o: jparse_main.c Makefile
 	${CC} ${CFLAGS} jparse_main.c -c
 
-jparse: jparse/jparse.h jparse/jparse.l jparse/jparse.y jparse/Makefile
+jparse: jparse/jparse.h jparse/jparse.l jparse/jparse.y
 	@${MAKE} -C jparse CFLAGS="${CFLAGS} -Wno-unused-function -Wno-unneeded-internal-declaration"
 	@${CP} -f jparse/jparse.1 man/jparse.1
 	@${CP} -f jparse/jparse_test.8 man/jparse_test.8
@@ -806,7 +806,7 @@ legacy_clean:
 	${RM} -f jauthchk.o jinfochk.o
 
 clean: clean_generated_obj legacy_clean
-	${RM} -f ${OBJFILES}
+	${RM} -f ${OBJFILES} *.o
 	${RM} -f ${GENERATED_OBJ}
 	${RM} -f ${LESS_PICKY_OBJ}
 	${RM} -rf ${DSYMDIRS}
@@ -825,6 +825,7 @@ clobber: clean prep_clobber
 	${MAKE} -C dbg clobber
 	${MAKE} -C test_ioccc clobber
 	${MAKE} -C soup clobber
+	${MAKE} -C jparse clobber
 
 distclean nuke: clobber
 
--- a/soup/Makefile
+++ b/soup/Makefile
@@ -46,10 +46,10 @@ soup:
 	fi
 
 clean:
-	${RM} -f *.o
+	${RM} *.o
 
 clobber: clean
-	${RM} -f .soup
+	${RM} .soup
 
 install: all
 	@${ECHO} "nothing to $@. Try eating the soup instead."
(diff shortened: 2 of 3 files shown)


lcn2:
> Fix 'No rule to make target `jparse/Makefile', needed by `jparse'' in Makefile.
> 
> Fix strange issue where running make clobber would pass to rm -f -f instead of just -f. I'm not even sure why this happened - I don't have the time or energy to try figuring it out. It seems odd because there's no reference to -f -f in any file. It did not occur with the dbg subdirectory but it does with soup and test_ioccc/. The fix was removing the single -f from the rm invocation in those Makefiles. It's very funny because even after removing the -f from the invocation in e.g. soup/Makefile if one does `cd soup; make clobber' you'll see that it does in fact run rm -f.

We are not sure how the extra `-f` was added.  However we are not certain that this fix addresses the problem properly.

Moreover the use of `rm .. *.o` is sort of a "sledgehammer" approach to the problem that should be avoided in possible.

We are going to reluctantly accept this pull request and then try to fix the problem directly.

lcn2:
Well this pull request was a bad idea it seems:

```sh
$ make clobber
...
make -C test_ioccc clobber
rm -f *.o
rm -f utf8_test jnum_chk dyn_test
make -C soup clobber
rm -f *.o
rm -f .soup
make -C jparse clobber
make[1]: *** No rule to make target `clobber'.  Stop.
make: *** [clobber] Error 2
zsh: exit 2     make clobber
```

lcn2:
We reverted this pull request.

lcn2:
There is no `jparse/Makefile` checked in, BTW.

xexyl:
> There is no `jparse/Makefile` checked in, BTW.

That could have been easily fixed though. I have a copy of it at least. Strange that I didn't add it. Must have been because yesterday was kind of hectic.

xexyl:
> Well this pull request was a bad idea it seems:
> 

The problem was I forgot, it seems, to add the Makefile. For those rules do exist on my end. Still though the idea you had about a static library is a great idea.

> ```shell
> $ make clobber
> ...
> make -C test_ioccc clobber
> rm -f *.o
> rm -f utf8_test jnum_chk dyn_test
> make -C soup clobber
> rm -f *.o
> rm -f .soup
> make -C jparse clobber
> make[1]: *** No rule to make target `clobber'.  Stop.
> make: *** [clobber] Error 2
> zsh: exit 2     make clobber
> ```

Again that's just because I forgot to add the Makefile. Would have been easy to fix that part.

xexyl:
> > Fix 'No rule to make target `jparse/Makefile', needed by `jparse'' in Makefile.
> > Fix strange issue where running make clobber would pass to rm -f -f instead of just -f. I'm not even sure why this happened - I don't have the time or energy to try figuring it out. It seems odd because there's no reference to -f -f in any file. It did not occur with the dbg subdirectory but it does with soup and test_ioccc/. The fix was removing the single -f from the rm invocation in those Makefiles. It's very funny because even after removing the -f from the invocation in e.g. soup/Makefile if one does `cd soup; make clobber' you'll see that it does in fact run rm -f.
> 
> We are not sure how the extra `-f` was added. However we are not certain that this fix addresses the problem properly.
> 
Actually I agree and wanted to look at it more. I think it was that a Makefile (or maybe two) need a clean up. Or I should say needed a clean up.

> Moreover the use of `rm .. *.o` is sort of a "sledgehammer" approach to the problem that should be avoided in possible.
> 
> We are going to reluctantly accept this pull request and then try to fix the problem directly.

Well I did ask you for that reason - I thought that that was what you wanted though. Maybe you thought it over again? Well anyway I'll take this time to work on other things. Some here but other things as well.

Have a great time in Canada - and a safe time! And a happy holiday too! Stay safe.

PR #421, opened 2022-11-21
Revert "Makefiles clean up and fixes"

lcn2:
Reverts ioccc-src/mkiocccentry#420

Due to `make clobber` failure, plus other issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -417,7 +417,7 @@ jparse.tab.o: jparse.tab.c Makefile
 jparse_main.o: jparse_main.c Makefile
 	${CC} ${CFLAGS} jparse_main.c -c
 
-jparse: jparse/jparse.h jparse/jparse.l jparse/jparse.y
+jparse: jparse/jparse.h jparse/jparse.l jparse/jparse.y jparse/Makefile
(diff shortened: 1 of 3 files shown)


xexyl:
What other issues please ? But I see maybe you already answered these in the other comments. Hopefully it's not something that could have easily been fixed. Still I appreciate the heads up so I could copy everything to a tarball so I know how I did it - at least some of it will be useful.

xexyl:
What really baffles me is that the `cd .. ; make ...` did not work for you. I specifically tested those and they were fine. The rules were in fact there.

Oh well. As you say once a library system is in place it'll be better. Also I again really appreciate that you pointed out that it is indeed not my fault!

Hope you're sleeping well. I'm going to try resting soon myself.

xexyl:
But looking at `Makefile` I still see that `soup/Makefile` and `dbg/Makefile` are used. Could it be that the fact I forgot to add the Makefile for jparse be the real issue here ?

Well anyway I'll hold off as you suggested. Though if I have sudden motivation to look at the static library issue I might at least say something in one of the threads.

PR #424, opened 2022-11-21
Add depend to soup/Makefile

xexyl:
Question for you:

Should the `dbg/Makefile` also have a depend rule? If so what to do about the soup/fmt_depend.sh script ?

And..really resting now. Maybe I can address other things later on today but if so it's totally unrelated to any of this.


xexyl:
> Question for you:
> 
> Should the `dbg/Makefile` also have a depend rule? If so what to do about the soup/fmt_depend.sh script ?

The obvious solution is to make a make depend script. If we have to rely on the formatting of it that's the solution anyway.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help.

PR #427, opened 2022-11-22
Move hexval in util to json_util

xexyl:
Since hexval is strictly for JSON it belongs in json_util. A point of interest is it's not actually used in json_util but rather json_parse so why can't we have it in json_parse? We could but it would require renaming either it or the variable in jencchk() in json_parse.c as if the array was in json_parse.c it would be a name clash. Renaming the unsigned int hexval probably should be done anyway because it shares the name of an array but for now it's not done.


xexyl:
I was hoping to do some other changes but I don't think I will be doing those today. Going to rest again and when I'm up I don't think I'll work on this for the day. Really tired and feeling off.

lcn2:
Off to Canada 🇨🇦 .. thanks @xexyl

xexyl:
> Off to Canada 🇨🇦 .. thanks @xexyl

Have a safe time and a good holiday!

PR #433, opened 2022-12-01
Lower MAX_MASTODON_LEN

xexyl:
Although technically correct at what it was, it was suggested to lower it. I actually thought of this as well so I went ahead and did it. A note on it though. The 255 to 64 does not really need an explanation as 64 is also the value of MAX_URL_LEN so it is a reasonable limit. If this is ever a problem for either it could of course be increased.

But the 20 instead of 19 is because unlike twitter, mastodon handles have two '@'s so the limit should allow for the same number of characters in twitter handles, as far as user names go, which was 18.


lcn2:
Processing this useful merge, but it might be a few days before we pull it down to our local repo.

xexyl:
> Processing this useful merge, but it might be a few days before we pull it down to our local repo.

That must be so frustrating :( Best of luck with it!

PR #434, opened 2022-12-04
Bug fixes and prepare for jparse/ subdirectory

xexyl:
If a binary fails to compile make depend could fail. This might seem like a good idea but the problem is sometimes this leads to being unable to run make depend which is actually needed to successfully compile the binary in the first place when actually running make depend would solve the problem. In other words the make depend should not depend on having to compile everything first because this can prevent correct and successful compilation and in fact prevent one from doing anything at all until the Makefile dependencies are manually updated which defeats the purpose of having make depend.

The files changed in this commit are the main Makefile and soup/Makefile; the dyn_array/Makefile and dbg/Makefile don't yet have a depend rule.

To make it clear it means the make depend should not have 'all' as a dependency. I am pointing this out for when dbg/Makefile (and the external dbg repo Makefile) and the dyn_array/Makefile (and the future dyn_array repo Makefile).


xexyl:
With commits https://github.com/ioccc-src/mkiocccentry/pull/434/commits/75791dd5161fc93ef4457eb536e6c92a6ce5754d we will have the first step of jparse subdirectory done. It also fixes some bugs in bug_report.sh that were introduced when moving the previous `jparse/` back to the root directory.

I can now focus on making some of those other files that you mentioned in the comment in the other merged pull request.

Going to try resting again. Not sure if I'll do more today. If I do I hope to start working on those other things that will allow for correcting the man pages - which will also make it so fewer things have to be done when the jparse is its own repo.

xexyl:
Converting to a draft as I'm aware of an issue .. going to fix it in a bit - have something I have to do first.

xexyl:
Issues fixed. Doing something else now and I doubt I'll do anything here until at least tomorrow.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

I'm happy to help!

I'm here a bit longer and then will be leaving for the day. I do hope to work on _something_ tomorrow but I don't know what yet - or if I will manage. I do have a zoom meeting n the morning but nothing else planned.

PR #435, opened 2022-12-05
Various updates and fixes

xexyl:
Although they're not strictly correct just yet they are better th an they were and I need to get the modified files out of the way but don't want to stash them. The reason they're not completely correct is they should not actually refer to the IOCCC at all but currently do. The reason they should not refer to it is when there's a jparse repo (after the IOCCCMOCK I will do this) they will also be there as they're not IOCCC specific. However the way they're worded for now is still an improvement.

I also want to clarify something I said in commit
0185b5f378bb67ed558c39a4e18bf8742ffa2c6e: make chkentry does of course show 'is up to date' when it's compiled and nothing has been updated that necessitates a recompilation. But make all, make dbg and make dyn_array do not and so neither should make jparse which is (as of that commit) a subdirectory (just like dbg and dyn_array).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,7 @@ have_timegm
 iocccsize
 jnum_chk
 jnum_gen
-jparse
+jparse/jparse
 jparse.c
 jparse.lex.h
 jparse.tab.c
--- a/Makefile
+++ b/Makefile
@@ -460,12 +460,12 @@ soup: soup/chk.auth.head.c soup/chk.auth.ptch.h soup/chk.info.head.c soup/chk.in
 	soup/chk_sem_info.c  soup/chk_validate.h
 	@${MAKE} -C soup CFLAGS="${CFLAGS}"
 
-chkentry.o: chkentry.c chkentry.h jparse.tab.h Makefile
+chkentry.o: chkentry.c chkentry.h jparse.tab.h oebxergfB.h Makefile
 	${CC} ${CFLAGS} -Isoup chkentry.c -c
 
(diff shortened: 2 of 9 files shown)


xexyl:
Hope you're okay. I'm sure you are but just wanted to say so anyway.

I am thinking there's something I was going to say but I have no idea what or if I'm making that up. If I do it's probably OT so unimportant.

xexyl:
Ah .. I remember what one of those things is. It's something I just shared on mastodon. Did you know that Ritchie was given a patent for suid bit ? I gave a link over there of the actual patent.

lcn2:
> Ah .. I remember what one of those things is. It's something I just shared on mastodon. Did you know that Ritchie was given a patent for suid bit ? I gave a link over there of the actual patent.

We are very aware of the so-called _set UID patent_ bit.  Landon was one of the main people in the POSUIX P1003.1 standards body who forced AT&T to deed that patent into the public domain.  The outcome of that action was the change the way IEEE standards bodies deal with patents.  Some of those changes were for the better, some for the worse.

xexyl:
> > Ah .. I remember what one of those things is. It's something I just shared on mastodon. Did you know that Ritchie was given a patent for suid bit ? I gave a link over there of the actual patent.
> 
> We are very aware of the so-called _set UID patent_ bit. Landon was one of the main people in the POSUIX P1003.1 standards body who forced AT&T to deed that patent into the public domain. The outcome of that action was the change the way IEEE standards bodies deal with patents. Some of those changes were for the better, some for the worse.

That does not surprise me - any part of it!

And part of it I of course knew - you being part of the POSIX body.

xexyl:
> Thanks @xexyl .. will look at the merge later when networks are better.

Of course. Nothing major really.

PR #440, opened 2022-12-21
Fix EXTERN_PROG in jparse/Makefile

xexyl:
It was missing jnum_gen which means that one would have to do make -C jparse before jnum_gen was built. This prevented bug_report.sh from working as it could not execute jparse/jnum_gen since it was not compiled.

### DIFF
--- a/jparse/Makefile
+++ b/jparse/Makefile
@@ -259,7 +259,7 @@ EXTERN_H= jparse.h
 EXTERN_O=
 EXTERN_MAN= ${ALL_MAN_TARGETS}
 EXTERN_LIBA= jparse.a
-EXTERN_PROG= jparse jsemtblgen jsemcgen.sh jstrdecode jstrencode
+EXTERN_PROG= jparse jsemtblgen jsemcgen.sh jnum_gen jstrdecode jstrencode
 
 # NOTE: ${EXTERN_CLOBBER} used outside of this directory and removed by make clobber
 #


xexyl:
Doing other things now and I mean to not do anything else here today. That of course could change but quite unlikely.

lcn2:
We are having a winter Solstice party.  It is -16.6C with a wind chill of -25C.  However the fire is going well and we are comfy and warm.

Thanks for the fix.  I will be limited in my access thru this week.

xexyl:
> We are having a winter Solstice party. It is -16.6C with a wind chill of -25C. However the fire is going well and we are comfy and warm.
> 
> Thanks for the fix. I will be limited in my access thru this week.

That’s really cold! 

Stay safe and have a great holiday!

lcn2:
> > We are having a winter Solstice party. It is -16.6C with a wind chill of -25C. However the fire is going well and we are comfy and warm.
> 
> > 
> 
> > Thanks for the fix. I will be limited in my access thru this week.
> 
> 
> 
> That’s really cold! 
> 
> 
> 
> Stay safe and have a great holiday!

Happy Winter Solstice (we presume you are currently in the northern hemisphere :))!

xexyl:
> > > We are having a winter Solstice party. It is -16.6C with a wind chill of -25C. However the fire is going well and we are comfy and warm.
> > 
> > 
> > > 
> > 
> > 
> > > Thanks for the fix. I will be limited in my access thru this week.
> > 
> > 
> > That’s really cold!
> > Stay safe and have a great holiday!
> 
> Happy Winter Solstice (we presume you are currently in the northern hemisphere :))!

I am indeed! Though it doesn’t get nearly that cold here. We have had mid to high 30s in the winter! But fortunately it is not so hot right now. I have always felt better when it’s cooler. 

Stay warm but have a great time!

PR #441, opened 2022-12-22
Remove external sanity checks in txzchk

xexyl:
Remove inclusion of jparse/jparse.h in txzchk.h. This is an artefact from a long time ago where the tool ran all the sanity checks including the old (manual) parsing of .info.json and .auth.json (then as .author.json) files. Some of the functions it called no longer exist and the json sanity checks were not being done by txzchk since ages.

I also removed the call to ioccc_sanity_chks() as it's not strictly needed. The rationale was similar for the json checks - that if something is wrong with the core then we should not proceed. But mkiocccentry does call the function and also calls txzchk. Since no other tool uses the function I don't think txzchk needs to either. It probably never did.

Make depend due to no longer including jparse/jparse.h in txzchk.h.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -963,10 +963,10 @@ iocccsize.o: iocccsize.c iocccsize_err.h iocccsize.h
 fnamchk.o: fnamchk.c fnamchk.h dbg/dbg.h jparse/jparse.h jparse/../dbg/dbg.h jparse/util.h \
 	jparse/../dyn_array/dyn_array.h jparse/../dyn_array/../dbg/dbg.h jparse/json_parse.h \
 	jparse/json_util.h jparse/json_sem.h jparse/jparse.tab.h limit_ioccc.h version.h utf8_posix_map.h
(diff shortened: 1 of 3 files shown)


xexyl:
I might manage to start work on the man page suggestions I mentioned yesterday later on today but we'll see. Given the time it's less likely but I'll see how I feel in a bit.

xexyl:
> I might manage to start work on the man page suggestions I mentioned yesterday later on today but we'll see. Given the time it's less likely but I'll see how I feel in a bit.

It's proving to be a rather difficult day so I will be looking at this another day. If I find the time I might discuss some of the issues in the respective thread or threads but I doubt I'll actually do the things today. Hopefully I can tomorrow if not though the man page changes will take a while I'm thinking.

xexyl:
Ahhh I know why the header file was included now.

It's because you moved the util.h to `jparse/`. It still works because `txzchk.h` also includes `entry_util.h` which also includes `jparse/jparse.h`. I'm not sure if it should do it this way or if it should also include `jparse/jparse.h`.

However it does make the log inaccurate so I'm going to close this and recreate the pull request with the other change. The jparse/jparse.h can stay included.

PR #444, opened 2022-12-24
Add man pages for different types of (dbg) messages

xexyl:
This means that the following main man pages were added:

    err.3
    warn.3
    werr.3
    warn_or_err.3
    printf_usage.3
    msg.3

and the dbg.3 has been modified to only refer to (in description) the dbg() specific functions.

Also because a lot of man pages have been added aside from these I moved the man pages to man/man3 (like in an installed system).

A note is that there is room for improvement here. This is a work in progress but enough work has been done to commit.

An important point is that although the Makefile now installs those man pages it does not install the man pages for the additional man pages with the .so man3/foo.3 in it. This is because there are a lot of them and the way to go about it might be up to discussion first. There even are some other ideas as opposed to using .so.

Another method might be to use symlinks so that one can do

     man ./man/man3/werr.3

for example and get the man page output whereas this will not happen if it's not installed. Another option is to have make install use ln to make symlinks. The way it currently stands is that one can only do:

    man man/man3/err.3
    man man/man3/warn.3
    man man/man3/werr.3
    man man/man3/warn_or_err.3
    man man/man3/printf_usage.3
    man man/man3/msg.3

and expect the right output as each one will have the group of functions. But whether we want to make it symlinks instead - that might be an idea. Depending on what is decided will change how make install works.

The groups I selected could be changed as well but this is a good start.


xexyl:
This is all I'll be doing today but I feel like it's a good progress. I'll make a comment in the other thread raising some points I brought up here and then will be doing other things for the day. You can expect that tomorrow I won't be doing anything so probably the earliest I'll do something here next is Monday.

lcn2:
Thanks for this trial pull request, @xexyl. Let us pause for a bit and see if we like this approach before considering this for other parts of thee we _mkiocccentry repo_.

xexyl:
> Thanks for this trial pull request, @xexyl. Let us pause for a bit and see if we like this approach before considering this for other parts of thee we _mkiocccentry repo_.

I completely agree with that.

I'm just watching a film. Will possibly be able to reply to anything else later on today but most likely not. I'm sure I'll say something tomorrow though if for no other reason but to wish you a happy Christmas.

xexyl:
The more I think on it the more I think that it might be better to do symlinks. The reason is that one can then do man locally (as in the local directory): without it being installed.

That's the way curses (or at least ncurses) does it. What do you think?

lcn2:
> The more I think on it the more I think that it might be better to do symlinks. The reason is that one can then do man locally (as in the local directory): without it being installed.
> 
> That's the way curses (or at least ncurses) does it. What do you think?

The way it it is now under `dbg/`, commands such as (where the man pages are **NOT installed**):

```sh
$ man dbg/man/man3/werrp.3
```

**DOES NOT** work.

Where as, if one did this (where the man pages are **NOT installed**):

```sh
$ ln -s -f werr.3 werrp.3
$ man dbg/man/man3/werrp.3
```

then it **DOES** work.

BTW: We have **NOT** synced the `mkiocccentry/dbg/` tree with the _dbg repo_.  We suggest that once this map page stuff and Makefile issue #422 is resolved, we should re-sync back to the  _dbg repo_: until then, it is OK for these to be divergent.

xexyl:
> > The more I think on it the more I think that it might be better to do symlinks. The reason is that one can then do man locally (as in the local directory): without it being installed.
> > That's the way curses (or at least ncurses) does it. What do you think?
> 
> The way it it is now under `mkiocccentry/dbg/`, commands such as (where the man pages are **NOT installed**):
> 
> ```shell
> $ man dbg/man/man3/werrp.3
> ```
> 
> **DOES NOT** work.
> 
> Where as, if one did this (where the man pages are **NOT installed**):
> 
> ```shell
> $ cd dbg/man/man3
> $ ln -s -f werr.3 werrp.3
> $ cd ../../..
> $ man dbg/man/man3/werrp.3
> ```
> 
> then it **DOES** work. So yes, the symlink idea you suggest in [comment 1364697431](https://github.com/ioccc-src/mkiocccentry/pull/444#issuecomment-1364697431) is a better approach.
> 
> ### UPDATE 0a
> Please consider forming `mkiocccentry/man/man1/`, `mkiocccentry/man/man3/` and `mkiocccentry/man/man8/` directories.
> 
> Please, @xexyl, update the `mkiocccentry/dbg/` tree as well as the `mkiocccentry/man/` trees (and their respective Makefiles) accordingly.
> 
> BTW: We have **NOT** synced the `mkiocccentry/dbg/` tree with the _dbg repo_. We suggest that once this map page stuff and Makefile issue #422 is resolved, we should re-sync back to the _dbg repo_: until then, it is OK for these to be divergent.

Yes. I pointed out this problem initially. I think that a symlink would solve it but I will have to play with it. 

I will try looking at this sometime in the week. 

Way past bedtime for me. Good night and sleep well!

xexyl:
Ah. I see you also tried that. Okay. Well hopefully I can start on it tomorrow. Good night for now.

xexyl:
> > > The more I think on it the more I think that it might be better to do symlinks. The reason is that one can then do man locally (as in the local directory): without it being installed.
> > > That's the way curses (or at least ncurses) does it. What do you think?
> > 
> > 
> > The way it it is now under `mkiocccentry/dbg/`, commands such as (where the man pages are **NOT installed**):
> > ```shell
> > $ man dbg/man/man3/werrp.3
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **DOES NOT** work.
> > Where as, if one did this (where the man pages are **NOT installed**):
> > ```shell
> > $ cd dbg/man/man3
> > $ ln -s -f werr.3 werrp.3
> > $ cd ../../..
> > $ man dbg/man/man3/werrp.3
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then it **DOES** work. So yes, the symlink idea you suggest in [comment 1364697431](https://github.com/ioccc-src/mkiocccentry/pull/444#issuecomment-1364697431) is a better approach.
> > ### UPDATE 0a
> > Please consider forming `mkiocccentry/man/man1/`, `mkiocccentry/man/man3/` and `mkiocccentry/man/man8/` directories.
> > Please, @xexyl, update the `mkiocccentry/dbg/` tree as well as the `mkiocccentry/man/` trees (and their respective Makefiles) accordingly.
> > BTW: We have **NOT** synced the `mkiocccentry/dbg/` tree with the _dbg repo_. We suggest that once this map page stuff and Makefile issue #422 is resolved, we should re-sync back to the _dbg repo_: until then, it is OK for these to be divergent.
> 
> Yes. I pointed out this problem initially. I think that a symlink would solve it but I will have to play with it.
> 
> I will try looking at this sometime in the week.
> 
> Way past bedtime for me. Good night and sleep well!

At the phone now. I will be getting up shortly I hope. I am not sure if I will get to this today but at least the symlink part should be relatively easy. 

The other part - which I started with dbg (the subdirectories for each man page section) - might also be easy but it depends on if each of those need a Makefile. I also have to look at the Makefile that does exist to see if there is anything that might complicate matters. 

One thing that I probably won’t tackle yet is the installation of all the new symlink man pages. But we will see. 

I did manage to sleep in today but I am still very very tired. A lot of stimulation yesterday which is exhausting for me. It was totally worth it but it probably will take some days to fully recover. 

I hope you had a great Christmas yesterday!

If I don’t get to this today I hopefully can manage it tomorrow. And your idea about synching the Makefiles is good I think. 

If I had to take a guess I would say that I very possibly might not get to this today. But we shall see.

xexyl:
> > The more I think on it the more I think that it might be better to do symlinks. The reason is that one can then do man locally (as in the local directory): without it being installed.
> > That's the way curses (or at least ncurses) does it. What do you think?
> 
> The way it it is now under `mkiocccentry/dbg/`, commands such as (where the man pages are **NOT installed**):
> 
> ```shell
> $ man dbg/man/man3/werrp.3
> ```
> 
> **DOES NOT** work.
> 
> Where as, if one did this (where the man pages are **NOT installed**):
> 
> ```shell
> $ cd dbg/man/man3
> $ ln -s -f werr.3 werrp.3
> $ cd ../../..
> $ man dbg/man/man3/werrp.3
> ```
> 
> then it **DOES** work. So yes, the symlink idea you suggest in [comment 1364697431](https://github.com/ioccc-src/mkiocccentry/pull/444#issuecomment-1364697431) is a better approach.
> 

Please see https://github.com/lcn2/dbg/pull/17/commits/7e861b75b6e71cc88acdac12278ebff256fca581 and https://github.com/ioccc-src/mkiocccentry/pull/445/commits/a7ef52ec81f3b880252eb412400a62060236de45.


> ### UPDATE 0a
> Please consider forming `mkiocccentry/man/man1/`, `mkiocccentry/man/man3/` and `mkiocccentry/man/man8/` directories.
> 

This might not happen today but I will definitely do it. Then again it might be today. I did the above as you can see which I deemed more important.

> Please, @xexyl, update the `mkiocccentry/dbg/` tree as well as the `mkiocccentry/man/` trees (and their respective Makefiles) accordingly.
> 

I'll address this when I go for the separating the man pages by section like I started with dbg man pages. At this point the installation of the linking man pages (as I called them .. not sure if there's a better or correct term) could be done as well.

> BTW: We have **NOT** synced the `mkiocccentry/dbg/` tree with the _dbg repo_. We suggest that once this map page stuff and Makefile issue #422 is resolved, we should re-sync back to the _dbg repo_: until then, it is OK for these to be divergent.

Good idea.

xexyl:
> > ### UPDATE 0a
> > Please consider forming `mkiocccentry/man/man1/`, `mkiocccentry/man/man3/` and `mkiocccentry/man/man8/` directories.
> 
> This might not happen today but I will definitely do it. Then again it might be today. I did the above as you can see which I deemed more important.
> 
> > Please, @xexyl, update the `mkiocccentry/dbg/` tree as well as the `mkiocccentry/man/` trees (and their respective Makefiles) accordingly.
> 
> I'll address this when I go for the separating the man pages by section like I started with dbg man pages. At this point the installation of the linking man pages (as I called them .. not sure if there's a better or correct term) could be done as well.
> 

Please see 1a0ca6082f40ce776c27e4dee01b19357b2235e2.

This still does not install the man pages that link to others. This I guess is not hard but I'm not sure if I have the patience to do that right now as it's redundant and I am tired. I might look at it but I might delay it until another time.

PR #445, opened 2022-12-26
Make linking man pages symlinks instead of .so

xexyl:
In other words instead of using .so man3/foo.3 it is just a symlink. This way running man on the file when it's not installed will work.


lcn2:
Appreciated @xexyl

xexyl:
> Appreciated @xexyl

Happy to help.

PR #447, opened 2022-12-29
Fix the make all repeatedly compiling everything

xexyl:
There might still be some issues but at least now running make all more
than once will try recompiling everything. This is on a different branch
because of some problems I had some of which I'm not sure entirely of
the source.

Specifically one to do with permissions. This was very possibly because
of earlier on trying to do make && sudo make install but then the make
install as root actually compiling the object files again - but it might
not be because make clobber all sometimes (not always) resulted in the
problem again.

Sometimes I got an error trying to link dbg but that was because, it
appears, that a Makefile (not in dbg/) did make dbg instead of dbg.o or
dbg.a. I changed it to dbg.a but in retrospect it might need to be
dbg.o: I'm not sure but noting this in case it is a problem. In
particular it's in test_ioccc/Makefile:

     ../dbg/dbg.a: ../dbg/Makefile
    -       @${MAKE} -C ../dbg dbg extern_liba
    +       @${MAKE} -C ../dbg dbg.a extern_liba

Given that it's extern_liba it probably is okay for dbg.a but just in
case I'm noting it (I didn't see the extern_liba until now as I was
focused on the actual problem).

Some objects were added to .PHONY rules and some were removed. Some
rules ('all') had a no-op @: added to them to silence make when nothing
was to be done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -299,6 +299,7 @@ TARGETS= mkiocccentry iocccsize txzchk chkentry ${SH_TARGETS}
 ######################################
 
 all: fast_hostchk just_all
+	@:
 
 # The original make all that bypasses running hostchk.sh
 #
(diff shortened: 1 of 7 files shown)


lcn2:
Very good catches, @xexyl Thank you.

Our problem was too long hours and being too deep into the Makefiles to see these problems.

xexyl:
> Very good catches, @xexyl Thank you.
> 
> Our problem was too long hours and being too deep into the Makefiles to see these problems.

Thank you for the wonderful compliment as well! I'm happy I could help.

xexyl:
And being too deep involved in something can indeed be a problem sometimes. Makes it harder to see a problem. Or as Elrond says of Saruman:

> 'This is grievous news concerning Saruman,' he said; 'for we trusted him and
he is deep in all our counsels. It is perilous to study too deeply the arts of
the Enemy, for good or for ill. But such falls and betrayals, alas, have
happened before.

Not quite the same but it's an interesting thought - though I'm not sure I exactly agree. Another quote of interest is:

> ' "In which case it is no longer white," said I. "And he that
breaks a thing to find out what it is has left the path of wisdom.

Though I have heard a story where Tolkien took a clock apart to try and figure out how it works and couldn't put it back together.

I'm not sure either are valid but why not bring some Tolkien joy to the situation?

I'll be leaving now for the day. I will be up a bit later tonight but hopefully it won't be too much of a problem tomorrow.

Good night!

Issue #449, opened 2022-12-30
Enhancement: move the test_JSON tree


lcn2:
## Background

> ...shouldn't `vermod.sh` be put under `jparse` and not `soup` ?

The propose of `vermod.sh` is _IOCCC specific_: to update version numbers that matter to the _mkiocccentry repo_.  

While `jparse/test_jparse/test_JSON/` contains what seems to be _IOCCC data_, data that `vermod.sh` is designed to modify: what the _JSON parser_ needs are JSON test files.  It just happens that the _IOCCC inspired_  `jparse/test_jparse/test_JSON/` happened to "_foot the bill_" (as that expression goes).

## move the test_JSON tree

We suggest that `jparse/test_jparse/test_JSON/` should be moved to `test_ioccc/test_JSON/` and that `vermod.sh` be modified to impact that directory tree.

Now, the _JSON parser_ needs test JSON files of both valid and invalid JSON variety.  So while the **CONTENT** under `jparse/test_jparse/test_JSON/` should be moved to `test_ioccc/test_JSON/`, the _JSON parser_ **STILL NEEDS** a `parse/test_jparse/test_JSON/` test tree: just something that is **NOT** _IOCCC / mkiocccentry_ related.  We suggest that the "something" that the _JSON parser_ should do is to make the new `jparse/test_jparse/test_JSON/` based on the theme that `jparse/test_jparse/sample_JSON/party.json` uses.

For the above to be accomplished well, the respective tools that test the JSON parser need a clear mechanism to identify the directory under which the _test_JSON_ stuff is found.  Moreover, the directory structure and filename structure under the current `jparse/test_jparse/test_JSON/` directory needs to be maintained.  

The current `jparse/test_jparse/jparse_test.sh` does this with the `-d json_tree` argument that defaults to the "`$JSON_TREE`" (`./jparse/test_jparse/test_JSON`) value.   So there may not be much that needs to change.  Perhaps what is needed (in addition to the above melted tree move and new _test_JSON_ tree contents based on _party.json_ theme) is that the _test_ Makefile rule of `json/test_jparse/Makefile` needs to point at the _test_JSON_ tree contents based on the _party.json_ theme WHEREAS the  _test_ Makefile rule of `test_ioccc/Makefile` would invoke `jparse/test_jparse/jparse_test.sh` executed with `-d test_ioccc/test_JSON` (containing the current JSON contents that is under `jparse/test_jparse/test_JSON/` now).

[...]

PR #451, opened 2022-12-31
Resolve issue #449 - move test_JSON tree

xexyl:
The jparse/test_jparse/test_JSON has been renamed to test_ioccc/test_JSON.

The soup/vermod.sh was changed so that the default tree points to the new location.

The jparse/test_jparse/sample_JSON was renamed to a new test_JSON (as jparse/test_jparse/test_JSON) and there is a good subdirectory with party.json. There are no bad files yet - this can happen another time.

The test rule in the jparse/test_jparse/Makefile refers to the above mentioned test_JSON under jparse/test_jparse.

The jparse_test.sh script has a new -s option for subdirectory. The default is general.json for IOCCC tools but perhaps it should be the other way round. This can be changed later - I only just thought of this as I'm more alert now. The man page has been updated.

Rules that referred to jparse/test_jparse/test_JSON have been updated to refer to test_ioccc/test_JSON.

Make dependencies updated via one of the scripts (perhaps make prep but not sure).

There might be other things that have changed and it's certainly possible that more has to be done here but I believe this issue is entirely resolved the -s option to the jparse_test.sh idea that I mentioned above notwithstanding.


lcn2:
Good move 😉 @xexyl

xexyl:
> Good move 😉 @xexyl

Thank you. I hope it solves the problem - again except the idea I had with changing the order of the `-s` flag.

PR #454, opened 2023-01-01
Fix make test on other systems (than mine)

xexyl:
The reason make test did not fail for me is that I had made an empty directory bad/ under jparse/test_jparse/test_JSON but I forgot to add a file which means that the directory could not be found. Thus I have added a jparse/test_jparse/test_JSON/bad/party.json (to be in the theme of the good/party.json).

I also came to realise that diacritics are allowed in json so I changed the diacritics being escaped to diacritics. Perhaps though we should have some of the \u codes as well.

An aside for those who look at the party.json file and who are also interested in The Lord of the Rings. It is NOT an error to say that an alias of Gollum is Dígol. This is because in the earlier draft (see History of Middle-earth [HoMe] VI - The Return of the Shadow) there was no friend and it was ONLY Dígol. There was even at one point (immediately struck out) that he was a kind of goblin kind! Also, if we're going to go into history: in the first edition of The Hobbit Gollum was willing to give up the Ring and was seriously distressed when he could not find it. Bilbo let him off if he showed him the way out. This is because at that point it was only meant to be a literary device to make one invisible. When a sequel was called for the two obvious things that could be a plot point were the Ring and the Necromancer (which was originally just so that Gandalf would have to leave the Dwarves - NOT dwarfs - and Bilbo for a bit). But I've digressed into another love of mine: the point is Dígol is a valid alias for Gollum earlier on! :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1053,7 +1053,7 @@ install: all dbg/Makefile man/Makefile
 # make depend #
 ###############
 
-depend: ${ALL_CSRC} all
+depend: ${ALL_CSRC}
 	@echo
 	@echo "${OUR_NAME}: make $@ starting"
 	@echo
--- a/dbg/Makefile
+++ b/dbg/Makefile
@@ -544,7 +544,7 @@ install: all
 # make depend #
 ###############
 
-depend: ${ALL_CSRC} all
+depend: ${ALL_CSRC}
 	@echo
 	@echo "${OUR_NAME}: make $@ starting"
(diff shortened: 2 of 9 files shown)


xexyl:
I had to throw in some fun in the log to also correct anyone who might be unaware of earlier facts. :)

I have another commit coming. Then I might look at the man page installation. After that I can open another issue about those three man pages. I don't know that I'll get to the major issue today. We shall see.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Just saw this .. happy to help.

PR #462, opened 2023-01-04
Various updates, fixes and enhancements

xexyl:


xexyl:
> As for commit [5b62287](https://github.com/ioccc-src/mkiocccentry/commit/5b62287cd05be4f732736f4f36751845acc21191):
> 
> ```
> Enable make verbosity in bug_report.sh
> 
> Via the (settable) MAKE_FLAGS variable (by default) the make flags are:
> 
>     Q= V=@ MAKE_CD_Q= -w
> 
> If you wish to disable these you can do:
> 
>     ./bug_report.sh -M ''
> 
> Alternatively, if you want to add an option or only disable some of the
> actions you can do so like:
> 
>     ./bug_report.sh -M 'Q='
> 
> for example.
> 
> We do NOT recommend using the option when actually reporting a bug but
> it is there because there are cases where it's useful to have.
> 
> Updated the man page for this option as well as clarifying the -x option
> in the EXIT STATUS section.
> ```
> 
> I think this is the way to go about it as sometimes we might not need the extra verbosity. I did not add to the man page that it should not be used (`-M`) when reporting bug reports. I'll do that now.

Done with commit https://github.com/ioccc-src/mkiocccentry/pull/462/commits/77394fab2616662f0d086a4d34d0ade58ec8ae77.

xexyl:
The `dbg` related problem with `bug_report.sh` is a fun one. I've discovered the source of the problem. Now I have to find out a way to fix it. I'll have to leave soon so it might not be until after I get back but I know now the problem at least! 

I think you'll find some kind of amusement out of it esp since `make clobber all` will show that it's not compiled but then running `make picky` and suddenly it is!

xexyl:
> The `dbg` related problem with `bug_report.sh` is a fun one. I've discovered the source of the problem. Now I have to find out a way to fix it. I'll have to leave soon so it might not be until after I get back but I know now the problem at least!
> 
> I think you'll find some kind of amusement out of it esp since `make clobber all` will show that it's not compiled but then running `make picky` and suddenly it is!

Actually it was also jparse/test_jparse/jnum_chk but make clobber was broken: it didn't work in jparse/test_jparse. That's now been fixed but the problem is that the files are not being compiled. In the case of the dbg_test it's because dbg_test.c does not exist until it's copied via the Makefile from dbg.c.

I might end up reporting this and worrying about something else but I'll look at it a bit longer first.

xexyl:
Fixed in 4c489e5620d3c0beed17137de852181eac593763:

    Fix funny make all picky problem
    
    A problem existed where make all would not compile everything but make
    picky would then go ahead and do the task. This caused a problem with
    make prep as bug_report.sh -t would not run make all (or any other make
    action). But masking the problem of why two binaries did not exist at
    the time of invocation but then existed after make prep finished was
    make picky actually compiling them!
    
    The problem was a couple missing rules in the main Makefile. One was
    referred to (all_dbg) but there was no actual rule. The other rule was
    not referred to at all and that was the new rule jparse_test_all.
    
    Now running make clobber all will compile everything without the need
    for make picky.
    
    Thus the bug_report.sh -t in test_ioccc/prep.sh will work fine without
    any problems.
    
    I also fixed make clobber wrt jparse/test_jparse: it did not run make
    clobber under that directory.

lcn2:
Thanks 🙏 @xexyl that was some fun detective work.

xexyl:
> Thanks 🙏 @xexyl that was some fun detective work.

Thank you for the wonderful compliment too! I also thought it was fun and funny as well. If I'm honest I still do not know why `make picky` copied .. ah I know why: because it depends on all source files. But still not sure why it compiled it. Still at least the problem is resolved now.

Hope you're having a nice sleep! I have a cat between my legs and it's pouring rain which is both rare and something I love (always have loved the rain and listening to it and reading is excellent though I won't be reading now) though I do hope it's not raining when I'm out later this morning.

lcn2:
The `make release` and `make prep` are both excellent and impressive!

We tested an adding a deliberate error into the process and it worked well.  (We removed the error again 🙂)

xexyl:
> The `make release` and `make prep` are both excellent and impressive!
> 

Thank you for the wonderful compliment!

> We tested an adding a deliberate error into the process and it worked well. (We removed the error again 🙂)

The funny thing is I did that too and I forgot to fix it before a commit - that's why I had to add commit dcf940c2f852bbdfd3c804e2e506e7784062e85e. I had meant to do a commit amend but I pushed the commit before I could do that.

PR #464, opened 2023-01-06
Fix bug in bug_report.sh wrt cat version

xexyl:
The problem was that the script first checks the return value of cat --version and if all is good it then ran not cat --version but just cat which means that it was blocking, waiting for input on stdin, thus hanging. Why this did not show itself under macOS I do not know but this should fix the problem under linux.

### DIFF
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -605,7 +605,7 @@ get_version_minimal() {
     command "${COMMAND}" --version >/dev/null 2>&1
     status=$?
     if [[ "$status" -eq 0 ]]; then
-	exec_command "${COMMAND}"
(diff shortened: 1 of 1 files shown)


lcn2:
Sorry (tm Canada 🇨🇦) you appeared have gone for the night so we made a proposed fix:

See commit 2ba9f0f939f144e0a7ea91e43577ee42254da026

We manually merged this pull request afterwards:

See commit cb67c961f3b962882339cfdbba4439fbabfed29f

xexyl:
That’s okay. It is fixed and that is what matters. Doesn’t matter if you did it or if I did it. 

If I have any additional ideas I will address them tomorrow. 

Good night!

PR #467, opened 2023-01-07
Update picky URL + improve URLs in Makefiles

xexyl:
A while back the original picky website was no longer online so I made a picky GitHub repo which is at https://github.com/xexyl/picky. The Makefiles now refer to this URL instead of the old one and I also made the text more consistent with some of the other rules that also refer to a URL (though it's not entirely finished - there are still some inconsistencies in text).

I also made _URL variables in the Makefiles for tools that we give a URL for. This is to make it easier to refer to in the Makefile without having to update the raw URL in every Makefile. Instead we simply update the variable and nothing else needs to be changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,12 @@ SLEEP= sleep
 TEE= tee
 TR= tr
 
+######################
+# URLs for utilities #
+######################
+CHECKNR_URL = https://github.com/lcn2/checknr
+PICKY_URL= https://github.com/xexyl/picky
+SEQCEXIT_URL= https://github.com/lcn2/seqcexit
 
 ####################
 # Makefile control #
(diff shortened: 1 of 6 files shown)


lcn2:
See [comment 1374552210](https://github.com/ioccc-src/mkiocccentry/issues/171#issuecomment-1374552210)

xexyl:
Don't worry about it. .. I wasn't sure at all about it. I almost didn't do it as I said.

xexyl:
And go ahead and update that other comment.

xexyl:
It's also why I put it on another branch of course.

lcn2:
> Don't worry about it. .. I wasn't sure at all about it. I almost didn't do it as I said.

Thank you for your kind response.

xexyl:
> > Don't worry about it. .. I wasn't sure at all about it. I almost didn't do it as I said.
> 
> Thank you for your kind response.

Of course. It's not a big deal. I was mixed on it but I felt why not go for it as I had already started it? If it's an idea worth considering good - if not it's not a big deal either.

I'll have to update the other url another time though. I'm doing other things now and about to change to doing something that takes no thought at all.

xexyl:
> See [comment 1374552210](https://github.com/ioccc-src/mkiocccentry/issues/171#issuecomment-1374552210)
> 
> We plan to reject this pull request, however we don't see a clean way to do this on an iPhone so we will have to wait until we are sitting down an a computer 💻.

No need. I just closed it myself before you even saw your comment.

PR #468, opened 2023-01-07
Improve bug_report.sh

xexyl:
New bug_report.sh version.

The script now collects all types commands that are needed as well as those that are optional. In other words it uses `type -a` on each command.

Additionally in the version check functions we don't report a problem if the command is not executable (don't call is_exec but instead is_exec_quiet). Instead it will go to what it always did - check for built in (though it might be in one of the functions this did not happen) - and then it will work on that. An important note here though is that for some tools like what(1) it does need a path. But doing it this way will allow for better collection of versions since most often it'll be a file anyway but in case it's not we might be able to obtain the version of the built-in (if there is one). The script could be modified if necessary to only run what(1) and the others like it if the command is a file. This probably should be done but is not for now.

There was a minor fix as well: the exit code for invalid command line was being used for invalid function calls. Thus exit code 4 has been allocated for invalid function calls (though currently this should not happen).

### DIFF
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -57,7 +57,7 @@ export TOOLS="
     ./soup/vermod.sh
     "
 
-export BUG_REPORT_VERSION="0.8 2023-01-04"
+export BUG_REPORT_VERSION="0.9 2023-01-07"
 export FAILURE_SUMMARY=
 export WARNING_SUMMARY=
 export DBG_LEVEL="0"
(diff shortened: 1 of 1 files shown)


lcn2:
Thank you for helping us make progress towards our release goal, @xexyl

xexyl:
> Thank you for helping us make progress towards our release goal, @xexyl

Happy to help and good day!

I'll be able to do more tomorrow though as you say there might not be much more to do for now. We shall see I guess.

PR #472, opened 2023-01-09
Add support for alt_url for authors

xexyl:
After discussion it was thought that a single alternative URL would be reasonable to allow for authors to have more than one URL. This commit has a lot of changed files as well as some new files.

I've updated the mkiocccentry_test.sh script to have the correct number of lines per author. It tests both empty (just hitting enter) URLs and otherwise. It does not try and check if the user has provided a main URL before allowing for an alternative URL as this seems like unnecessary complications to the code.

The test_ioccc/test_JSON/auth.json files were updated and in particular the good/ subdirectory. What I did not do here is add an alt_url field to every file under auth.json/bad. This might want to be done but there are a lot of files to do this with. I did add it to some including one that is an invalid value but whether all should have at least a null value alt_url I don't know. Obviously the good/ files have had it added.

Many files under soup/ were changed and it is this that gave me some trouble. I believe I have it sorted now as make test and make release both work. Running chkentry on various of the auth.json files also works as expected.


xexyl:
Just a couple important notes to follow about this pull request. Will make it in another comment.

xexyl:
I'm not sure if these are issues but I want to bring them up in case so that it can be fixed or adjusted prior to merging.

## ISSUE 0

Not all the auth.json/bad files have had the `alt_url` added to them. In fact most don't have it. One has it with an invalid value. That's a new file. There are a lot of files to update though and I cannot think of a good way to automate it. One at least should have it missing of course and that might be the empty one.

## ISSUE 1

This one gave me some trouble. See the diff:

```diff
diff --git a/soup/chk.auth.ptch.c b/soup/chk.auth.ptch.c
index 1cae0f6d2c5b6f5c91f5ef9277fb13f89538cc01..4313edc2eeaae2797812bd14bbd132790852c52a 100644
--- a/soup/chk.auth.ptch.c
+++ b/soup/chk.auth.ptch.c
@@ -1,45 +1,26 @@
---- ref/auth.reference.json.c      2023-01-09 08:20:23
-+++ chk_sem_auth.c 2023-01-09 08:20:03
-@@ -40,25 +40,25 @@
+--- soup/ref/auth.reference.json.c   2022-12-13 07:46:35
++++ soup/chk_sem_auth.c      2022-12-13 07:47:55
+@@ -40,9 +40,9 @@
  struct json_sem sem_auth[SEM_AUTH_LEN+1] = {
  /* depth    type        min     max   count   index  name_len validate  name */
    { 5,        JTYPE_NUMBER,   1,      5,      5,      0,      0,      NULL,   NULL },
--  { 5,    JTYPE_STRING,   1,      114,    114,    1,      0,      NULL,   NULL },
+-  { 5,      JTYPE_STRING,   1,      104,    104,    1,      0,      NULL,   NULL },
 -  { 5,        JTYPE_BOOL,     1,      10,     10,     2,      0,      NULL,   NULL },
 -  { 5,        JTYPE_NULL,     1,      1,      1,      3,      0,      NULL,   NULL },
-+  { 5,    JTYPE_STRING,   16,     114,    114,    1,      0,      NULL,   NULL },
++  { 5,      JTYPE_STRING,   16,     105,    105,    1,      0,      NULL,   NULL },
 +  { 5,        JTYPE_BOOL,     2,      10,     10,     2,      0,      NULL,   NULL },
-+  { 5,    JTYPE_NULL,     0,      30,     30,     3,      0,      NULL,   NULL },
++  { 5,      JTYPE_NULL,     0,      25,     25,     3,      0,      NULL,   NULL },
```

In particular the changing of 105 to 114 and 25 to 30.

Without this it ended up being a count error. Since there's no name for those fields I'm not sure if there is something that should be done differently. My guess is that it's the amount of times that type of json object can be seen in the file full stop. The only thing that makes me wonder if that's not it is that 114 - 109 is not 10 but 9 and with five alternative URLs (one per the five authors) I'd think that the name `alt_url` and the value would be each one of the strings so 10 not 9.

The null type was increased by 5 which makes sense - if that's the idea behind it - because five more items that can be null or a string would be - well 5.

xexyl:
Oh I forgot to say in the log that I fixed some bugs as well in mkiocccentry. Or was it chkentry? It might have been one of the support files for chkentry. I think so. There were a few cases where the wrong pointer was checked for NULL.

Yes that was it. In function `object2author` see the diffs:

```diff
     auth->mastodon = strdup(mastodon);
-    if (auth->url == NULL) {
+    if (auth->mastodon == NULL) {
        if (val_err != NULL) {
-       *val_err = werrp_sem_val(114, node, depth, sem, __func__,
+     *val_err = werrp_sem_val(119, node, depth, sem, __func__,
                                     "author array index[%d]: strdup of mastodon failed", author_num);
        }
```

```diff
     errno = 0;         /* pre-clear errno for werrp_sem_val() */
     auth->github = strdup(github);
-    if (auth->url == NULL) {
+    if (auth->github == NULL) {
        if (val_err != NULL) {
-       *val_err = werrp_sem_val(115, node, depth, sem, __func__,
+     *val_err = werrp_sem_val(120, node, depth, sem, __func__,
                                     "author array index[%d]: strdup of github failed", author_num);
        }
```

These could be not fun to debug of course but my eyes zoomed on these as you would expect.

xexyl:
> 
> ## ISSUE 1
> This one gave me some trouble. See the diff:
> 
> ```diff
> diff --git a/soup/chk.auth.ptch.c b/soup/chk.auth.ptch.c
> index 1cae0f6d2c5b6f5c91f5ef9277fb13f89538cc01..4313edc2eeaae2797812bd14bbd132790852c52a 100644
> --- a/soup/chk.auth.ptch.c
> +++ b/soup/chk.auth.ptch.c
> @@ -1,45 +1,26 @@
> ---- ref/auth.reference.json.c      2023-01-09 08:20:23
> -+++ chk_sem_auth.c 2023-01-09 08:20:03
> -@@ -40,25 +40,25 @@
> +--- soup/ref/auth.reference.json.c   2022-12-13 07:46:35
> ++++ soup/chk_sem_auth.c      2022-12-13 07:47:55
> +@@ -40,9 +40,9 @@
>   struct json_sem sem_auth[SEM_AUTH_LEN+1] = {
>   /* depth    type        min     max   count   index  name_len validate  name */
>     { 5,        JTYPE_NUMBER,   1,      5,      5,      0,      0,      NULL,   NULL },
> --  { 5,    JTYPE_STRING,   1,      114,    114,    1,      0,      NULL,   NULL },
> +-  { 5,      JTYPE_STRING,   1,      104,    104,    1,      0,      NULL,   NULL },
>  -  { 5,        JTYPE_BOOL,     1,      10,     10,     2,      0,      NULL,   NULL },
>  -  { 5,        JTYPE_NULL,     1,      1,      1,      3,      0,      NULL,   NULL },
> -+  { 5,    JTYPE_STRING,   16,     114,    114,    1,      0,      NULL,   NULL },
> ++  { 5,      JTYPE_STRING,   16,     105,    105,    1,      0,      NULL,   NULL },
>  +  { 5,        JTYPE_BOOL,     2,      10,     10,     2,      0,      NULL,   NULL },
> -+  { 5,    JTYPE_NULL,     0,      30,     30,     3,      0,      NULL,   NULL },
> ++  { 5,      JTYPE_NULL,     0,      25,     25,     3,      0,      NULL,   NULL },
> ```
> 
> In particular the changing of 105 to 114 and 25 to 30.
> 
> Without this it ended up being a count error. Since there's no name for those fields I'm not sure if there is something that should be done differently. My guess is that it's the amount of times that type of json object can be seen in the file full stop. The only thing that makes me wonder if that's not it is that 114 - 105 is not 10 but 9 and with five alternative URLs (one per the five authors) I'd think that the name `alt_url` and the value would be each one of the strings so 10 not 9.
> 
> The null type was increased by 5 which makes sense - if that's the idea behind it - because five more items that can be null or a string would be - well 5.
> 
> ## EDIT 0
> Typo fix in a number.

I notice in that file there are other unnamed lines with the same type and different values that don't seem to have any impact on the checks. I'm not sure if this is intentional or not but if it's not and is a problem now you know. But these make me wonder if my thought above is not the full story (never did actually think that even).

I'm going to make a comment in the Makefile variable issue and then going to do something else. If there are any issues here that need to be fixed I will do that but please let me fix it first as there are many changes and some (as noted above) are bug fixes that have nothing to do with this change. As far as fixing whether that's today or tomorrow would depend on the time but hopefully there's nothing that has to be changed - except perhaps adding the `alt_url` to more of the `bad/` files.

xexyl:
As far as the issue of the numbers being changed this is what happened without it:

```
What follows are semantic errors for .auth.json file: test_ioccc/test_JSON/auth.json/good/auth.reference.json
.auth.json count error sem_tbl[1]: count: 114 > max: 105 / error: node type JTYPE_STRING parse tree depth 5: found 114 > maximum: 105 
ERROR[1]: main: JSON semantic check failed for .auth.json file: test_ioccc/test_JSON/auth.json/good/auth.reference.json
ERROR[1]: main: JSON semantic check failed
```

A similar one happened for the other limit.

xexyl:
The missing `"alt_url"` lines have been added to all the bad/ json files.

lcn2:
Good catches and followup pull request, @xexyl Thanks!

xexyl:
> Good catches and followup pull request, @xexyl Thanks!

Thank you!

Did I interpret those values right then? Is there a reason that there's more than one of some of t hose?

lcn2:
> > Good catches and followup pull request, @xexyl Thanks!
> 
> Thank you!
> 
> Did I interpret those values right then? Is there a reason that there's more than one of some of those?

We think so.

xexyl:
.. and I'm mostly done for the day btw.

Just watching a film before I get ready for sleep. I can reply to the other issues tomorrow.

xexyl:
> > > Good catches and followup pull request, @xexyl Thanks!
> > 
> > 
> > Thank you!
> > Did I interpret those values right then? Is there a reason that there's more than one of some of those?
> 
> We think so.

Great. Thank you.

PR #477, opened 2023-01-11
Various enhancements, updates and fixes

xexyl:
We now collect shell options that are set (shopt -s).

When we're getting the bash version, if the basename is "bash", we get the values of the variables:

    $BASH_VERSION
    ${BASH_VERSINFO[0-4]} # as in each of the five indices
    $MACHTYPE

A note about the BASH_VERSINFO array. We do not string it up as one string but rather echo each element one at a time. This is a cosmetic choice as it would be a long line in the bug report. Additionally for each one the status is shown by not the element or array but what it stands for. For instance for ${BASH_VERSINFO[4]} we show something like:

    ## RUNNING: echo ${BASH_VERSINFO[4]}
    release
    ## BASH RELEASE STATUS ABOVE

The $MACHTYPE is the same value as ${BASH_VERSINFO[5]} but to be more clear we use $MACHTYPE. We still do it if it's the command passed to get_version is "bash" as it's a bash thing (though maybe other shells have it - I don't actually know now).

We still try --version after this just in case some or all of the above information could not be obtained.

An example output of the new information is as follows:

    ## RUNNING: shopt -s
    checkwinsize    on
    cmdhist         on
    complete_fullquote      on
    extquote        on
    force_fignore   on
    globasciiranges on
    globskipdots    on
    hostcomplete    on
    interactive_comments    on
    patsub_replacement      on
    progcomp        on
    promptvars      on
    sourcepath      on
    ## OUTPUT OF shopt -s ABOVE

    ## VERSION CHECK FOR: bash
    ## RUNNING: echo $BASH_VERSION
    5.2.9(1)-release
    ## $BASH_VERSION ABOVE
    ## RUNNING: echo ${BASH_VERSINFO[0]}
    5
    ## BASH MAJOR VERSION ABOVE
    ## RUNNING: echo ${BASH_VERSINFO[1]}
    2
    ## BASH MINOR VERSION ABOVE
    ## RUNNING: echo ${BASH_VERSINFO[2]}
    9
    ## BASH PATCH LEVEL ABOVE
    ## RUNNING: echo ${BASH_VERSINFO[3]}
    1
    ## BASH BUILD VERSION ABOVE
    ## RUNNING: echo ${BASH_VERSINFO[4]}
    release
    ## BASH RELEASE STATUS ABOVE
    ## RUNNING: echo $MACHTYPE
    aarch64-apple-darwin22.1.0
    ## $MACHTYPE ABOVE: BASH SYSTEM TYPE

Of course this is a lot more output but some of this might prove useful when trying to figure out a problem - even if we should normally get this information from bash --version (at least as far as the bash variables go).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,7 +359,7 @@ just_all: ${ALL_SUBDIRS} ${TARGETS} build_man
 # fast build environment sanity check
 #
 fast_hostchk: test_ioccc/hostchk.sh
-	-${Q} ./test_ioccc/hostchk.sh -f -v 0; \
+	-${Q} ./test_ioccc/hostchk.sh -f -v 0 -Z .; \
 	EXIT_CODE="$$?"; \
 	if [ "$$EXIT_CODE" -ne 0 ]; then \
 	    ${MAKE} hostchk_warning; \
@@ -383,7 +383,7 @@ bug_report-txl: bug_report.sh
 # slower more verbose build environment sanity check
 #
 hostchk: test_ioccc/hostchk.sh
-	-${Q} ./test_ioccc/hostchk.sh -v 1; \
+	-${Q} ./test_ioccc/hostchk.sh -v 1 -Z .; \
 	EXIT_CODE="$$?"; \
 	if [ "$$EXIT_CODE" -ne 0 ]; then \
 	    ${MAKE} hostchk_warning; \
--- a/bug_report.sh
+++ b/bug_report.sh
@@ -57,7 +57,7 @@ export TOOLS="
     ./txzchk
     "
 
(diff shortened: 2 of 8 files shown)


xexyl:
I'm going to look at the todo list now and I hope to do at least one or two of those things. I will have to leave for a bit in a while but I probably can return to this after that.

lcn2:
There are some really good changes here, thanks @xexyl

xexyl:
> There are some really good changes here, thanks @xexyl

Thank you.

I hope to do a bit more tomorrow but my doctor appointment is kind of early so it might be I won't get as much done unless I'm up to more after I get home again.

Good luck on preparing for the class you'll be giving!

PR #479, opened 2023-01-14
Bug fixes and enhancements

xexyl:
A funny problem with commit 89f8a4b9d9b6f3533b3577398dbd559f09e27ecc happens under linux which makes a number of steps in prep.sh fail. This is because a number of Makefile rules depend on make all. But since a compilation error occurs under linux due to the change in the commit they all fail.

The problem is:

    In file included from jnum_test.c:37:0:
    jnum_chk.h:83:13: error: 'quiet' defined but not used [-Werror=unused-variable]
     static bool quiet = false;    /* true ==> quiet mode */
		 ^

This is because jnum_test.c does not use quiet but as described in the static_analysis.md file to have another header file would be problematic in another way. The solution therefore is to add another include guard. Doing this solves the problem.

This raises a question of whether or not the other rules (including some that do not even work on source files) should depend on make all but that's another matter entirely.

As part of make prep seqcexit has sequenced the exit codes in util.c.


xexyl:
I think you might enjoy commit c6a46b9f919aee9922da466ee48d52063c3a7094:

    Have some fun throwing tar and feathers :-)
    
    Rename the txz_info struct and variable to tarball. This way one can
    look at the code and have things like:
    
                ++tarball.total_feathers;
    
    instead of
    
                ++txz_info.total_feathers;
    
    which is not nearly as fun - and does not go as well with the joke of
    tarring and feathering that I added ages ago.
    
    The function show_txz_info was renamed to show_tarball_info as
    show_tarball() would suggest showing the contents - but that's done
    indirectly via the tar command that txzchk invokes.
    
    There might be future fun updates here but for now it'll suffice.

xexyl:
I'm aware of a problem with make depend in one of these commits. I'm on a zoom call and I'll fix it after that.

xexyl:
> I'm aware of a problem with make depend in one of these commits. I'm on a zoom call and I'll fix it after that.
> 
See updated comment - fixed. I am only making a new comment about it so that you might see it more easily.

xexyl:
I want to apologise to you, Landon, for having such huge pull requests. I got so carried away with the fixing the warnings that it did not dawn on me just how many commits there were.

I am about to make another commit which will complete the final checks on exit codes consistency but that might be all I do today. Yesterday was hard and I need to recover a bit.

lcn2:
There is a good deal of careful and detained work in this pull request.  No problem with the size .. sorry (tm Canada 🇨🇦) for the long delay in processing this excellent work, @xexyl

xexyl:
> There is a good deal of careful and detained work in this pull request. No problem with the size .. sorry (tm Canada 🇨🇦) for the long delay in processing this excellent work, @xexyl

No worries and thank you! I hope that the class you taught went well! I hope to do more tomorrow. I will reply to any other comments then as well. 

Good night!

xexyl:
BTW: I'm not sure if you saw it but if not you should check out comment https://github.com/ioccc-src/mkiocccentry/pull/479#issuecomment-1382868639 :)

PR #480, opened 2023-01-20
More updates

xexyl:
Because the other pull request is so large I'm making a new one. I'm hoping to get a few more commits in today. Right now I'm having a difficult time figuring out the error messages of `fnamchk` .. something doesn't seem right. It might be fine as is and I'm more tired than I think (very possible) but it seems there are some mistakes and I wonder if there are other ways to word it to make it make more sense anyway.

What I do hope to do today is finish the checking of the exit codes.


xexyl:
I want to apologise to you, Landon, for having such huge pull requests. I got so carried away with the fixing the warnings that it did not dawn on me just how many commits there were.

I am about to make another commit which will complete the final checks on exit codes consistency but that might be all I do today. Yesterday was hard and I need to recover a bit.

lcn2:
Thank you for the detailed set of changes, @xexyl

xexyl:
> Thank you for the detailed set of changes, @xexyl

You're very welcome as always my friend!

PR #482, opened 2023-01-23
Minor adjustments to variable rename (txzchk.[ch])

xexyl:
With commit b5fc2a3658e8eeea711eb908a2fd6c6806b58bae the txzpath was changed to txzpath_path which seems like it might not be what was intended. In any case it seems clearer as tarball_path so that's what it now is called in all cases including the man page and usage message.


xexyl:
Oops ... I made an error in the commit log. Redoing this.

xexyl:
...Oh I didn't make an error. Okay I'll redo this and then stop for now. Not that I felt like I could do much after it anyway, for now, but better not to do anything important when this tired.

PR #488, opened 2023-01-25
Explain Wpoison-system-directories

xexyl:
Due to another issue I was checking the built-in rules of Makefiles and I noticed the source of the problem of why indeed the /usr/local/include is searched even though we don't explicitly add it. It's not a problem usually and under linux there is no warning but macOS does have a problem with it. We already noted that this can be ignored but previously stated it is bogus. But it's actually because of the variable:

    .INCLUDE_DIRS

which appears to be read-only as well. I'm not sure if all implementations of make have this variable but GNU make certainly does.

There still are some commit ids that need to be removed or else fixed but this can come later.

### DIFF
--- a/test_ioccc/static_analysis.md
+++ b/test_ioccc/static_analysis.md
@@ -495,7 +495,7 @@ What you'll see is something like:
 
 
 ```sh
-$ make prep
+make prep
 make_action 10 clobber OK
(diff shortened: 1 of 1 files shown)


xexyl:
Closing this .. mistake in log message.

PR #489, opened 2023-01-25
Explain -Wpoison-system-directories

xexyl:
Due to another issue I was checking the built-in rules of Makefiles and I noticed the source of the problem of why indeed the /usr/local/include is searched even though we don't explicitly add it. It's not a problem usually and under linux there is no warning but macOS does have a problem with it. We already noted that this can be ignored but previously stated it is bogus. But it's actually because of the variable:

    .INCLUDE_DIRS

which appears to be read-only as well. I'm not sure if all implementations of make have this variable but GNU make certainly does.

There still are some commit ids that need to be removed or else fixed but this can come later.


xexyl:
With d288159706d5942bf93bfd52e77dbc102fc0f739 man section 3 pages now have been checked for using macros in the SYNOPSIS. There are a couple lines in those that do not use the macros in the synopsis but I'm not worried about those as it simplifies it.

This means that I can now work on consistency and typos in the man pages.

I'm not sure if I'm up to that right now and I might not be today either .. we'll have to see but right now I don't feel that well. However I can't rest so I might keep at it.

xexyl:
With commit 3e98f10b7aecee899e79840d48a294b53553cb03 some of the jparse/man/man8 pages have been updated with macros for SYNOPSIS and OPTIONS.

But I am too tired to continue so the rest will have to be done another time. Those are:

        ./run_bison.8
        ./jsemtblgen.8
        ./verge.8
        ./jstr_test.8
        ./run_flex.8

but whether those will be done later today or not I don't know.

I'm starting to fear that I won't finish this stuff before the end of January. :( Sorry about that .. been quite exhausted and busy as well. Hopefully I can finish it in the first part of February.

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Sure.

I have a segfault fix in mkiocccentry ... one moment.

PR #490, opened 2023-01-26
Fix segfault when author name too long

xexyl:
The problem was a missing 'else' in front of the 'if' that happened right after the free and set to NULL author name. Perhaps this was missed because of the way the comments were below the last if.

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -3649,12 +3649,11 @@ get_author_info(struct author **author_set_p)
 		    free(author_set[i].name);
 		    author_set[i].name = NULL;
 		}
-	    }
 
 	    /*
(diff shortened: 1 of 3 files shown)


lcn2:
good catch, @xexyl

xexyl:
> good catch, @xexyl

Yes. Easy to happen. I guess it really is to do with the comments. See, comments are actually bad for you! :-)

Well maybe not but I think in this case it did make it harder to see that there was a missing 'else' to the 'if'.

I'll be leaving shortly. I hope I can do a bit tomorrow before I have to leave but if not I can surely do more Saturday. It's unlikely I'll be up to doing anything when I'm home tomorrow.

PR #492, opened 2023-01-27
Fix make clobber (via legacy_clobber rule)

xexyl:
The legacy_clobber rule deleted '.build.log.' but there won't be any files with that name: in the Makefile the variable TMP_BUILD_LOG is defined as .build.log.$$$$ which translates to the shell $$ which is the PID so .build.log.<pid> where <pid> is the pid of the command. Thus we now delete .build.log.* in case the rule is interrupted before it can be renamed to 'build.log'.


lcn2:
Thanks 🙏 @xexyl

xexyl:
> Thanks 🙏 @xexyl

I'm happy to help!

PR #493, opened 2023-01-27
Typo fix in soup/location.c

xexyl:
Although ions have their place when peroxide ions in particular combine with water and diluted acids it forms hydrogen peroxide which has ended up killing people when those unaware of the dangers misunderstand the request that someone would like a glass of H2O too. Thus we should get rid of stray ions in the repo and now have.

Yes there's a point (or actually two points) in the above. In order to help discern what they are you should start by looking at the diff.

### DIFF
--- a/soup/location.c
+++ b/soup/location.c
@@ -475,7 +475,7 @@ check_location_table(void)
  * lookup_location_name - convert a ISO 3166-1 Alpha-2 into a location name
  *
  * Even though the ISO 3166-1 Alpha-2 code is only UPPER CASE, the
- * code is compared ion a case independent way.  That is, a non-canonical
+ * code is compared in a case independent way.  That is, a non-canonical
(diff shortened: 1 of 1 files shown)


xexyl:
I hope you enjoy the joke (or jokes) :-)

Will be leaving fairly soon.

lcn2:
Fun, thanks @xexyl

xexyl:
> Fun, thanks @xexyl

Quite welcome and I am glad you got a laugh or two out of it!

I also hope you’re having a nice sleep. I slept in a bit .. still waking up but I hope that I can get working on the repo soon. Going to get laptop out shortly.  

First thing I will work on is the usage() function changes we discussed yesterday. Then we shall see. 

But I do hope that I will be able to get a lot done today. Well it’s time to get the laptop out and start if I can!

PR #494, opened 2023-01-28
Better usage() usage in all tools + fix exit codes + many man pages updated

xexyl:
The usage() function in all tools (C files) now have the same order of parameters: exit code, program name, string. Previously the second and third were swapped but it makes more sense to have the program name first. Also some tools had extra parameters which were confusing and made some assumptions on the call to the functions (specifically argc and required number of args but this was a problem in how it checked for and what message it printed based on it).

Remove superfluous messages in usage() calls like "-h help mode" and "invalid -flag".

Control getopt() error messages. Via the new function in util.c check_invalid_option it mimics the getopt() but we control it in a way that the usage() function does not end up printing things out of order and redundantly so. Now if one does -h it will just show the usage; if one does not provide an option argument it reports this and then prints the usage (the latter part like before) (previously it would print the missing option argument via getopt() and then also say invalid -flag); and if an invalid option is specified it is also specified right before using usage() (previously it also said invalid -flag or similar right after getopt()'s error message).

We control this by putting in front of the optstring a ':' and then updating the switch() to add the cases of '?' and ':' which both (fall through) the new function mentioned above. We do not have it call usage() because this function is private to each tool.

In some cases there were exit codes that were incorrect (as in wrong exit code used for invalid command line).

In one case (jsemtblgen.c) the usual 'i' was a size_t not used for getopt() but as an interator (instead getopt() used an int c). These two variables were swapped to be consistent and also so that the call to the new function is the same in each file.

Updated comments for usage() functions to be correct usage example (except for the adding of /*ooo*/ as that would be a problem as it's within a /*..*/ comment).


xexyl:
I think, Landon, that you will get amusement out of commit https://github.com/ioccc-src/mkiocccentry/pull/494/commits/f9d3d99a5029c26f697e8a07537db7dbd4dd65f7 and a231dba384d0ea86ae9291965fcb938879d15f59:

    There were some places (namely txzchk.1 and txzchk_test.8) where a
    certain name was bold (not in all places but in some) namely my name. To
    make it more consistent with all the other names (and my name in other
    places - except perhaps the README but I would have to check on that
    one) I have decided to be less bold. :-)


...

    I've decided that Landon Curt Noll also needs to be less bold so I have
    removed his boldness from the man pages! :-)

lcn2:
Thank you @xexyl

xexyl:
> Thank you @xexyl

Very welcome.

I'm working on more error message checks. Hoping I can finish that before I leave for the day. Making good progress but still a few more files to go and some will take some time though I hope that the one that I'm working on will be the longest (mkiocccentry.c).

No idea how to do this with the scripts .. guess I should look at lines that use `echo` and just go through all of them.

PR #495, opened 2023-01-29
Check final C file for error messages

xexyl:
With the exception of perhaps dyn_array.c all error messages have been checked. It is inevitable that I missed some things as it was a lot to go through and I did not do it by scrolling down from the top line by line but rather in ways that would help find errors and similar messages.

In mkiocccentry.c other types of messages were checked and in some cases improved.

One directory that might need to be rechecked for debug messages is jparse/ as I did not look for json_dbg() (I did under soup/) but this can happen another time.

Besides that the only other error (and related) messages to check are those in shell scripts.

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -370,7 +370,7 @@ main(int argc, char *argv[])
 	dbg(DBG_HIGH, "info.tstamp: %jd", (intmax_t)info.tstamp);
     }
     info.usec = tp.tv_usec;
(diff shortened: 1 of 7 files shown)


lcn2:
Thank you @xexyl !!

xexyl:
> Thank you @xexyl !!

You’re very welcome my friend!

As I noted elsewhere there are a few places left but mostly it’s done. I will also tackle the shell scripts and their error messages tomorrow. I should think that that will go fairly quickly. 

It’s dyn_array.c I am not entirely sure of just yet.

PR #497, opened 2023-01-30
Various updates, fixes and enhancements

xexyl:
As for rewording: one should not try downloading and installing GNU tar at the GNU website as they'll very likely be unable to log in to it :-) Instead suggest that they can try downloading it FROM the website and then compile and install it so the tools may find it.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -51,13 +51,14 @@ for more details on what you can do to help us.
 
 
 If your tar does not support the `-J` option you can either use a system
-which does have such a tar or you can try downloading and installing GNU Tar at
-the [GNU website](https://www.gnu.org/software/tar/).
+which does have such a tar or you can try downloading GNU Tar from
+the [GNU website](https://www.gnu.org/software/tar/) and compile and install it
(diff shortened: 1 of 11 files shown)


lcn2:
Thank you 🙏 @xexyl

xexyl:
Glad you got a laugh out of my log entry.

You might also like:

```sh
git diff 70780a4ae587454ee0f4ca49b36023c1300158be..e39371ada21df980b18f236aaacf71807256d544 bug_report.sh
```

specifically the part about `cat` and `yes`.

xexyl:
Or better yet:


```sh
# type -a cat: find which cat the user owns :-) (okay - types of cats)
#           
#   We didn't mean no harm, but they jumps on us like
#   cats on poor mices, they did, precious.
#           
type_of 15 "cat"

# type -a yes: get all yesses (yes Preciouss, yesses is a valid spelling just as yeses is :-) )
type_of 36 "yes"

```

though Tolkien might have taken an issue with part of that.

PR #501, opened 2023-01-31
Improve getopt() calls + update README.md

xexyl:
In utf8_test.c we now have a usage() function. VERSION macro renamed UTF8_TEST_VERSION, usage string now named usage_msg. Now uses the function check_invalid_option() in jparse/util.c (it linked in the jparse.a anyway).

Further compress the cases in the getopt() switch blocks so that default is in the case '?' and case ':' as the function check_invalid_option() only acts if the char is ':' or '?' so it will effectively be a no-op anyway. But since the default: does the same as the others except it does not call the check_invalid_option() function this works well. The default case probably is not even needed but we have it there for safety.

Update CHANGES.md. That task should be able to be ticked though I did not remove the XXX line of adding more as more is remembered or done. I went through the git log so that the former part is resolved but if any last minute changes are made they should be added.


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Sure. 

I hope you have a good night. It’s been a very long day but I’m afraid that it is still a bit too early to sleep. 

I will be trying in about an hour probably so anything after that I can address tomorrow.

PR #503, opened 2023-02-01
Various issues detected by clang scan-build fixed

xexyl:
As for commit 1e64743d6d81a261c8c3e334b928a534a7aab2d0 which has to do with jparse:

We ignore the warnings as there is nothing wrong with our jparse grammar
    (other places, perhaps, but not there :-) ) and we can't modify the
    generated code anyway.


xexyl:
> Thank you 🙏 @xexyl
> 
> We are doing maintenance on our primary development system, which has delayed our return for much of the day.




> Thank you 🙏 @xexyl

I am happy to help! Can you think of any last minute changes to FAQ that might be needed or desirable ?

> 
> We are doing maintenance on our primary development system, which has delayed our return for much of the day.

Hope it went well! That reminds me that recently a new version of fedora was released and I better upgrade my system sooner than later. That's the downside with staying on the bleeding edge - have to update regularly. Of course that's only for one system. CentOS is more stable but sadly Red Hat axed it. Next year it will reach EOL. I think June so I will have to upgrade to something else before rather than just upgrading (though that would also require a reinstall most likely).

Ah well .. I need to fix the LVM a bit anyway but still not looking forward to it. I'll choose rocky linux instead. The ridiculous thing about Red Hat doing this is that they killed CentOS 8 before 7. Fortunately I had not installed 8 yet mostly because (I had to put a new install in as I upgraded the server and wanted raid 5, not raid 1, amongst other things) I hate dnf.


I hope you're having a nice sleep. I had a difficult night and been awake somewhere between midnight and 1 ..


... but I see that you're awake since I just got an email from an other thread here from you ..

PR #504, opened 2023-02-02
Fix coredump in verge + remove dead functions

xexyl:
Removed many unused string_to functions and made verge use string_to_intmax().

When testing the above change I discovered an invalid free() resulting in a coredump in verge. For instance under linux (in this case fedora):

    $ ./jparse/verge j.0.0 2.2.2
    free(): invalid pointer
    Aborted (core dumped)

and under macOS:

    $ ./jparse/verge j.0.0 2.2.2
    verge(21105,0x1de8a0100) malloc: *** error for object 0x6000029f0012: pointer being freed was not allocated
    verge(21105,0x1de8a0100) malloc: *** set a breakpoint in malloc_error_break to debug
    Abort trap: 6

The problem was that the first non-numerical character was the very first character and the fact that the duplicated string (wstr) was being freed from an offset. If a non-digit is found in the string later it was okay but not at the beginning. Making a new pointer (wstr_start) that simply points to the beginning of the duplicated string and then freeing that instead of wstr solves the problem. Now one will get:

    $ ./jparse/verge j.0.0 2.2.2
    $ echo $?
    1

which means that the first version is < second version which is correct, documented behaviour.

As for the unused functions that were removed:

    extern long string_to_long(char const *str);
    extern long long string_to_long_long(char const *str);
    extern int string_to_int(char const *str);
    extern unsigned long string_to_unsigned_long(char const *str);
    extern unsigned long long string_to_unsigned_long_long(char const *str);
    extern uintmax_t string_to_uintmax(char const *str);
    extern long double string_to_float(char const *str);

Now everything works as it should and the crash no longer happens either.

Updated CHANGES.md for these changes.


lcn2:
Thank you @xexyl

xexyl:
> Thank you @xexyl

Very welcome! 

I look forward to seeing what you are working on. Just finishing with comments and then will get a snack and then sleep. Or I might try sleeping first in hope that I am not hungry enough that ut will keep me awake.

PR #505, opened 2023-02-04
Minor updates to party.json files

xexyl:
The bad/party.json file had invalid json. Instead make a new file which has bad json and let the other one be invalid in semantics.

Some of this will have to be changed when the time comes for the jparse repo because it's too complex now (and this commit makes it more complex in a way). I was going to remove some of the files but as we're essentially at the point of freezing the code I wanted to get this in quickly and easily.


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Thank you as well!

I look forward to seeing what you do today.

Just reading.

Have a great day!

PR #507, opened 2023-02-05
Improve some of the test scripts

xexyl:
In order to test other implementations of tar the txzchk_test.sh script has been enhanced to take an optional tar path. To make the tool more consistent with other tools (except in a way txzchk itself, see below) the script now is -t tar and -T txzchk. This is because in mkiocccentry that's the meaning of the options.

The mkiocccentry_test.sh and ioccc_test.sh scripts now also take -t tar and mkiocccentry_test.sh additionally takes an option for cp (-c cp), ls (-l ls), txzchk (-T txzchk) and fnamchk (-F fnamchk).

The way these changes are not consistent with txzchk is simply that txzchk takes a -T option to set text file (or if you prefer 'test') mode and -t for tar and additionally mkiocccentry takes a -T txzchk so the scripts that now take a -T that doesn't set txzchk won't match but then one would hope that txzchk does not need a path to itself! :-)

Updated man pages.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -231,10 +231,6 @@ The `make tags` now forms both local directory tags files by the name of
 .local.dir.tags as well as a tags file from the accumulation of tags from
 related source directories.
 
-Major updates to CHANGES.md. See
-[CHANGES.md](https://github.com/ioccc-src/mkiocccentry/blob/master/CHANGES.md)
-for details. :-)
-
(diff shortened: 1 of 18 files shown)


xexyl:
Now to solve the chkentry_test.sh script not working when directly invoked. This might or might not be worth it. We shall see.

xexyl:
> Now to solve the chkentry_test.sh script not working when directly invoked. This might or might not be worth it. We shall see.

... been done as I noted in the other thread. Simple matter of updating the default path. Must not have been done when the files were moved (I guess that's how it happened).

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

I am happy to help!

I am just sitting in bed. Will be awake a while longer though I don’t have the laptop out. 

I can possibly reply to some comments before I sleep if you make any. 

Otherwise I will follow up tomorrow.

PR #509, opened 2023-02-22
Make hostchk grep usage POSIX compliant

xexyl:
For whatever reason POSIX grep does not require -o or -h options which we used before. The -o we never actually needed but the -h we do need because we check more than one file and need only the contents not the files the contents are in.

Instead of using grep -o -h we pipe the grep to `cut -f 2- -d:' and pipe that to sort -u (the order is important). This will remove the filenames from the output and it'll show the exact lines we need but in a POSIX portable way. This is because (as part of POSIX) if more than one file is specified to search then the output is prefixed with:

    "%s:", <file>

but we only need the second field in all our cases.

This makes hostchk.sh work on hosts that do not have one or both of grep -o or grep -h options.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@ Test if ctags is available before attempting to run ctags in Makefile rules.
 The `jparse/Makefile` `make clobber` rule will call the same rule in
 `jparse/Makefile/test_jparse` as well.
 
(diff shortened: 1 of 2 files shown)


lcn2:
Thanks @xexyl

PR #513, opened 2023-03-09
Fix up to date issues in make parser rule

xexyl:
An annoying problem was in jparse/Makefile when running make parser that made it appear that the scripts were not running. There are two parts to this.

First the script was set to 0 verbosity level and I did not realise this so it seemed as it was not running (only obscure hints that it was). The second part is that the make parser rule had the lines:

	${E} ${MAKE} jparse.tab.c jparse.tab.h
	${E} ${MAKE} jparse.c jparse.lex.h

and the problem with that is that the first of each implies the second so we would get the messages:

    make[2]: `jparse.tab.h' is up to date.

    make[2]: `jparse.lex.h' is up to date.

This made me believe that the problem I was facing was the scripts were not running but instead it's that the rule was run twice which is unnecessary. Thus a comment above each one notes that the rule implies the other and no longer runs both.

### DIFF
--- a/jparse/Makefile
+++ b/jparse/Makefile
@@ -560,10 +560,12 @@ man/man8/jparse_test.8:
 #
 parser: jparse.y jparse.l
 	${RM} -f jparse.tab.c jparse.tab.h
-	${E} ${MAKE} jparse.tab.c jparse.tab.h
+	@# make jparse.tab.c implies make jparse.tab.h too
+	${E} ${MAKE} jparse.tab.c
 	${E} ${MAKE} jparse.tab.o
 	${RM} -f jparse.c jparse.lex.h
-	${E} ${MAKE} jparse.c jparse.lex.h
+	@# make jparse.c implies make jparse.lex.h too
+	${E} ${MAKE} jparse.c
 	${E} ${MAKE} jparse.o
 	${RM} -f jparse.tab.ref.c
 	${CP} -f -v jparse.tab.c jparse.tab.ref.c


xexyl:
I've decided to just make the pull request now .. the other change can be fixed another time.

lcn2:
We recommend that you also make a change to the `CHANEGS.md` file documenting this change, @xexyl.

There is no need to update version numbers for this change.  Moreover it will become part of the 1.0.1 release when that happens.

Thanks!

xexyl:
> We recommend that you also make a change to the `CHANEGS.md` file documenting this change, @xexyl.
> 
> There is no need to update version numbers for this change. Moreover it will become part of the 1.0.1 release when that happens.
> 
> Thanks!

Good idea .. I'll do that though I'm trying to fix the problem with prefix. Just doesn't want to work. I am very tired (woke up a bit after 3, got up at 330) and was chatting with a friend but I might need a break. I will certainly update this file though soon.

Hope you're having a nice sleep my friend!

xexyl:
Even though you already know: the CHANGES.md was updated in commit https://github.com/ioccc-src/mkiocccentry/pull/514/commits/354dc09278f3783b80995de613fc74a9d73794ac. However I think I left off the renaming of the malloc_path to calloc_path as I wanted that in the correct commit but then forgot it.

I'm going to be making another pull request here (minor one) which I'll update the CHANGES.md (already did updated for the commit I'm about to make) if I did indeed leave it out.

Thank you for the reminder and I do hope you're having a nice sleep my friend!

PR #514, opened 2023-03-10
Add jparse_ prefix to json parser - new version

xexyl:
This is so that one can have multiple parsers in a program that links in
    the jparse library. This took extra effort for some reason but it should
    be okay. Nevertheless we have yet to test jparse in other applications.
    New version of jparse is 1.0.1 2023-03-10.
    
    Add jparse.c to make depend list in jparse/Makefile and run make depend.
    The file jparse.c was missing in the variable BUILT_C_SRC and this was
    potentially causing problems as one of the issues in working out the
    jparse_ prefix (which really shouldn't have been as complicated as it
    turned out to be).
    
    Updated CHANGES.md for these changes as well as the fix in
    jparse/Makefile in commit 18cd69d63aa57b82ce6dfd78094b68f40ad21e0c made
    yesterday.

This is all I'll do today but I hope to return to the other repo tomorrow. Off to do other things. Good day!


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

You're most welcome my friend.

xexyl:
That's funny .. only the second commit had the checks applied? Yet the jparse one had many more changes; the second one was simply a name change. See:

Screenshot 2023-03-11 at 04 33 45

Ah well ... both should be good anyway. I'm not sure about this part and I want to get it to work otherwise but it might have to happen another time so I can get back to the [other repo](https://github.com/ioccc-src/temp-test-ioccc):

```
/*
 * jparse.tab.h - generated by bison
 */
#if !defined(YY_JPARSE_TAB_H_INCLUDED)
#define YYSTYPE JPARSE_STYPE
#define YYLTYPE JPARSE_LTYPE
#define YY_JPARSE_TAB_H_INCLUDED
#include "jparse.tab.h"
#endif
```

and specifically the lines:

```c
#define YYSTYPE JPARSE_STYPE
#define YYLTYPE JPARSE_LTYPE
```
as we didn't have to do that before with the satirical prefixes. It's for flex(1).

BTW: the include guard define is actually `YY_JPARSE_JPARSE_TAB_H_INCLUDED` but changing it to that results in another compiler error (or errors):



```c


./jparse.h:99:23: error: unknown type name 'JPARSE_STYPE'
extern int jparse_lex(JPARSE_STYPE *yylval_param, JPARSE_LTYPE *yylloc_param, yyscan_t scanner);
                      ^
./jparse.h:99:51: error: unknown type name 'JPARSE_LTYPE'
extern int jparse_lex(JPARSE_STYPE *yylval_param, JPARSE_LTYPE *yylloc_param, yyscan_t scanner);
                                                  ^
./jparse.h:99:79: error: unknown type name 'yyscan_t'
extern int jparse_lex(JPARSE_STYPE *yylval_param, JPARSE_LTYPE *yylloc_param, yyscan_t scanner);
                              
...

jparse.tab.c:1214:1: error: use of undeclared identifier 'JPARSE_LTYPE'
YYLTYPE yylloc = yyloc_default;
^
./jparse.h:72:17: note: expanded from macro 'YYLTYPE'
#define YYLTYPE JPARSE_LTYPE
                ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]

```

which seemed to not make sense to me as that header file is the one that creates the struct and typedefs!

Still though these should probably be worked out. I'm not clear on how exactly the names don't clash with other parsers as the parser/scanner just define the yy prefix to be equal to the prefix. I should maybe find out as it seems like it would still have name conflicts.

PR #515, opened 2023-03-11
Rename MAX_(NUL|LOW)_REPORTED to have BYTES

xexyl:
Rename macro MAX_NUL_REPORTED -> MAX_NUL_BYTES_REPORTED and allow overriding it (default is 5; uses #if !defined (MAX_NUL_BYTES_REPORTED).

Rename macro MAX_LOW_REPORTED MAX_LOW_BYTES_REPORTED and allow overriding it (default is 5; uses #if !defined (MAX_LOW_BYTES_REPORTED).

Fix version of jparse; the prefix added was supposed to be 1.0.1 2023-03-10 and was changed initially but when trying to get a problem resolved it must have been undone. Now it's instead 1.0.2 2023-03-11.

Updated CHANGES.md for the above as well as adding the fix of malloc_path() to calloc_path() made in commit
a680a98247a9d3c975fa4a466c9db703a66e1f28. This was not done because when I updated the CHANGES.md in commit
354dc09278f3783b80995de613fc74a9d73794ac I did not want to update it for this function change as I did not commit those changes then as that commit was something else; I just discovered the name discrepancy prior to the first commit (354dc09278f3783b80995de613fc74a9d73794ac).


lcn2:
THanks @xexyl

xexyl:
> THanks @xexyl

You're most welcome!

PR #516, opened 2023-04-14
Fix writing of answers file (past winner, handle)

xexyl:
When the past winner and winner handle prompts were added to the mkiocccentry tool the tool was changed to read the values from the answers file but it did not write it to the answers file.

But why did the test script work? It's simply because the script (of course) does not run the interactive tool but merely reads from the answers file which the script itself creates.

Now one can use the -a/-A feature and expect -i to work.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.2 2023-04-14
+
+Fix mkiocccentry to write past winner and author handle to the answer file. It
+already read from these from the file but did not write them and therefore using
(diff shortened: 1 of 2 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

My pleasure! And really indeed mean my pleasure! Been having A LOT of fun with all these! I will continue finding entries that no longer work _after_ I finish consistency fixes. I decided that's the better way to do it even if it kind of interrupts each README.md. Maybe it's the wrong way or not useful since it all has to be done and since you've merged quite a few pull requests I might want to not worry about that :-) but we'll see!

PR #517, opened 2023-04-24
Fix warnings compiling jparse

xexyl:
Two warnings were being triggered under macOS: one about unused but set
variable (in generated code). That was solved by adding
-Wno-unused-but-set-variable to the WARN_FLAGS. Note that clang in some
platforms (like CentOS 7) do not have that option but it is not fatal so
this is okay (though in that platform it would generate additional
warnings).

The other was a more recent one:

    json_util.c:129:18: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
    json_warn_allowed()
                     ^
                      void

The solution is to add void to the function arg list.

Aesthetic changes that don't really matter but were done and by accident
made in this commit (I already saved and quit - this is an emendation): 
off-by-one char in a comment and better wording of a comment as well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.2 2023-04-14
+
+Fix mkiocccentry to write past winner and author handle to the answer file. It
+already read from these from the file but did not write them and therefore using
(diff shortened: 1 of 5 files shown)


xexyl:
> Thanks @xexyl

You are most welcome my good friend! Safe travels / everything in the outback!

PR #524, opened 2023-05-25
Initial jprint code

xexyl:
Right now it only links in the jparse, dynamic array and debug libraries and when run it exits 0.

The Makefile has been updated and make depend was run as well.

This tool will I believe take some time and careful thought and it definitely will be developed over many commits but now it should be possible to just start writing the code without having to worry about editing the Makefile etc. More editing of the Makefile will have to be done later - for instance when man pages are added or some dependency changes or if a new file has to be added but at this point I can start working on the code once I have a better idea of what is needed. This won't happen until a discussion is had on GitHub though I might be able to get some code in prior to that should I find the time and energy. I do hope to do a bit more on this later today.

Updated .gitignore: add jparse/jprint.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ jparse.tab.c
 jparse.tab.h
 jparse.xml
 jparse/jparse
+jparse/jprint
 jparse/lex.yy.c
 jparse_test.log
 jsemtblgen
--- a/jparse/Makefile
+++ b/jparse/Makefile
@@ -164,10 +164,10 @@ CFLAGS= ${C_STD} ${C_OPT} ${WARN_FLAGS} ${LDFLAGS}
 # source files that are permanent (not made, nor removed)
 #
 C_SRC= jparse_main.c json_parse.c json_sem.c json_util.c \
-       jsemtblgen.c jstrdecode.c jstrencode.c util.c verge.c
+       jsemtblgen.c jstrdecode.c jstrencode.c util.c verge.c jprint.c
 H_SRC= jparse.h jparse_main.h jsemtblgen.h json_parse.h json_sem.h json_util.h \
        jstrdecode.h jstrencode.h sorry.tm.ca.h util.h verge.h \
-       jparse.tab.ref.h
+       jparse.tab.ref.h jprint.h
 
 # source files that do not conform to strict picky standards
 #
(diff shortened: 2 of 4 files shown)


xexyl:
... taking a break now .. will try and reply to the issue later on today but this is a good start even if the code does nothing for now. It'll help for later. One thing I noticed is we should come up with a usage string. I'll bring that up in that issue later on most likely today.

lcn2:
A start! @xexyl

xexyl:
> A start! @xexyl

Yes! A lot more to go but it's a good start indeed!

lcn2:
We are "_clearing the decks_" in preparation for an important [temp-test-ioccc pull request](https://github.com/ioccc-src/temp-test-ioccc/pull/811).  See [temp-test-ioccc comment 1567576879](https://github.com/ioccc-src/temp-test-ioccc/issues/5#issuecomment-1567576879) for details.

xexyl:
> We are "_clearing the decks_" in preparation for an important [temp-test-ioccc pull request](https://github.com/ioccc-src/temp-test-ioccc/pull/811).  See [temp-test-ioccc comment 1567576879](https://github.com/ioccc-src/temp-test-ioccc/issues/5#issuecomment-1567576879) for details.

I just look forward to it being resolved! 

The delays for pull request being merged is well worth it. I will end up saying more in the other issue. 

It's also a rather convenient time for this as I have some other things I have to deal with lately. 

I am just thankful that yesterday I had the energy and patience (BOTH are important for something like this!) to get the last HDD into the raid6 volumes. Though I might end up making the root VG a bit smaller: not sure yet. But that will depend on if I need more space in the data VG. 

If I am thinking right in my tired head the fact that I have raid6 volumes (except for `/boot` of course which is raid1) with five disks means I can have up to three disks dying before I would lose any data which is a nice reassurance though of course I do have two forms of nightly backups. 

My next expense will be to get more RDX cartridges though the most important is to get a larger one. I now have two docking stations so I can use one for the 1TB which might be strictly for the compressed backups (i.e. bacula) and the other for uncompressed but I will figure this all out at the time which will not be for some months at the very least. 

But I do feel great that a major stressor is gone!

I don't know if it will help with the recent sleepiness but it will make things a bit easier nonetheless. 

Hope you're sleeping well my friend!

I can't rest and I am more awake now so I hope that I can reply to the actual comments now. Just going to open the laptop lid and then the tab in Safari.

PR #531, opened 2023-06-03
New jprint version - test and report valid JSON

xexyl:
New number of required args is 1: the file to check. It was an error to
require a pattern to match according to the spec and usage string as
well.

The jprint tool now checks that the first arg exists as a file, is a
regular file and that it can be opened for reading. If any of these are
not true it is an error. Otherwise the file is checked for valid JSON.

Once file is parsed (or we attempt to parse) as JSON the file is closed
and the file pointer is set to NULL. If invalid JSON print invalid and 
exit. If valid it depends on whether another arg is specified or not. If
another arg is specified we will show the pattern requested (this will 
not be done later on) and exit 0; otherwise we will print that no 
pattern is requested (this will also not be done later) and exit 1.

Besides the initial parsing of options (to the extent that we have 
details) we can do no more here until more details are ironed out.

TODO discuss option details and parsing, add code to parse and then 
refine anything necessary. After this we can write the rest of the code,
the pattern matching. We also need to decide if more than one name_arg 
can be specified. I'm mixed on this one: with grep -E one can do this so
I think it would be wise to do this but this will be determined at a
later date.


xexyl:
Although MUCH MORE has to be done this feels like tremendous progress! I did forget to mention that the CHANGES.md and the jparse/README.md files were updated as well.

PR #533, opened 2023-06-03
Clarify use of jprint in code and README.md

xexyl:
It's important to note that the use of the tool is not to determine if a
file is valid JSON (though it technically can be used this way) but 
rather to print patterns in a valid JSON file. For now the tool does 
report if it's valid or invalid json but this will be removed. The tool
also for now prints a pattern requested or (if none requested) that 
there is no pattern requested.

Update README.md regarding this too.


lcn2:
Thanks for the progress @xexyl

xexyl:
> Thanks for the progress @xexyl

Very welcome. I hope I can do more tomorrow.

PR #539, opened 2023-06-04
jprint can now read from stdin + free tree

xexyl:
The usual - for stdin now is accepted.

Free tree prior to exiting if an error is not encountered. This probably should be in a separate pull request but it escaped my mind that I already added the other code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,11 +17,13 @@ functionality is limited to:
 - accepting SOME Of the options but not parsing any except `-h`.
 - if not enough args (at least 1 arg - the file to parse) it prints usage string
 (same as `-h` but with error message about not enough args).
-- checks that first arg exists and is a regular file and if so it tries to parse
-it as a JSON file. Depending on status of whether it's valid or json it will
-report an error or report valid json.
+- checks that first arg exists and is a regular file (or `stdin`) and if so it
+tries to parse it as a JSON file. Depending on status of whether it's valid or
+json it will report an error or report valid json. Later it will only report
+valid JSON as a debug message.
 - parsing options except that only the basic option arguments are parsed.
 - note that `jprint` will make use of the `regex(3)` library.
+- frees the parse tree prior to exiting.
 
 Updated [jparse/README.md](jparse/README.md) to note the `jprint` tool.
 
--- a/jparse/jprint.c
+++ b/jparse/jprint.c
@@ -121,7 +121,7 @@ static const char * const usage_msg1 =
     "\t\t\tUse of -g conflicts with -S.\n"
     "\t-c\t\tOnly show count of matches found\n"
     "\n"
-    "\tfile.json\tJSON file to parse\n"
+    "\tfile.json\tJSON file to parse (- indicates stdin)\n"
     "\tname_arg\tJSON element to print\n\n"
     "\tExit codes:\n"
     "\t\t0\tall is OK, file is valid JSON, match(s) found or no name_arg given\n"
@@ -144,6 +144,7 @@ int main(int argc, char **argv)
     extern char *optarg;
     extern int optind;
     struct json *json_tree = NULL;	/* json tree */
+    bool is_stdin = false;		/* reading from stdin */
     bool is_valid = true;		/* if file is valid json */
     bool match_found = false;		/* true if a pattern is specified and there is a match */
     bool pattern_specified = false;	/* true if a pattern was specified */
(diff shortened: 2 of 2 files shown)


lcn2:
What would this eventually do:

```sh
jprint - < /dev/zero
```

or perhaps:

```sh
yes '{' | jprint -
```
Obviously at some point the realloc would fail.  Might be easier to simulate the failure then to try trigger the "too much data" failure by reading from a never-ending data source.

Just curious.

lcn2:
p.s. `make prep` now fails in the "picky" phase, due to this pull request.

xexyl:
> p.s. `make prep` now fails in the "picky" phase, due to this pull request.

I forgot to check that. 

I will fix it as soon as I have the laptop up tomorrow morning. 

Thank you for the note! Sorry about that.

xexyl:
> > p.s. `make prep` now fails in the "picky" phase, due to this pull request.
> 
> I forgot to check that.
> 
> I will fix it as soon as I have the laptop up tomorrow morning.
> 
> Thank you for the note! Sorry about that.

See commit https://github.com/ioccc-src/mkiocccentry/pull/540/commits/f8d7182342c9b63c2f8c007f4268e711456a5d3e.

lcn2:
> > What would this eventually do:
> 
> > 
> 
> > ```shell
> 
> > jprint - < /dev/zero
> 
> > ```
> 
> > 
> 
> > or perhaps:
> 
> > 
> 
> > ```shell
> 
> > yes '{' | jprint -
> 
> > ```
> 
> > 
> 
> > Obviously at some point the realloc would fail. Might be easier to simulate the failure then to try trigger the "too much data" failure by reading from a never-ending data source.
> 
> > 
> 
> > Just curious.
> 
> > 
> 
> > ### UPDATE 0
> 
> > This is a "what happens when the dynamic array can no longer grow" problem as called by say, `read_all()`. It might be as simple as the low level code detects the `realloc()` failure and exits.
> 
> > 
> 
> > Just curious.
> 
> 
> 
> Good question!
> 
> 
> 
> I'm not sure though I don't use `read_all()` explicitly. Rather I use the stream version of the json parser. Of course this uses `read_all()` itself so it might be a problem in that way.

This isn't a problem: just curious as to what happens when the `realloc()`, called by the dynamic array code that in turn is called by `read_all()` that in turn is called by JSON parser code.

You might wish to simulate `realloc()` returning NULL in the dynamic array code: you might want to trigger this condition after a number of previously successful calls growing the dynamic array block that `read_all()` uses.  Use of a static variable to start the count at 0 and an if statement with a count++ to simulate `realloc()` returning NULL after just a few calls.

> What do you suggest is done? This could be said to be a problem for other uses of the function too, right? 

See above.
 
> The only other ones to use it are `jstrencode` and `jstrdecode`. What do you think should or could be done ?

See above.

xexyl:
> > > What would this eventually do:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > jprint - < /dev/zero
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > or perhaps:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > yes '{' | jprint -
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > Obviously at some point the realloc would fail. Might be easier to simulate the failure then to try trigger the "too much data" failure by reading from a never-ending data source.
> > 
> > 
> > > 
> > 
> > 
> > > Just curious.
> > 
> > 
> > > 
> > 
> > 
> > > ### UPDATE 0
> > 
> > 
> > > This is a "what happens when the dynamic array can no longer grow" problem as called by say, `read_all()`. It might be as simple as the low level code detects the `realloc()` failure and exits.
> > 
> > 
> > > 
> > 
> > 
> > > Just curious.
> > 
> > 
> > Good question!
> > I'm not sure though I don't use `read_all()` explicitly. Rather I use the stream version of the json parser. Of course this uses `read_all()` itself so it might be a problem in that way.
> 
> This isn't a problem: just curious as to what happens when the `realloc()`, called by the dynamic array code that in turn is called by `read_all()` that in turn is called by JSON parser code.
> 
> You might wish to simulate `realloc()` returning NULL in the dynamic array code: you might want to trigger this condition after a number of previously successful calls growing the dynamic array block that `read_all()` uses. Use of a static variable to start the count at 0 and an if statement with a count++ to simulate `realloc()` returning NULL after just a few calls.
> 
> > What do you suggest is done? This could be said to be a problem for other uses of the function too, right?
> 
> See above.
> 
> > The only other ones to use it are `jstrencode` and `jstrdecode`. What do you think should or could be done ?
> 
> See above.

Please add this to a todo in the jprint issue thread and link to your comment: https://github.com/ioccc-src/mkiocccentry/pull/539#issuecomment-1577858983. This way I can worry about the tool itself and focus on this problem later on. Right now it's beyond my faculties, this issue or 'issue' as it might or might not be. Thanks!

lcn2:
> > > > What would this eventually do:
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > jprint - < /dev/zero
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > or perhaps:
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > yes '{' | jprint -
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > Obviously at some point the realloc would fail. Might be easier to simulate the failure then to try trigger the "too much data" failure by reading from a never-ending data source.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > Just curious.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > ### UPDATE 0
> 
> > > 
> 
> > > 
> 
> > > > This is a "what happens when the dynamic array can no longer grow" problem as called by say, `read_all()`. It might be as simple as the low level code detects the `realloc()` failure and exits.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > Just curious.
> 
> > > 
> 
> > > 
> 
> > > Good question!
> 
> > > I'm not sure though I don't use `read_all()` explicitly. Rather I use the stream version of the json parser. Of course this uses `read_all()` itself so it might be a problem in that way.
> 
> > 
> 
> > This isn't a problem: just curious as to what happens when the `realloc()`, called by the dynamic array code that in turn is called by `read_all()` that in turn is called by JSON parser code.
> 
> > 
> 
> > You might wish to simulate `realloc()` returning NULL in the dynamic array code: you might want to trigger this condition after a number of previously successful calls growing the dynamic array block that `read_all()` uses. Use of a static variable to start the count at 0 and an if statement with a count++ to simulate `realloc()` returning NULL after just a few calls.
> 
> > 
> 
> > > What do you suggest is done? This could be said to be a problem for other uses of the function too, right?
> 
> > 
> 
> > See above.
> 
> > 
> 
> > > The only other ones to use it are `jstrencode` and `jstrdecode`. What do you think should or could be done ?
> 
> > 
> 
> > See above.
> 
> 
> 
> Please add this to a todo in the jprint issue thread and link to your comment: https://github.com/ioccc-src/mkiocccentry/pull/539#issuecomment-1577858983. This way I can worry about the tool itself and focus on this problem later on. Right now it's beyond my faculties, this issue or 'issue' as it might or might not be. Thanks!

Done to the bottom in the opening comment in the issue.

PR #544, opened 2023-06-05
Add jprint_util.{h,c}

xexyl:
There are currently no functions in the files but there will be in time
when I add routines that will be used by jprint itself.

Update Makefile and run make depend.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,15 @@
 # Major changes to the IOCCC entry toolkit
 
-## Release 1.0.4 2023-06-05
+## Release 1.0.5 2023-06-05
+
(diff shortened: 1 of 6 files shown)


xexyl:
BTW Landon: do you think the parsing option functions also belong in these files ? I was on the fence and another thought I had was another set of files.

What do you think ?

lcn2:
> BTW Landon: do you think the parsing option functions also belong in these files ? I was on the fence and another thought I had was another set of files.

Yes.

xexyl:
> > BTW Landon: do you think the parsing option functions also belong in these files ? I was on the fence and another thought I had was another set of files.
> 
> Yes.

As you know I ended up doing it so it's good we're on the same page.

PR #551, opened 2023-06-06
Remove extraneous newlines in jprint usage message

xexyl:
At the end of the usage_msg strings there should be only one '\n' as it 
will otherwise (print an additional newline which means two extra lines
between each portion of the usage string when only one is needed).


lcn2:
Thank
You
@xexyl

PR #554, opened 2023-06-06
Add code to parse jprint -p option

xexyl:
There currently is no code to check whether name or value or both was 
specified and just like with -t it might prove another way is better 
when I get to the point of working on the actual processing of the json 
to print.


xexyl:
... doing other things now .. I expect to do more tomorrow. Good day!

lcn2:
thanks @xexyl

xexyl:
> thanks @xexyl

You're welcome.

PR #555, opened 2023-06-07
Start parsing of jprint -l, -i, -N

xexyl:
Add missing -l option itself to getopt().

Add two structs, one contained in the other, for number ranges. This is a work in progress and might change in details.

Add beginning to parsing of these options. Right now it zeroes out the structs and returns true as long as optarg != NULL and *optarg != '\0'.

More will be done in a future commit.

### DIFF
--- a/jparse/jprint.c
+++ b/jparse/jprint.c
@@ -161,8 +161,9 @@ int main(int argc, char **argv)
     bool encode_strings = false;	/* -e used */
     bool quote_strings = false;		/* -Q used */
     uintmax_t type = JPRINT_TYPE_SIMPLE;/* -t type used */
-    uintmax_t max_matches = 0;		/* -i count specified - don't show more than this many matches */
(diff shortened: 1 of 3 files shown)


xexyl:
.. really taking a break now ..

xexyl:
I've finished this part but I am running `make prep` first. Will open a new pull request soon .. might be all I do today but we'll see. I think it's great progress either way as it finishes parsing three options in one go. Semantics might or might not change in the future when I start working on the functionality but I think it's at least a good tentative design.

lcn2:
We will respond to the open question marked with 👀, which is our sign for us to look at the comment in more detail.

xexyl:
> We will respond to the open question marked with 👀, which is our sign for us to look at the comment in more detail.

Looks great! Will reply to them later. Hopefully not too long from now. 

Hope you’re sleeping well!

PR #556, opened 2023-06-07
Fix print types variable in jprint -p option

xexyl:
In the previous commit I also ended up changing the 'print' variable in
jprint_parse_print_option() to be 'print_type'. Now it's 'print_types' 
as it isn't just one necessarily but up to two.

### DIFF
--- a/jparse/jprint.c
+++ b/jparse/jprint.c
@@ -161,8 +161,9 @@ int main(int argc, char **argv)
     bool encode_strings = false;	/* -e used */
     bool quote_strings = false;		/* -Q used */
     uintmax_t type = JPRINT_TYPE_SIMPLE;/* -t type used */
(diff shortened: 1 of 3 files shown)


xexyl:
... _really_ going to rest now .. though I think probably not for long. If I don't get to rest much I might or might not continue here then or later.

lcn2:
> ... _really_ going to rest now .. though I think probably not for long. If I don't get to rest much I might or might not continue here then or later.

Best wishes on your resy @xexyl

xexyl:
> > ... _really_ going to rest now .. though I think probably not for long. If I don't get to rest much I might or might not continue here then or later.
> 
> Best wishes on your resy @xexyl

Thank you .. I believe I fell asleep at that point thankfully.

PR #557, opened 2023-06-07
Complete parsing number range options -l, -i, -n

xexyl:
Using two structs, one within the other, along with some bools it's now
easy to test if a range is required or if an exact number is required or
if a min (number >=) or max (number <=) is required.

The function that does the parsing is:

    /* jprint_parse_number_range        - parse a number range for options -l, -n, -i
     *
     * given:
     *
     *  option      - option string (e.g. "-l"). Used for error and debug messages.
     *  optarg      - the option argument
     *  number      - pointer to struct number
     *
     * Returns true if successfully parsed.
     *
     * The following rules apply:
     *
     * (0) an exact number is a number optional arg by itself e.g. -l 5 or -l5.
     * (1) an inclusive range is <min>:<max> e.g. -l 5:10
     *     (1a) the minimum must be <= the max 
     * (2) a minimum number, that is num >= minimum, is <num>:
     * (3) a maximum number, that is num <= maximum, is :<num>
     * (4) anything else is an error
     *
     * See also the structs jprint_number_range and jprint_number in jprint_util.h
     * for more details.
     *
     * NOTE: currently (as of 7 June 2023) the numbers are signed. This might or
     * might not change depending on what is needed.
     *  
     * NOTE: this function does not return on syntax error or NULL number.
     */ 
    bool
    jprint_parse_number_range(const char *option, char *optarg, struct jprint_number *number);
     
Whether the numbers should be negative is TBD later. I don't see the use
of it but there was something in the original concept that suggested it
might be good to have them signed instead of unsigned.

Debug levels will change in time.

Comments might be improved.

### DIFF
--- a/jparse/jprint.c
+++ b/jparse/jprint.c
@@ -161,8 +161,9 @@ int main(int argc, char **argv)
     bool encode_strings = false;	/* -e used */
     bool quote_strings = false;		/* -Q used */
     uintmax_t type = JPRINT_TYPE_SIMPLE;/* -t type used */
-    uintmax_t max_matches = 0;		/* -i count specified - don't show more than this many matches */
-    uintmax_t min_matches = 0;		/* -N count specified - minimum matches required */
+    struct jprint_number jprint_max_matches = { 0 }; /* -i count specified */
+    struct jprint_number jprint_min_matches = { 0 }; /* -N count specified */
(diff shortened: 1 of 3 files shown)


lcn2:
-thanks @xexyl

xexyl:
> -thanks @xexyl

Wait .. is that a thank you or not ? 😃 

(you’re very welcome!)

PR #559, opened 2023-06-07
Finish parsing jprint options (-b, -I)

xexyl:
Since '\t' is 8 spaces and since the other mode for the options is 
number of spaces if optarg is "t" or "tab" set the variable to 8.
Otherwise attempt to parse as a number. If it fails to parse it's an
error.

This completes at least for now the parsing of all options if I'm not
very much mistaken. As long as I did not miss any in the long list this
is done. Once I start adding functionality (the purpose of the tool
itself) some of the options might very well change.

### DIFF
--- a/jparse/jprint.c
+++ b/jparse/jprint.c
@@ -161,8 +161,9 @@ int main(int argc, char **argv)
     bool encode_strings = false;	/* -e used */
     bool quote_strings = false;		/* -Q used */
     uintmax_t type = JPRINT_TYPE_SIMPLE;/* -t type used */
-    uintmax_t max_matches = 0;		/* -i count specified - don't show more than this many matches */
-    uintmax_t min_matches = 0;		/* -N count specified - minimum matches required */
+    struct jprint_number jprint_max_matches = { 0 }; /* -i count specified */
+    struct jprint_number jprint_min_matches = { 0 }; /* -N count specified */
+    struct jprint_number jprint_levels = { 0 }; /* -l level specified */
     uintmax_t print_type = JPRINT_PRINT_VALUE;	/* -p type specified */
     uintmax_t num_spaces = 0;		/* -b specified */
     bool print_json_levels = false;	/* -L specified */
@@ -182,7 +183,7 @@ int main(int argc, char **argv)
      * parse args
      */
     program = argv[0];
-    while ((i = getopt(argc, argv, ":hVv:J:eQt:qj:i:N:p:b:LTCBI:jEISgcm:")) != -1) {
+    while ((i = getopt(argc, argv, ":hVv:J:l:eQt:qj:i:N:p:b:LTCBI:jEISgcm:")) != -1) {
 	switch (i) {
 	case 'h':		/* -h - print help to stderr and exit 0 */
 	    usage(2, program, "");	/*ooo*/
@@ -205,6 +206,9 @@ int main(int argc, char **argv)
 	     */
 	    json_verbosity_level = parse_verbosity(program, optarg);
 	    break;
+	case 'l':
+	    jprint_parse_number_range(optarg, &jprint_levels);
+	    break;
 	case 'e':
 	    encode_strings = true;
 	    break;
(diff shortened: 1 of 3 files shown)


lcn2:
`-b tab` should cause the code to emit a "\t" instead of spaces.  So should `-b 8` generate 8 spaces instead?

xexyl:
> `-b tab` should cause the code to emit a "\t" instead of spaces.  So should `-b 8` generate 8 spaces instead?

I actually thought if that. There's an argument for each. `\t` seems to be 8 but it technically is a different character. 

But then at the same time having to not have both options simplifies the code. 

The real question is do you want me to add additional code so that it can use tab character instead of eight spaces? 

A third but perhaps not a very good option is if the count is 8 use `\t`. 

Which do you prefer? 

I will be happy to do whatever is needed. 

I will reply to everything else tomorrow morning. 

I am going to be helping my auntie learn how to use Find My as she's going on holiday and last time someone stole her watch so she would like to know how to use it for all her devices just in case. 

After that I am going to bed. 

Anyway let me know what you prefer here and I will do it!

Good night!

lcn2:
> > `-b tab` should cause the code to emit a "\t" instead of spaces.  So should `-b 8` generate 8 spaces instead?
> 
> I actually thought if that. There's an argument for each. `\t` seems to be 8 but it technically is a different character.
> 
> But then at the same time having to not have both options simplifies the code.
> 
> The real question is do you want me to add additional code so that it can use tab character instead of eight spaces?
> 
> A third but perhaps not a very good option is if the count is 8 use `\t`.
> 
> Which do you prefer?
> 
> I will be happy to do whatever is needed.
> 
> I will reply to everything else tomorrow morning.
> 
> I am going to be helping my auntie learn how to use Find My as she's going on holiday and last time someone stole her watch so she would like to know how to use it for all her devices just in case.
> 
> After that I am going to bed.
> 
> Anyway let me know what you prefer here and I will do it!
> 
> Good night!

What about using:

```
    -b number[t]    print spaces or tabs between name and value (def: 1 space)
                              NOTE: print of spaces, or if it ends in t (i.e., 1t) print tabs
                              NOTE: With -T, effects whitespace between name and : and between : and value
                              NOTE: Requires use of -p b or -p both

    -I number[t]    When printing JSON syntax, indent levels with spaces or tabs (def: do not indent)
                              NOTE: print of spaces, or if it ends in t (i.e., 1t) print tabs
                              NOTE: Requires use of -j
```

This would allow one to use `-b 2t` or `-I 2t` to refer to 2 tabs.   There would be no `-b tab` nor `-I tab`.

And the use of `-b number[t]` with `-p b` (pr `-b both`) would be mandatory.
And the use of `-I number[t]` with `-j` would be mandatory.

And the "_number_" can be >= 0.

xexyl:
> > > `-b tab` should cause the code to emit a "\t" instead of spaces.  So should `-b 8` generate 8 spaces instead?
> > 
> > 
> > I actually thought if that. There's an argument for each. `\t` seems to be 8 but it technically is a different character.
> > But then at the same time having to not have both options simplifies the code.
> > The real question is do you want me to add additional code so that it can use tab character instead of eight spaces?
> > A third but perhaps not a very good option is if the count is 8 use `\t`.
> > Which do you prefer?
> > I will be happy to do whatever is needed.
> > I will reply to everything else tomorrow morning.
> > I am going to be helping my auntie learn how to use Find My as she's going on holiday and last time someone stole her watch so she would like to know how to use it for all her devices just in case.
> > After that I am going to bed.
> > Anyway let me know what you prefer here and I will do it!
> > Good night!
> 
> What about using:
> 
> ```
>     -b number[t]    print spaces or tabs between name and value (def: 1 space)
>                               NOTE: print of spaces, or if it ends in t (i.e., 1t) print tabs
>                               NOTE: With -T, effects whitespace between name and : and between : and value
>                               NOTE: Requires use of -p b or -p both
> 
>     -I number[t]    When printing JSON syntax, indent levels with spaces or tabs (def: do not indent)
>                               NOTE: print of spaces, or if it ends in t (i.e., 1t) print tabs
>                               NOTE: Requires use of -j
> ```
> 
> This would allow one to use `-b 2t` or `-I 2t` to refer to 2 tabs. There would be no `-b tab` nor `-I tab`.
> 
> And the use of `-b number[t]` with `-p b` (pr `-b both`) would be mandatory. And the use of `-I number[t]` with `-j` would be mandatory.
> 
> And the "_number_" can be >= 0.

I will look at this later today but if I don’t PLEASE remind me! I think you might have referred to it in the other thread and if so I will definitely react to it. 

Thank you for the ideas!

PR #563, opened 2023-06-08
New jprint version - all option arg checks added

xexyl:
New release for mkiocccentry toolkit: 0.0.7 2023-06-08. Great progress
has been made with jprint at its new version 0.0.7 2023-06-08!

At this time I believe all known checks for jprint options have been
added!

The next step is to write some test functions (likely printing debug
messages for different options and their option arguments).

If all is OK the code to traverse the tree to look for simple matches (this does
not mean the JPRINT_TYPE_SIMPLE) can be added. At first the tool will not
check for the constraints but rather just print the name and value (even though
the default is value (JPRINT_PRINT_VALUE) only I want to make sure that the
ability to to print both is there).  This will help make sure that the
traversing works okay before constraints are added. Prior to the following
step, described below, the grep-like functionality, using regex.h, can be
added (it might be better to instead add the grep-like functionality after the
below - the constraints - are added but this will be determined at the time).

Once the above is okay the constraints can be added. The tests should be easier
to do than the traversing and following parts but my hope is that in the coming
days more than tests can be added.

Once this is all done a comprehensive test script can be added to the repo that
is called by the entire test suite (including bug_report.sh) so that we can
verify that jprint works as expected. Then any issues can be fixed and the
tests can be run again. This might take more than a few days but hopefully
things will move along nicely.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@ what the limit is, indicating that 0 is no limit and 256 is the default.
 bitvector but this might change as more is developed (bitvector was not the
 first thought).
 
+New `jprint` version "0.0.6 2023-06-07". It now parses all options and most test
+functions for options being set are added as well. This version is backdated to
+7 June because this was done on the 7th of June but the version was not updated.
 
 ## Release 1.0.4 2023-06-04
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,34 @@
 # Major changes to the IOCCC entry toolkit
 
+
+## Release 1.0.5 2023-06-08
+
+`jprint` version now "0.0.7 2023-06-08". At this time I (Cody) believe all known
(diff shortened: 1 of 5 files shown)


lcn2:
Good version update @xexyl

xexyl:
> Good version update @xexyl

Thanks. It’s sometimes a difficult question to answer whether or not to update the version but I think at this point it is okay. 

When I fix the bit testing (which strangely is only showing up for some bits even though it’s the same code) I will update the version again. 

I won’t look at fixing it until after I reply to all the many good comments. That will take a while I am afraid. I am away from the laptop atm but I hope when I am back there I can continue on the comments of the issue itself. 

I slept in today: woke up at 3 but I am feeling quite fatigued for the moment. Having the fan on would help but it will make my hands and feet too cold so I will have to deal with it.

PR #564, opened 2023-06-08
New jprint version - initial test code added

xexyl:
Add option -K which runs test functions and exits 0 if successful and 1
otherwise. This currently conflicts with the exit codes but this will be
fixed in a future commit.

Number range functions added and everything runs fine.

Next is the bit options but this will take a bit more time and
must be done later as I have to head off for a bit. I hope to do it
today also but this is uncertain.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,11 +3,14 @@
 
 ## Release 1.0.5 2023-06-08
 
-`jprint` version now "0.0.7 2023-06-08". At this time I (Cody) believe all known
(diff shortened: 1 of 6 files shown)


lcn2:
Oh, K! @xexyl

xexyl:
> Oh, K! @xexyl

Thank goodness I didn’t make the option `-J`: I utterly hate citrus and you might have said ‘Oh, J’!

Which I believe is sometimes used for orange juice 🤮

PR #566, opened 2023-06-08
Fix usage strings in jprint

xexyl:
By accident I had put the -K option description in the middle of the -m
and note about that option.

Also it doesn't test the options parsed but rather constraints whether 
they were passed prior to the -K or not (even if any were passed it 
won't use those) so the description was fixed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,11 +3,14 @@
 
 ## Release 1.0.5 2023-06-08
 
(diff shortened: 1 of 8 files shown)


lcn2:
Rather than "JSON_INFINITE_DEPTH: use this with extreme caution\n.", how about:

"JSON_INFINITE_DEPTH. Warning this option may\n"
"\t\t\t     unsafe even if you have infinite about of memory. :-)\n"

PR #568, opened 2023-06-08
Use strdup() on optarg in jprint -t parsing

xexyl:
This is for the test functionality so that one can pass in say 
"num,bool,str" to the function rather than rely on the command line 
itself (which does not need strdup()).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,10 +3,10 @@
 
 ## Release 1.0.5 2023-06-08
 
-`jprint` version "0.0.8 2023-06-08". At this time I (Cody) believe all known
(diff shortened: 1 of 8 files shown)


xexyl:
... okay I've made great progress but I'm seeing a really odd problem in the testing of the bits and since I've done a lot today I'll stop for now or for the day and continue tomorrow. I'm sure once I have a break and look at it again things will be clearer.

So off to do other things now. Good day!

lcn2:
Good progress @xexyl

xexyl:
> Good progress @xexyl

Thank you!

PR #591, opened 2023-06-10
Reword a comment in jprint_run_tests()

xexyl:
In this case trying to make it a bit clearer how the number range of 
<min>:-<max> works.


xexyl:
... unfortunately I have to do other things now .. simply too tired. I'll do more tomorrow if I don't do something later on today.

PR #594, opened 2023-06-10
Fix exit codes in jprint_util.c

xexyl:


lcn2:
So much progress @xexyl wow.  We will approve the set of 26 pull requests so that when we reply to 👀-marked comments we will do so from the result of these pull requests.

xexyl:
> So much progress @xexyl wow. 

Thank you very much!

> We will approve the set of 26 pull requests so that when we reply to 👀-marked comments we will do so from the result of these pull requests.

.. which I see you've done now, thanks! I'll look at them in a bit.

Issue #595, opened 2023-06-11
Bug report: jparse column on invalid token is incorrect


xexyl:
## Describe the bug

When an invalid token is found in the json parser the line number shown is correct in the error message but the column is not unless it's actually at column 1.

## To Reproduce

Steps to reproduce the behaviour (adjust steps as needed):

0. Create an invalid json file (or use stdin) and run the json parser on it.
1. Observe the error message and compare to what it should be.

See the relevant output section for details.

## Expected behaviour

The column in this case should be 13, not 1. The line is correct.

## Relevant Output

```sh
$ ./jparse -
{"test":false,
"foo":"bar"},
syntax error node type JTYPE_OBJECT in file - at line 2 column 1: <,>
ERROR[1]: ./jparse: invalid JSON
```

Notice that it thinks the offending `,` is at column 1 even though it's actually at column 13!

## bug report attachment

To save time and because it's not relevant I'm not doing this. All the relevant output is above and the `bug_report.sh` will not show anything wrong here. I know because I wrote it (well that and it doesn't detect expected output just that bad json files are invalid .. maybe it should like my `txzchk_test.sh` script ?) :-)

## Please provide the following information:

 - Did you attach the above requested bug report file?  [yes, no & why not]
As above no for the above noted reason: to save time (I have a lot of other things to do with this repo!) and because it's not going to help in this case.

 - What is the airspeed Velocity of an Unladen Swallow? [n/a, :-), X m/s, unknown, etc.]

42.

## Additional context

Please add any other context about the problem here.

Feel free to use [markdown](https://www.markdownguide.org/getting-started/) in your issue report.

## Thank you, sorry (tm Canada :-) ), next steps

> We are sorry (tm Canada :-)) that you seem to have encountered an problem or issue that prompted
you to file this issue.  We **DO** appreciate your taking the time to inform us.

> **Please be patent!**  We work as volunteers and have a number of other pressing matters.
While we are often **very busy** we **WILL** look at your issue.

No problem: it's my fault anyway! :-) :-) :-)

Please assign it to me and I'll look at it in the coming days: trying to fit it in with the `jprint` tool that I'm also working on plus other things unrelated to this and the other repo.

xexyl:
I _might_ have made progress but I'm not sure. I got it to show column 9 now.

```sh
$ ./jparse bar.json
syntax error node type JTYPE_OBJECT in file bar.json at line 2 column 9: <,>
[...]

PR #597, opened 2023-06-11
Fix jparse locations - resolve issue #595

xexyl:
New jparse version:

The problem was that the column in invalid tokens appeared to be always 1 even when it should have been something else.

For instance given the following input:

    $ ./jparse -
    {"test":false,
    "foo":"bar"},

one would get:

    syntax error node type JTYPE_OBJECT in file - at line 2 column 1: <,>
    ERROR[1]: ./jparse: invalid JSON

Following the questionable spec that one cannot end the last element with a comma, the ',' on line 2 which is at column 13 is in error since it's the last element of the file. As can be seen it showed column 1 line 2 so the line was correct but the line was not. Now it shows the correct line and column:

    $ ./jparse -
    {"test":false,
    "foo":"bar"},
    syntax error node type JTYPE_OBJECT in file - at line 2 column 13: <,>
    ERROR[1]: ./jparse: invalid JSON

Modifying the file a bit so the input is:

    $ ./jparse -
    {
    "test":false,
    "foo":"bar",
    }

one will get:

    syntax error in file - at line 4 column 1: <}>
    Warning: ./jparse: JSON parse tree is NULL
    ERROR[1]: ./jparse: invalid JSON

A question one might have is why is it line 4 column 1 when the invalid character is actually the final ','? It's because there is more input available and after the ',' on the line:

    "foo":"bar"

it does not expect a '}' as that is an error in syntax.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.8 2023-06-11
+
+Fix `jparse` column location calculations where error messages just showed the
+invalid token at column 1 even when the bad token is not on column 1.
+
+
 ## Release 1.0.7 2023-06-10
 
 Release `jprint` version "0.0.13 2023-06-10".
--- a/jparse/jparse.h
+++ b/jparse/jparse.h
@@ -52,7 +52,7 @@
 /*
  * official jparse version
  */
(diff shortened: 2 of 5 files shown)


lcn2:
nice fix @xexyl

You might want to add a test case to detect any possible regression.

PR #598, opened 2023-06-11
Clarify renaming of remarks.example.md for entries

xexyl:
It's better to first copy the example file to a remarks.md file and
modify that file instead.


lcn2:
thanks @xexyl

PR #599, opened 2023-06-11
Make more explicit usage of no name_arg in jprint

xexyl:
Although no name_arg specified always resulted in an exit code of 0 (if
valid JSON) it was not explicit in the code that it will print the 
entire file. There is now a new boolean (admittedly redundant but it 
makes it clearer) that is set to true if there is no name_arg. The 
boolean is called 'print_entire_file'. This boolean has been added 
explicitly to checks even though this is redundant again just to make it
clearer in the code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.8 2023-06-11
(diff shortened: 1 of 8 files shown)


lcn2:
true :-) @xexyl

PR #602, opened 2023-06-11
Move many jprint options to struct jprint_options

xexyl:
This is useful to help organise things and also because I think it's
going to be necessary to have a callback or possibly callbacks and since
there are many options / settings it will be helpful to have everything
in a single struct.

Not every variable is in the struct - only those specific to options. I
am still not completely satisfied with some of the variable names but 
this can be worried about another time.


lcn2:
thanks @xexyl

PR #605, opened 2023-06-12
Fix check of file arg '-' for stdin in jprint

xexyl:
The check was

    if (*argv[0] == '-') { /* ... */ }

but this is incorrect because image if someone does:

    ./jprint -555

Now the entire argv[0] is checked via strcmp(). Why argv[0]? Because 
after options are parsed argv and argc are shifted.

Interestingly this was discovered as I was working on a (for now much
like jparse_test.sh) test script with the funny error in the log: 
 
     ... in file -555 ...

which is obviously very confusing and bogus. The test script has another
issue to be worked out but this is more important to get in as the tool
is not yet complete anyway.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.9 2023-06-12
+
+Make `jprint -h` exit codes formatting consistent with `jparse`.
+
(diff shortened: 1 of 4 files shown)


xexyl:
.. closing this, will reopen in a moment ..

Part of the log is wrong and confusing. I couldn't figure it out in my head at the time but I just did.

PR #608, opened 2023-06-12
Typo fix in jprint.c + update dbg level

xexyl:
For the string "valid JSON" it goes without saying that the JSON is 
valid at that point so there's no need to report it normally if at all.
For now it's being printed at DBG_MED.


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

You're most welcome!

PR #614, opened 2023-06-13
Make a '\t' 8 spaces for error reporting in jparse

xexyl:
Is 8 spaces correct? I'm not sure in every case but it appears to be the case for C itself although I've not verified it in the standard.

New version for json parser and jparse itself. Updated CHANGES.md.

This fixes a problem in some of the test json files where the column was wrongly reported. The jparse_test.sh script will be updated hopefully in the not too distant future to be how I wrote the txzchk script - verify the exact output of the error messages to make sure that the errors are correct including where applicable the columns. There might be a new subdirectory for this so that we can not worry about the others but this will be determined later. Either way this does fix one file column width being wrongly reported.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.10 2023-06-13
+
+New `jparse` version at "1.0.3 2023-06-12" and json parser version at "1.0.4
+2023-06-12". The json parser version was bumped by .2 rather than .1 since it
+should have been updated the other day when `jparse` was updated and it might be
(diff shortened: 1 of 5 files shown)


xexyl:
> Yes, by default a tab stop is every 8 columns.
> 
Great! I had the thought of do we need to worry about vertical tabs or backspace. I'll look into this. I just am waking up from a nap. Sorry you were awake.


> Thanks @xexyl

Welcome.

xexyl:
.. and the answer to that question is no. The `\b` is detected (of course) by the low byte test prior to a full attempt at reading and the vtab is reported correctly too.

That being said they're good test cases so I'll probably add them to the repo anyway.

PR #615, opened 2023-06-13
Add jparse location error testing capability

xexyl:
The jparse_test.sh script (new version is 1.0.2 2023-06-13) now acts on a new subdirectory in jparse/test_jparse/test_JSON called 'bad_loc'. This is to greatly simplify the other test function and to isolate the specific tests to the specific files.

It works very much like how I made txzchk_test.sh: it needs an error file for each json file that must fail with the exact error message. This means that if we change the error messages in jparse or a new file is added one must regenerate the error files. In the jparse/test_jparse directory is a new README.md file with the details. This information is also in the man page jparse_test(8). But to have it in the log I include the exact test in the man page below.

Note that more files need to be added. Right now there is only one as I wanted to just get the functionality in and be able to add more files later as time goes on.

   Location errors
	The purpose of this is to verify that the error locations are
	correctly reported.  The error files have the errors that
	jparse(1) MUST EXACTLY report.

	Whenever a JSON file is added that has a syntax error that is
	needed to test that location reporting in the JSON parser error
	messages is correct, one must do the following things.

	First change to the test subdirectory.  Assuming you're at the
	top level directory:

	    cd jparse/test_jparse

	Next you must delete all the error files in test_JSON/bad_loc
	like:

	    rm -f test_JSON/bad_loc/*.err

	Then you must generate the error files again like:

            for i in ./test_JSON/bad_loc/*.json; do ../jparse -- "$i" 2>"$i.err" ; done

	Then you must go back to the top level directory and run the
	test suite:

            make test

	Assuming everything is OK you can then add the JSON files and the
	error files like:


xexyl:
Just noticed incomplete git log here .. and of course already did another commit and this one has several files modified. Oh well. The documentation is noted but it's unfortunate this is incomplete. Maybe when I add more test cases I'll add this information again.

xexyl:
Has to do with the subdirectory option ... that could be a problem. Have to think about this.

xexyl:
Fixed it. Will be making a commit soon.

xexyl:
Well it's easy to fix for running under `jparse/` but not under the top level directory.

xexyl:
Going to have to look at this later .. sorry.

xexyl:
I got this to work seemingly but it might not be a good idea to have this after all. Another option is to have a different script that the test suite calls. That would possibly work but it would be I think ugly (though probably not as ugly as what I have now) and I don't know what to call it either.

Any suggestions?

xexyl:
Making another pull request that relies on this one and which will fix the problem so reopening.

PR #616, opened 2023-06-13
-Y added to jprint to search by value

xexyl:
New version for jprint "0.0.16 2023-06-13".

Added -Y type option which will search by value. So why not V? As the
comments say: 

    /*
     * Why is this option -Y? Why not Y? Because Y, that's why! Why
     * besides, all the other good options were already taken. :-)
     * Specifically the letter Y has a V in it and V would have been the
     * obvious choice but both v and V are taken. This is why you'll
     * have to believe us when we tell you that this is a good enough
     * reason why! :-)
     */

This arg will later on only support one name_arg or at least that's 
what's been discussed but at this time no limit is imposed. The parsing
of -Y is very similar to the parsing of -t except it only allows one 
type and it doesn't support all types. Some might change. For instance 
it allows type num which is an alias for different types. This was due 
to it being in the usage string and it's for no other reason.

Debug messages have been updated to show whether it will search by name
or value.

The original thought was to have -Y have no option arg and instead it 
would have to be where you specify two args after the end of options 
(and the file) but this seemed like more complication so I made it 
require the type as an option arg.

More details will be discussed and then more code will be added if 
necessary.


xexyl:
I hope that the problem with the other pull request is not going to cause a problem with this one when it comes to merging the CHANGES.md file but if it does I'll update it after the fact. The other pull request will have to be delayed for now I'm afraid. It's a strange issue - it seemed to be fine this morning but it's not. I think I just need a sleep later on today in order to get a clearer head here.

Leaving now. Might be back in a bit but that's not certain.

PR #617, opened 2023-06-13
Fix testing of jparse error locations

xexyl:
A number of problems occurred that did not show up at first that
resulted in having to make some changes in the way the testing is done.
In jparse/ the test rule will pass the new option -L to jparse_test.sh.
Only when that option is specified will the script run the error 
location tests. This is because the test_ioccc/ioccc_test.sh cannot run
the tests properly nor does it need to run these tests.

Also the way the set up of the error files has to be done is slightly
modified too. This is what has to be done.

First one must generate the error files from the directory
jparse/test_jparse like so:

    for i in test_JSON/./bad_loc/*.json; do ../jparse -- "$i" 2>"$i.err" ; done

Once this is done you must test it. You should be able to run make test
from the top level directory or from either the jparse or
jparse/test_jparse subdirectories. If everything goes okay you can then
add the new files including the error files like:

    git add ./jparse/test_jparse/test_JSON/bad_loc/*.json
    git add ./jparse/test_jparse/test_JSON/bad_loc/*.err

Then commit, push and make a pull request.

Removed a possibly problematic test file and added another one. Error 
file actually was added this time.

Updated man page with the new option and fixed details of creating the
right error files. The difference has to do with the subdirectory option
of the jparse_test.sh script and the fact the error files have to be 
exact matches.


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

You're very welcome!

... but I noticed a minor problem: if erroneous token is at the first column it thinks it is column 0 .. now I fixed the but then if the first token is a space and the second character is the incorrect character it shows that it's column 1!

I have to work this out too.

xexyl:
Fixed it! Expect a pull request soon.

xexyl:
.. however it now shows line one if first line is empty which means this is not yet fixed .. I have to rest soon so I'll address this later.

xexyl:
.. actually a minor problem but I need to rest. I fixed part of it but it's not complete. I'll fix this later .. I hope.

xexyl:
Couldn't sleep and it looks like I thought of the solution as I was resting. Seems all good. Will be doing a pull request if this is indeed the case. I have to modify the man page but I might do that in another commit. I've added some json files so those will be included as well.

After that I'm going to see if I can simplify the code but since it works now I want to make sure it goes in. If all goes well a pull request will be opened soon.

xexyl:
... went back to sleep by a stroke of luck. I'll be doing the updates to comments and documentation now and then maybe add more files. Once this is done I will run make release to make sure everything is sane and then do a pull request. Once that's done I will see about simplifying the fix a bit and if that's possible I will do another commit but not on another pull request as it's all related.

PR #628, opened 2023-06-15
Add flexibility to jprint regexps/regular patterns

xexyl:
NOTE: to see more rationale behind this see comment https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1593134282.

NOTE: it should not cause problems with substrings unless `-g` is specified. Then it's a command line error. Special care was made so that it can still be used with `-Y` as long as only one name arg is specified, either via `-G` or after the file.json arg. 

NOTE: I have taken special care to avoid problems with option conflicts but if I missed something please let me know and I'll happily fix it!

The way this works is that each pattern now has a boolean to indicate the type: regexp, regular, substring (conflicts with regexp) and also if it should be searching by value or name.

The commit log is below:

--

With -G regexp one can explicitly enable a search for a regexp pattern
without conflicting with the normal patterns (name_args) or -s. But how
does this work so it's not in conflict?

The patterns struct has booleans to indicate what was specified so that
when -G is specified that specific pattern will be added as a regexp but
after the command line if -g was not specified it will be a regular
pattern. If -s is specified -g is still an error. -G can be used with -Y
but the name_arg will be instead the regexp explicitly specified. This
means that the following two commands would result in the same
functionality with respect to -Y:

    jprint .. -Y int -G foo file.json
    jprint .. -Y int -g file.json foo

Without either -G or -g the -Y would search value without regexps
enabled. If however -G foo and file.json foo is used or more than one -G
foo is used or file.json foo foo is used it is an error. This requires a
special additional check in main() to verify this.

In making the above enhancements I discovered and fixed a bug that 
caused the file.json arg to be the wrong argv element depending on 
options specified.

This feature allows more flexibility in looking for patterns.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,15 @@ was `-s`.
 allowed and use of `-A` requires use of `-S`. The difference besides letters is
 that args now must not be empty though this might not be strictly necessary.
 
+Allow for both regular expressions and normal patterns via the `-G regexp`
+option. Does not conflict with `-s` because the regexp is added as a regexp but
+the `name_arg`s will allow substrings if `-s` specified - but not for the `-G`
+regexp. `-G` does not conflict with `-Y` as long as only `-G` is used only once
+and no additional `name_arg` is specified at the command line after `file.json`.
+
+Fix bug where the command line could be messed up and the file.json is thought
+to be a different arg specified.
+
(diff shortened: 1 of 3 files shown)


lcn2:
Gee @xexyl

xexyl:
> Gee @xexyl

HI ! ... JKLMNOP! :-)

PR #630, opened 2023-06-15
Remove no longer used function in jparse/util.[ch]

xexyl:


xexyl:
.. off to do other things again ... will do more tomorrow.

xexyl:
> Good to "get arounded" to this @xexyl

Good one. I'm afraid I'm not awake enough to come up with something clever to say back but it's a fun one: though I would have said 'get rounded'. How about just some technical synonyms and similar for round?

`cycloid` is considered a synonym for round but you probably know it in the mathematical sense. Do you know `discoid`? It's a technical term for shaped like a disc. Rare word for round: `globose`.

There are a number of others too of course! But that will have to do for now :-)

I hope to look at the commit in the other repo soon and maybe modifying `pipe_open()` to allow for writing.

PR #634, opened 2023-06-16
Complete jprint when no pattern specified or -W

xexyl:
When no pattern is specified or the new option -W (for 'write') is
specified to jprint the entire file is printed if valid JSON. If it's
not valid JSON it's an error and the file will not be printed. In the
case it is printed the exit code will be 0.

This change is important because it is one of the key features of the 
tool and it's now complete with a new option to do it explicitly (though 
it would technically conflict with a pattern specified so if a pattern 
is specified the -W option is ignored and maybe it should be a command 
line error?)!

New version is 0.0.19 2023-06-16.

### DIFF
--- a/jparse/jprint.c
+++ b/jparse/jprint.c
@@ -651,6 +651,7 @@ int main(int argc, char **argv)
     /* free tree */
     json_tree_free(json_tree, jprint->max_depth);
 
+    /* All Done!!! -- Jessica Noll, Age 2 */
     if (jprint->match_found || !jprint->pattern_specified || jprint->print_entire_file) {
 	free_jprint(jprint);	/* free jprint struct */
 	jprint = NULL;	/* set jprint to NULL even though we're just about to exit */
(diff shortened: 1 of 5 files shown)


xexyl:
Although simple this change did feel like great progress as it means that one mode of the program is complete!

.. though it does bring up the question of what if `read_all()` fails due to allocation error (I guess more likely out of memory but I don't know: the dynamic array facility is not one I'm very familiar with the internals). The question is what if it does fail? What should be done? What even happens in that case? Should `read_all()` return NULL and then the code exits? I don't have an answer to any of these questions.

What do you think Landon ?

lcn2:
The `read_all()` function in `util.c` returns malloc buffer containing the entire contents of stream, OR NULL is an error occurred.   However such errors are limited to things such as bogus function args OR corrupted dynamic array information.

In regards to a `realloc(3)` failure, however: The call to `realloc(3)` in `dyn_array_grow()` in `dyn_array/dyn_array.c` calls `errp()` and does **NOT** return.

This is because if the memory allocator is failing, say because of a lack of memory or perhaps something has corrupted the allocation space (due to a bogus memory smash), there is little that can be done without an insane amount of memory analysis and recovery (not worth it).

So the answer to what might happen with something like:

```sh
yes '{' | ./jprint -j -
```

When the call to `read_all()` ran out of memory, the process would exit non-zero with a message of the form:

```
... failed to reallocate the dynamic array from a size of XYZZY bytes to a size of PLUGH bytes
```

PR #636, opened 2023-06-16
Change the way pipe_open() handles modes

xexyl:
As pointed out on GitHub as it's only for write or read mode it can be 
misleading to have a char const *mode. Thus it is now a boolean which 
states whether it should be write or read. The boolean is write_mode.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,19 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.13 2023-06-16
+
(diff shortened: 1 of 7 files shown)


lcn2:
wtite(2) on @xexyl 😁

xexyl:
> wtite(2) on @xexyl 😁

I guess that was a typo for `write(2)` but I must confess I was really hoping for something like ...

```sh
$ echo '|' | ./goren 
       |
This is a pipe
```

:-)

PR #637, opened 2023-06-16
jprint now executes -S path if executable

xexyl:
Some important points to note. First it runs it via shell_cmd() to check
the exit code. This ends up printing out the output but since the exit
code was noted as significant it will do for now.

If it's non zero it will be an error. Otherwise it opens a write-only
pipe. If the pipe is not NULL it will be processed at a later time once
I am clear what needs to happen. If the pipe is NULL it is an error.
This means that the output is printed twice.

Second point of interest is that the -A args option I am not entirely
clear on whether they should be options that are ended by '--' or if
they should be args to the program. It might not matter even but I'm not
sure. Nevertheless right now it's used as options which. Even so one can
do for instance:

    $ cat foo.json
    {}

Specifying no pattern to search:

    $ ./jprint -S /usr/bin/grep -A '{' foo.json
    debug[0]: set tool path to: </usr/bin/grep>
    debug[0]: set tool args to: <{>
    debug[0]: maximum depth to traverse: 256 (default)
    {}
    debug[0]: no pattern requested or -o, will print entire file
    {}
    {}

..which prints the full file (because of no pattern specified) but runs
grep searching for '{' in the file too. If one specifies a pattern it'll
look like:

    $ ./jprint -S /usr/bin/grep -A '{' foo.json '{'
    debug[0]: set tool path to: </usr/bin/grep>
    debug[0]: set tool args to: <{>
    debug[0]: maximum depth to traverse: 256 (default)
    debug[0]: name pattern requested: {
    debug[0]: adding name pattern '{' to patterns list
    {}
    debug[0]: searching for name pattern '{' (substrings ignored)
    {}

..printing the last part only once as the tool does not search yet and
no full file printing was requested.

One could replace grep with any other tool as well even a shell script
that exits 1.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,19 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.13 2023-06-16
+
+New `jprint` version "0.0.19 2023-06-16" - will print entire file if no pattern
+specified OR the new option `-o` is specified but only if valid JSON (for both).
+A note that if `read_all()` fails to read in the entire file then there could be
+a problem but we only print the file if the data read is not NULL. This change
+is important as it means now all that needs to be done is add the handling of
+the JSON checker/semantics tool and then the compiling of regexps (if requested)
+and then printing for any matches found (the search routines have to be written
+too as does the code to traverse the tree - or 'climb down the tree' :-) ). In
+other words one of the features of the program is complete!
+
+
+
 ## Release 1.0.12 2023-06-15
 
 Change option letters of `jprint` a bit as described next.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,22 @@ other words one of the features of the program is complete! Note that use of
 Changed `pipe_open()` to allow for write mode. Takes a new boolean `write_mode`
 which says to open for write mode rather than read mode.
 
+Fix handling of `jprint -S` and `jprint -A` (specifically `-S`: a segfault in
+some conditions namely if the path did not exist which was not checked for and
+thus not triggered until this change). Verify path is a regular file and is
+executable. Check args and if specified make sure not empty and make sure that
+`-S` is specified as well. Checks `-S` path first.
+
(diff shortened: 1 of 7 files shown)


xexyl:
Making this a draft .. I'm aware of a couple problems. Will hopefully fix soon. If not I'll have to look at it later.

xexyl:
.. should be fixed now .. though I'm aware that I have to test some commands being used ..

xexyl:
I've converted this back to a draft as I think it might be a very bad idea. You wanted this for arbitrary semantics tools but the problem is if we can specify args (and even if we couldn't there's still something of a risk) one can do very bad things.

What do you want done with this?

xexyl:
Not sure how this got merged as I converted it to a draft. Maybe too late? I thought of it as I was doing something else that delayed me.

I might as well make the other not a draft then though ... but I think this issue should be discussed anyway.

lcn2:
> I've converted this back to a draft as I think it might be a very bad idea. You wanted this for arbitrary semantics tools but the problem is if we can specify args (and even if we couldn't there's still something of a risk) one can do very bad things.

There isn't a risk any more than a compiler allow someone to creat an executable that does stuff, or a `make(1)` allows one to do stuff, or the `bash(1)` shell allows one to do stuff.
 
> What do you want done with this?

Complete the ability of `jprint(1)` to invoke another program prior to scanning JSON for stuff.

xexyl:
> > I've converted this back to a draft as I think it might be a very bad idea. You wanted this for arbitrary semantics tools but the problem is if we can specify args (and even if we couldn't there's still something of a risk) one can do very bad things.
> 
> There isn't a risk any more than a compiler allow someone to creat an executable that does stuff, or a `make(1)` allows one to do stuff, or the `bash(1)` shell allows one to do stuff.
> 
I guess that's true.

> > What do you want done with this?
> 
> Complete the ability of `jprint(1)` to invoke another program prior to scanning JSON for stuff.

Sure .. not today though. Once the last thing is merged I'll do the txzchk version update - or if not today then tomorrow morning.

PR #638, opened 2023-06-16
Minor fixes in txzchk

xexyl:
An invalid check would have prevented a debug log from being shown even
if the debug level requested was high enough.

Also removed the clearing of errno where errno was not needed in two
places.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,19 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.13 2023-06-16
+
+New `jprint` version "0.0.19 2023-06-16" - will print entire file if no pattern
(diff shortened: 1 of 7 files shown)


xexyl:
.. going afk now .. will maybe do more later ..

lcn2:
The version of `txzchk` should be updated and the minor fix should be mentioned in `CHANGES.md` @xexyl

xexyl:
> The version of `txzchk` should be updated and the minor fix should be mentioned in `CHANGES.md` @xexyl

I almost did this but didn't for whatever reason. I can't do it until all the pull requests are merged so I can clean up first. Unfortunately though that involves two I made a draft as I have a concern with the concept. So it might be done today or else tomorrow.

PR #644, opened 2023-06-16
Change free_jprint to take a struct jprint **

xexyl:
This is so that we can set in the calling function the pointer to NULL 
so that each time the function is called we do not have to then also set
it to NULL (though this is not strictly necessary at this time since we
always exit after freeing it could happen in the future that this does
not happen). This does make the call to the function look a bit less 
nice but it's a worthy compromise I think especially as it's safer. It 
does not abort on jprint == NULL or *jprint == NULL: it's just a warning
like the old jprint == NULL (even though it's safe to free a NULL 
pointer it's obviously not safe to dereference a NULL struct jprint **).


xexyl:
... the question is: is this worth putting in the CHANGES.md ?

xexyl:
> > ... the question is: is this worth putting in the CHANGES.md ?
> 
> Probably not.

Noted by not noting it! :-)

PR #645, opened 2023-06-16
Fix handling of jprint -S + modularisation

xexyl:
We cannot explicitly close the file after the parse is done because the
function that does this does it itself. This caused a crash.
         
Also fixed problem in the calling of the shell_cmd() functions without 
use of -A.


xexyl:
I'm covering this to a draft - and that unfortunately includes the modularisation which was added prior to discovering the bug this commit fixed - because of a problem I thought of with the feature as I mentioned in comment https://github.com/ioccc-src/mkiocccentry/pull/637#issuecomment-1594977572.

Please advise what you'd like done.

xexyl:
I am concerned about this feature though for reasons I noted. But since the other one somehow was merged this fix should be added too.

PR #647, opened 2023-06-17
Fix handling of -Y wrt name_arg count check

xexyl:
New jprint version at "0.0.20 2023-06-17".

Fix special handling for -Y option wrt exactly one name arg. The idea behind the check is that if -G is specified then one cannot specify a name_arg. If it is not specified then you must specify one name_arg. If both -G and a name_arg are not specified it is also an error but there was a bug which required that -G was used.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,16 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.14 2023-06-17
+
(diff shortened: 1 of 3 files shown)


xexyl:
Hmm .. there's still a problem with one of the checks .. will have to fix later. Going to try resting again.

xexyl:
The commit I just pushed should fix it .. resting now or trying to.

PR #657, opened 2023-06-18
Note that -Y / -p is somewhat buggy in some cases

xexyl:
In some cases the use of the two options is problematic. This is due to
the way the traversal of the tree is done where the pattern name is 
added to the match list being incorrect. It ends up where one can have 
the wrong things printed.

This might have been a simple error on my part as I believe I initially
had it. This will have to be fixed in a future commit.


xexyl:
.. and indeed I did have it working right. I'll have to figure out what I changed and when. This does mean that it shouldn't be too much of a problem though.

PR #666, opened 2023-06-19
Partly process jprint -Y in matching code

xexyl:
This is incomplete in a number of ways. It does not check for types 
which are aliases for more than one type except number. But as for 
number the code does not check for specific number types like int or 
float. This has to be added later.

As well for strings the way the function is called has to be resolved to
work for strings. A workaround is in place so that values can be found 
as a string type but this also lets names be added. This will be fixed 
later. It is marked with an XXX comment (or two).


lcn2:
Making progress @xexyl 👍

xexyl:
> Making progress @xexyl 👍

Yes indeed! Another day of sleeping in too. That's good also.

xexyl:
And I must be the 'devil': issue #666 was mine, all mine! :-)

PR #673, opened 2023-06-20
Check jprint -t / -Y more for matches

xexyl:
Don't increment the count of matches if the type is not the same. By type this
refers to the `JTYPE_*` enum found in `json_parse.h`. This means the
`jprint_match` struct now has a type.

Add more strict checks for number types based on options.
 
Disable explicit check for strings in match function as the check there is
problematic anyway and this helps in another way until the function itself can
be fixed. Strings are more problematic than others but matching also
currently matches on either name or value when it should only match on 
one or the other depending on the options specified.

Type checking is very much a work in progress.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ matches.
 
 Change debug level 0 messages to debug level 1.
 
+Implement case-insensitive option `-i`.
 
 ## Release 1.0.16 2023-06-19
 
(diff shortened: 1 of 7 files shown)


xexyl:
> Nice progress @xexyl 

Thank you. Waiting for doctor now. I will do more tomorrow. 

Hope you have a good rest of your day! I hope also that the things I put in the other thread will be of value tomorrow.

PR #676, opened 2023-06-21
Fix make prep

xexyl:
The problem was that when I added the boolean is_integer to the struct 
json_number I forgot about the jnum test array so the array was
incomplete.

Also fixed a typo in an earlier comment in the array (in the code to 
generate the file).


lcn2:
Thanks for the fix, @xexyl

PR #677, opened 2023-06-21
jprint -l and -N cannot have max < 0

xexyl:
Only -n number ranges with max < 0 makes sense. -N especially does not
make sense. It is of questionable value whether or not -l should support
it but for now it does not. If it is desired to have it it can be added
another time.

Made sure to update test functions so that jprint -K does not fail so
that make test does not fail.

Verify that make prep works okay as this has failed a number of times
recently.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,18 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.18 2023-06-21
+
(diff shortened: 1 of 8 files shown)


lcn2:
Thanks for excluding negatives where they have negative effect 😁 @xexyl

xexyl:
> Thanks for excluding negatives where they have negative effect 😁 @xexyl

Which would make them positive! :-)

PR #682, opened 2023-06-21
Implement jprint -b option

xexyl:
Note that if one uses -b before -j the -j will override the -b 
previously specified. This probably should be changed so that if -b is 
previously specified it won't be set but for now the order matters.


lcn2:
Nice progress @xexyl

xexyl:
> Nice progress @xexyl

Thank you. I'll do more tomorrow including looking at the discussion. Have a great rest of your day!

PR #684, opened 2023-06-22
Modify behaviour of jprint -c to match grep -c

xexyl:
Instead of showing the prefix of each pattern (name_arg) matched show just a total count of the matches. This was my intended purpose in the first place when I suggested the option.


lcn2:
C @xexyl

PR #686, opened 2023-06-22
Add another good json test file

xexyl:
... with the H2G2 them. Why? Because I believe we should follow the following guideline:

    Don't Panic!

:-)

### DIFF
--- a/jparse/test_jparse/test_JSON/good/h2g2.json
+++ b/jparse/test_jparse/test_JSON/good/h2g2.json
@@ -1,0 +1,38 @@
+{
+    "number":42,
+    "number_as_str":"42",
+    "answer to life, the universe and everything" : 42,
+    "panic":false,
+    "panic":true,
+    "false":true,
+    "true":false,
+    "killers":"Vogons",
+    "Killers":"Vogons",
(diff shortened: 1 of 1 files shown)


xexyl:
... I'll look at doing the man page later on today or at least starting it. I think I might take a break now or soon. As there is a lot to write in the man page I don't know that I'll have the energy to write it all today but we shall see.

lcn2:
43 👍 @xexyl

xexyl:
> 43 👍 @xexyl

Funny fact: you know well that Douglas Adams made a joke with 42 and he even said so on USENET. There were some really funny and clever comments in that thread but did you know that some scientists claim that 42 is really the answer to the universe? See https://www.independent.co.uk/life-style/history/42-the-answer-to-life-the-universe-and-everything-2205734.html and https://www.independent.co.uk/news/yes-the-answer-to-the-universe-really-is-42-1351201.html. They spent years to 'prove' that and even go so far as to suggest (or at least those who wrote the articles suggest?) that Douglas was serious when obviously it was bogus. Well obvious to anyone who can think for themselves.

PR #688, opened 2023-06-22
Implement jprint -P option and modularity

xexyl:
Implement jprint -P. When this option is disabled the spaces specified still
separate the tokens if printing both name and value. Default is 1 space.

When printing both (without json syntax) print on the same line the name and
value.

Added boolean to indicate that -b was used. Currently this is unused but it
will allow not relying on the spaces being not 0 as the default is 1
(this changed from 0 to 1 as it was always supposed to be 1 space by
default). This boolean is not yet needed but it might be and it's a 
clear way to indicate if the option was used or not.

Make matching printing function a bit more modular by introducing a new
function, jprint_print_match(). Note that the concept of patterns is
incorrect and has to be fixed. Discussion is pending. Also note that
more functions will have to be added for printing matches.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,10 @@ is specified and found it is a total of the matches found.
 
 Implement `jprint -E` to such a degree as is currently possible.
 
+Minor improvements to `jprint` usage string including not hard coding in the
+string the default maximum depth to parse, instead using the macro
+`JSON_DEFAULT_MAX_DEPTH`.
+
 
 ## Release 1.0.18 2023-06-21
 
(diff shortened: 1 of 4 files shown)


lcn2:
Progress! @xexyl

xexyl:
> Progress! @xexyl

The option is `-P` but the commit was `+P` for progress!

I think most options are now implemented - at least as far as they can be at this time. It's the concept of the name_args and the traversing the tree that has to be worked out. I think there might be two or three more things that can be done but besides those it's just the tree / name_args and the man page (which as I noted in another comment a few minutes ago is mostly done for now).

PR #689, opened 2023-06-22
Fix C NULL / JSON null ambiguity in print function

xexyl:
When printing a json tree for json type 'null' the print routine showed
NULL. To make it more like JSON and to remove any ambiguity with C's 
NULL I changed it to print 'null'.

Updated jparse and json parser version to 1.0.7 2023-06-22.


xexyl:
... afk now .. not sure if I'll be doing more here today or not. It feels like one of those off days but that could easily change after a break. If nothing else some discussion points have been brought up which is an important thing to do.

PR #691, opened 2023-06-23
Minor fixes in README.md

xexyl:


xexyl:
Going for the night .. will do more tomorrow in some form or another.

lcn2:
FYI, GitHub reported:

jprint_test.c: In function ‘jprint_run_tests’:
  jprint_test.c:308:95: warning: value computed is not used [-Wunused-value]
    308 |            jprint_test_bits(false, bits, __LINE__, jprint_match_simple, "JPRINT_TYPE_SIMPLE") &&
        |                                                                                               ^~
  cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable  jprint.c -c
  jprint.c: In function ‘vjprint_json_search’:
  jprint.c:1379:63: warning: implicit declaration of function ‘strcasestr’; did you mean ‘strcasecmp’? [-Wimplicit-function-declaration]
   1379 |                                         (jprint->ignore_case&&strcasestr(str, pattern->pattern))) {
        |                                                               ^~~~~~~~~~
        |                                                               strcasecmp
  cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable  jprin

xexyl:
> FYI, GitHub reported:
> 
> jprint_test.c: In function ‘jprint_run_tests’: jprint_test.c:308:95: warning: value computed is not used [-Wunused-value] 308 | jprint_test_bits(false, bits, **LINE**, jprint_match_simple, "JPRINT_TYPE_SIMPLE") && | ^~ cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable jprint.c -c jprint.c: In function ‘vjprint_json_search’: jprint.c:1379:63: warning: implicit declaration of function ‘strcasestr’; did you mean ‘strcasecmp’? [-Wimplicit-function-declaration] 1379 | (jprint->ignore_case&&strcasestr(str, pattern->pattern))) { | ^~~~~~~~~~ | strcasecmp cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable jprin

Hmm ... I wonder if it's not standard? According to the man page in fedora:

```
       #define _GNU_SOURCE         /* See feature_test_macros(7) */
       #include <string.h>

```

it seems it needs a feature test macro. Do you think I should use that OR do you think we should make our own version? We do need it for the case-insensitive option either way.

xexyl:
Both warnings should be fixed in two open pull requests I made now and a few minutes ago. Some other fixes made as well. We really ought to compile this in linux more often!

xexyl:
I am going to rest now or try to and then I hope to finish the man page when I'm back. That might or might not happen but I hope to finish it anyway.

lcn2:
> > FYI, GitHub reported:
> 
> > 
> 
> > jprint_test.c: In function ‘jprint_run_tests’: jprint_test.c:308:95: warning: value computed is not used [-Wunused-value] 308 | jprint_test_bits(false, bits, **LINE**, jprint_match_simple, "JPRINT_TYPE_SIMPLE") && | ^~ cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable jprint.c -c jprint.c: In function ‘vjprint_json_search’: jprint.c:1379:63: warning: implicit declaration of function ‘strcasestr’; did you mean ‘strcasecmp’? [-Wimplicit-function-declaration] 1379 | (jprint->ignore_case&&strcasestr(str, pattern->pattern))) { | ^~~~~~~~~~ | strcasecmp cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable jprin
> 
> 
> 
> Hmm ... I wonder if it's not standard? According to the man page in fedora:
> 
> 
> 
> ```
> 
>        #define _GNU_SOURCE         /* See feature_test_macros(7) */
> 
>        #include <string.h>
> 
> 
> 
> ```
> 
> 
> 
> it seems it needs a feature test macro. Do you think I should use that OR do you think we should make our own version? We do need it for the case-insensitive option either way.

We think case-insensitive is common.  Perhaps the problem is the GitHub tool is using "-std=gnu11"?  Is `-std=gnu17` any better for the tool?  Besides one would have to co figure the GitHub tool, if that is even possible, to use it. 

Normally we would suggest a feature test macro use, but as you said we need the case-insensitive functionality anyway.  A feature test failure would require us to "roll our own version of the case-insensitive function which is a pain and might be error prone for little gain.

xexyl:
> > > FYI, GitHub reported:
> > 
> > 
> > > 
> > 
> > 
> > > jprint_test.c: In function ‘jprint_run_tests’: jprint_test.c:308:95: warning: value computed is not used [-Wunused-value] 308 | jprint_test_bits(false, bits, **LINE**, jprint_match_simple, "JPRINT_TYPE_SIMPLE") && | ^~ cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable jprint.c -c jprint.c: In function ‘vjprint_json_search’: jprint.c:1379:63: warning: implicit declaration of function ‘strcasestr’; did you mean ‘strcasecmp’? [-Wimplicit-function-declaration] 1379 | (jprint->ignore_case&&strcasestr(str, pattern->pattern))) { | ^~~~~~~~~~ | strcasecmp cc -std=gnu11 -O3 -g3 -pedantic -Wall -Wextra -Wno-unused-but-set-variable jprin
> > 
> > 
> > Hmm ... I wonder if it's not standard? According to the man page in fedora:
> > ```
> > 
> >        #define _GNU_SOURCE         /* See feature_test_macros(7) */
> > 
> >        #include <string.h>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it seems it needs a feature test macro. Do you think I should use that OR do you think we should make our own version? We do need it for the case-insensitive option either way.
> 
> We think case-insensitive is common. Perhaps the problem is the GitHub tool is using "-std=gnu11"? Is `-std=gnu17` any better for the tool? Besides one would have to co figure the GitHub tool, if that is even possible, to use it.
> 
> Normally we would suggest a feature test macro use, but as you said we need the case-insensitive functionality anyway. A feature test failure would require us to "roll our own version of the case-insensitive function which is a pain and might be error prone for little gain.

I used the feature test macro anyway. Did it solve the problem?

But rolling our own shouldn't be too difficult anyway I would think if it doesn't solve it.

xexyl:
Something like this would work?

```c
char *
strcstr(char *haystack, char *needle)
{
    char *hay, *need;

    if (!haystack || !needle)
	return NULL;

    while (*haystack) {
	need = needle;

	while (*haystack && tolower(*haystack) != tolower(*need))
	    haystack++;

	hay = haystack;

	while (*need && *haystack && tolower(*haystack) == tolower(*need)) {
	    need++;
	    haystack++;
	}

	if (!*need)
	    return hay;
    }

    return NULL;
}

```

lcn2:
> Something like this would work?
> 
> 
> 
> ```c
> 
> char *
> 
> strcstr(char *haystack, char *needle)
> 
> {
> 
>     char *hay, *need;
> 
> 
> 
>     if (!haystack || !needle)
> 
> 	return NULL;
> 
> 
> 
>     while (*haystack) {
> 
> 	need = needle;
> 
> 
> 
> 	while (*haystack && tolower(*haystack) != tolower(*need))
> 
> 	    haystack++;
> 
> 
> 
> 	hay = haystack;
> 
> 
> 
> 	while (*need && *haystack && tolower(*haystack) == tolower(*need)) {
> 
> 	    need++;
> 
> 	    haystack++;
> 
> 	}
> 
> 
> 
> 	if (!*need)
> 
> 	    return hay;
> 
>     }
> 
> 
> 
>     return NULL;
> 
> }
> 
> 
> 
> ```
> 
> 

With comments and suitable formatting it could be added to `util.c` and renamed for the usual C function, provided that the function was only compiled in if the feature was not present. We want the libc function to be used by default, not this implementation if function was missing according to the feature macro test.

xexyl:
> > Something like this would work?
> > ```c
> > char *
> > 
> > strcstr(char *haystack, char *needle)
> > 
> > {
> > 
> >     char *hay, *need;
> > 
> > 
> > 
> >     if (!haystack || !needle)
> > 
> > 	return NULL;
> > 
> > 
> > 
> >     while (*haystack) {
> > 
> > 	need = needle;
> > 
> > 
> > 
> > 	while (*haystack && tolower(*haystack) != tolower(*need))
> > 
> > 	    haystack++;
> > 
> > 
> > 
> > 	hay = haystack;
> > 
> > 
> > 
> > 	while (*need && *haystack && tolower(*haystack) == tolower(*need)) {
> > 
> > 	    need++;
> > 
> > 	    haystack++;
> > 
> > 	}
> > 
> > 
> > 
> > 	if (!*need)
> > 
> > 	    return hay;
> > 
> >     }
> > 
> > 
> > 
> >     return NULL;
> > 
> > }
> > ```
> 
> With comments and suitable formatting it could be added to `util.c` and renamed for the usual C function, provided that the function was only compiled in if the feature was not present. We want the libc function to be used by default, not this implementation if function was missing according to the feature macro test.

I agree the libc function would be better. With the feature test macro I think we're good? By renamed for the usual C function what do you mean? I don't think we can do it `strcasestr` so what did you have in mind? And can you think of a way to check if it's there or not?

lcn2:
> > > Something like this would work?
> > > ```c
> > > char *
> > > 
> > > strcstr(char *haystack, char *needle)
> > > 
> > > {
> > > 
> > >     char *hay, *need;
> > > 
> > > 
> > > 
> > >     if (!haystack || !needle)
> > > 
> > > 	return NULL;
> > > 
> > > 
> > > 
> > >     while (*haystack) {
> > > 
> > > 	need = needle;
> > > 
> > > 
> > > 
> > > 	while (*haystack && tolower(*haystack) != tolower(*need))
> > > 
> > > 	    haystack++;
> > > 
> > > 
> > > 
> > > 	hay = haystack;
> > > 
> > > 
> > > 
> > > 	while (*need && *haystack && tolower(*haystack) == tolower(*need)) {
> > > 
> > > 	    need++;
> > > 
> > > 	    haystack++;
> > > 
> > > 	}
> > > 
> > > 
> > > 
> > > 	if (!*need)
> > > 
> > > 	    return hay;
> > > 
> > >     }
> > > 
> > > 
> > > 
> > >     return NULL;
> > > 
> > > }
> > > ```
> > 
> > 
> > With comments and suitable formatting it could be added to `util.c` and renamed for the usual C function, provided that the function was only compiled in if the feature was not present. We want the libc function to be used by default, not this implementation if function was missing according to the feature macro test.
> 
> I agree the libc function would be better. With the feature test macro I think we're good? By renamed for the usual C function what do you mean? I don't think we can do it `strcasestr` so what did you have in mind? And can you think of a way to check if it's there or not?

Trying to deal with a lack of working `strcasestr(3)` instead of just forcing `_GNU_SOURCE` to be defined is very likely **NOT** worth the effort.

One can say that by forcing `_GNU_SOURCE` to be defined, non-GNU environments are out of luck.  But the effort to try to be "non-GNU inclusive" requires one to take on a bunch of compatibly responsibility. 

To see why / an example if what would be involved, look at what `calc(1)` (see [calc repo](https://github.com/lcn2/calc)) does with compiling "more universal code".

One would need to write some sort of  `have_strcasestr.c` program (e.g., a trivial `main()` function that if successfully compiled, calls `strcasestr(3)` with valid args and checks the return value) when successfully compiled and runs, outputs the contents of `have_strcasestr.h` AND if it fails to compile or does not execute correctly, also outputs the contents of ` have_strcasestr.h`.  The difference between in two to cases of forming `have_strcasestr.h` is that in for former case, a `#define HAVE_STRCASESTR` in put in the file AND in the later case, an `#undef HAVE_STRCASESTR` is put into the file.

So then the Makefile has to have a rule that forms `have_strcasestr.h`, and once that file is formed, files that include it (and which will have the `HAVE_STRCASESTR` be defined or undefined depending on if there is a `strcasestr(3)` function (in `libc`) that works.  Then C code that needs to call `strcasestr(3)` would:

... uses code such as:

```c
#include "have_strcasestr.h"    /* see if we have a working strcasestr(3) function */

...

#if !defined(HAVE_STRCASESTR)
extern char *strcasestr(const char *haystack, const char *needle);
#endif /* ! HAVE_STRCASESTR */
```

plus somewhere one would have, say, a file such as `our_strcasestr.c` that would have something like:

```c
#include "have_strcasestr.h"    /* see if we have a working strcasestr(3) function */

#if !defined(HAVE_STRCASESTR)
char *
strcasestr(const char *haystack, const char *needle)
{
    ... a hand made version of `strcasestr(3)` ...
}
#endif /* ! HAVE_STRCASESTR */
```

Such that IF one cannot compile C code that calls `strcasestr(3)`, then the resulting `our_strcasestr.o` contains what is needed to simulate it.

See [calc repo have_strdup.c source](https://github.com/lcn2/calc/blob/master/have_strdup.c) AND the [calc repo Makefile source](https://github.com/lcn2/calc/blob/master/Makefile) (the `HAVE_STRDUP` Makefile define AND the `have_strdup.h` Makefile rule) AND the [calc repo config.c source](https://github.com/lcn2/calc/blob/master/config.c) (the use of `#include "have_strdup.h"` and the use of `#if !defined(HAVE_STRDUP)`), etc.

Doing all that is a bit tricky to do correctly.  Moreover compiling gets delayed as the "have_foo.h" have to be built from attempts to compile "have_foo.c" (where the compile attempts have to be hidden from the user so as to not alarm them with compile failures).  One can no longer compile in parallel as these "have_foo.h" files force a serialization point that must be completed before other code can even be compiled.

etc. etc. etc.

Our point is doing all that in the name of trying to be "compile environment incisive" is a lot of work for almost nil gain.

The [calc repo](https://github.com/lcn2/calc] had to do thing like this because way back in 1984 and earlier, one had to do all kinds of things in order to try an cross-compile on lots of platforms (including a bunch of non-UNIX platforms).  Maintaining such stuff and trying to remain backward compatible requires effort for little gain.

So, even though forcing `_GNU_SOURCE` to be defined is a hack, the alternate can be a lot worse.

**DO NOT** worry if you don't understand all of the details in that calc example / in the comments above.  Just be glad you don't have to do it.  :-)

xexyl:
> > > > Something like this would work?
> > > > ```c
> > > > char *
> > > > 
> > > > strcstr(char *haystack, char *needle)
> > > > 
> > > > {
> > > > 
> > > >     char *hay, *need;
> > > > 
> > > > 
> > > > 
> > > >     if (!haystack || !needle)
> > > > 
> > > > 	return NULL;
> > > > 
> > > > 
> > > > 
> > > >     while (*haystack) {
> > > > 
> > > > 	need = needle;
> > > > 
> > > > 
> > > > 
> > > > 	while (*haystack && tolower(*haystack) != tolower(*need))
> > > > 
> > > > 	    haystack++;
> > > > 
> > > > 
> > > > 
> > > > 	hay = haystack;
> > > > 
> > > > 
> > > > 
> > > > 	while (*need && *haystack && tolower(*haystack) == tolower(*need)) {
> > > > 
> > > > 	    need++;
> > > > 
> > > > 	    haystack++;
> > > > 
> > > > 	}
> > > > 
> > > > 
> > > > 
> > > > 	if (!*need)
> > > > 
> > > > 	    return hay;
> > > > 
> > > >     }
> > > > 
> > > > 
> > > > 
> > > >     return NULL;
> > > > 
> > > > }
> > > > ```
> > > 
> > > 
> > > With comments and suitable formatting it could be added to `util.c` and renamed for the usual C function, provided that the function was only compiled in if the feature was not present. We want the libc function to be used by default, not this implementation if function was missing according to the feature macro test.
> > 
> > 
> > I agree the libc function would be better. With the feature test macro I think we're good? By renamed for the usual C function what do you mean? I don't think we can do it `strcasestr` so what did you have in mind? And can you think of a way to check if it's there or not?
> 
> Trying to deal with a lack of working `strcasestr(3)` instead of just forcing `_GNU_SOURCE` to be defined is very likely **NOT** worth the effort.

Probably true.

> 
> One can say that by forcing `_GNU_SOURCE` to be defined, non-GNU environments are out of luck. But the effort to try to be "non-GNU inclusive" requires one to take on a bunch of compatibly responsibility.
> 

Plus it appears that BSD systems might not even need it: certainly macOS does not. I don't know if all linux systems need it. And even without it in linux systems that need it it seems to work anyway just with a warning.

> To see why / an example if what would be involved, look at what `calc(1)` (see [calc repo](https://github.com/lcn2/calc)) does with compiling "more universal code".
> 
This might be useful in other ways too. I mean for me to see that is.

> One would need to write some sort of `have_strcasestr.c` program (e.g., a trivial `main()` function that if successfully compiled, calls `strcasestr(3)` with valid args and checks the return value) when successfully compiled and runs, outputs the contents of `have_strcasestr.h` AND if it fails to compile or does not execute correctly, also outputs the contents of ` have_strcasestr.h`. The difference between in two to cases of forming `have_strcasestr.h` is that in for former case, a `#define HAVE_STRCASESTR` in put in the file AND in the later case, an `#undef HAVE_STRCASESTR` is put into the file.
> 
As I said elsewhere, I think below, I'm not sure why the complications but it doesn't really matter since we do have the feature-test macro. It's interesting though!

> So then the Makefile has to have a rule that forms `have_strcasestr.h`, and once that file is formed, files that include it (and which will have the `HAVE_STRCASESTR` be defined or undefined depending on if there is a `strcasestr(3)` function (in `libc`) that works. Then C code that needs to call `strcasestr(3)` would:
> 
> ... uses code such as:
> 
> ```c
> #include "have_strcasestr.h"    /* see if we have a working strcasestr(3) function */
> 
> ...
> 
> #if !defined(HAVE_STRCASESTR)
> extern char *strcasestr(const char *haystack, const char *needle);
> #endif /* ! HAVE_STRCASESTR */
> ```
> 
> plus somewhere one would have, say, a file such as `our_strcasestr.c` that would have something like:
> 
> ```c
> #include "have_strcasestr.h"    /* see if we have a working strcasestr(3) function */
> 
> #if !defined(HAVE_STRCASESTR)
> char *
> strcasestr(const char *haystack, const char *needle)
> {
>     ... a hand made version of `strcasestr(3)` ...
> }
> #endif /* ! HAVE_STRCASESTR */
> ```
> 
> Such that IF one cannot compile C code that calls `strcasestr(3)`, then the resulting `our_strcasestr.o` contains what is needed to simulate it.
> 
Very interesting concepts either way - thanks!


> See [calc repo have_strdup.c source](https://github.com/lcn2/calc/blob/master/have_strdup.c) AND the [calc repo Makefile source](https://github.com/lcn2/calc/blob/master/Makefile) (the `HAVE_STRDUP` Makefile define AND the `have_strdup.h` Makefile rule) AND the [calc repo config.c source](https://github.com/lcn2/calc/blob/master/config.c) (the use of `#include "have_strdup.h"` and the use of `#if !defined(HAVE_STRDUP)`), etc.
> 
> Doing all that is a bit tricky to do correctly. Moreover compiling gets delayed as the "have_foo.h" have to be built from attempts to compile "have_foo.c" (where the compile attempts have to be hidden from the user so as to not alarm them with compile failures). One can no longer compile in parallel as these "have_foo.h" files force a serialization point that must be completed before other code can even be compiled.
> 
> etc. etc. etc.
> 
> Our point is doing all that in the name of trying to be "compile environment incisive" is a lot of work for almost nil gain.
> 

Although I don't see why we'd have to if we just use the function in place: kind of like the `base_name()` and other functions instead of the libc `basename()`.

> The [calc repo]([https://github.com/lcn2/calc]](https://github.com/lcn2/calc%5D) had to do thing like this because way back in 1984 and earlier, one had to do all kinds of things in order to try an cross-compile on lots of platforms (including a bunch of non-UNIX platforms). Maintaining such stuff and trying to remain backward compatible requires effort for little gain.
> 
Ah .. 1984. I see.

> So, even though forcing `_GNU_SOURCE` to be defined is a hack, the alternate can be a lot worse.
> 
> **DO NOT** worry if you don't understand all of the details in that calc example / in the comments above. Just be glad you don't have to do it. :-)
> 
Given that there are more important things to worry about I'm not going to try - at least not now .. I am curious with the details just because I love seeing new ideas in C (and some other things too of course)!

It all reminds me, I must say, of `have_timegm` that this repo used to have.

> ## UPDATE 0
> Yes, there are alternative solutions such as using "GNU autoconf" that could help. But down that road likes frustration, people who will file bug reports because the `configure.sh` produces almost unreadable Makefiles, genuine "GNU autoconf" bugs for some versions in some environments, and difficulty in trying to get the system to do what you want, and difficulty in trying figure out what one you want isn't what you get, and difficulty in trying understand why someone else fails to build your code on their system, etc. ...


Yes I wouldn't want to even consider `autoconf`!

PR #694, opened 2023-06-23
Fix implicit declaration of strcasestr()

xexyl:
Add feature test macro _GNU_SOURCE and string.h inclusion to resolve 
implicit declaration of function strcasestr(). This is needed for some 
systems.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.20 2023-06-23
+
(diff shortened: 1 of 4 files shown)


xexyl:
Did the feature test macro and the include (amazed I missed it) solve the problem btw ? It did with linux anyway.

lcn2:
> Did the feature test macro and the include (amazed I missed it) solve the problem btw ? It did with linux anyway.

Seems to have .. didn't notice it on the GitHub CodeQL output

PR #696, opened 2023-06-23
Fix use after free error in vcmdprintf()

xexyl:


lcn2:
Good catch @xexyl

xexyl:
> Good catch @xexyl

Thanks. In this case it was not really to my credit - it was a compiler warning. The one with the printing the wrong variable was my focusing in on things like typos though.

PR #698, opened 2023-06-23
Improve matching in jprint

xexyl:
This is still a work in progress but a bit of progress has been made 
anyway.

In particular with substring matching:

    $ grep num h2g2.json
    "number":42,
    "number_as_str":"42",

    $ ./jprint -P -s -p both h2g2.json num
    num : number
    num : number_as_str

    $ ./jprint -P -s -p n h2g2.json num
    num
    num

    $ ./jprint -P -s -p v h2g2.json num
    number
    number_as_str

What this means is that what matched is not the same as the pattern 
requested in all cases. Later it will be the actual name and/or value of
the json member but this is not currently possible. Note also that the 
-Y option is not yet working properly which will affect this.

As well the struct jprint_match now has two struct json * nodes: the 
name and value each of which depend upon if searching by value or name 
was requested. Later both will be set (along with the above problem 
being resolved) but for now only one will be non NULL. In any event 
they're not used yet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.20 2023-06-23
+
+New `jprint` version at "0.0.26 2023-06-23".
+
+Add booleans (including renaming one `uintmax_t` to be a boolean and adding a
+new `uintmax_t` in its place) in `struct jprint` to indicate that specific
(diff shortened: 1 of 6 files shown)


xexyl:
I want to address the printing of the full document next as well as get the man page done. I would like to do at least one today but preferably both. However atm it's like this might not happen today. I hope this changes as the day goes on.

PR #702, opened 2023-06-23
Fix match pointer check

xexyl:
When renaming variables in the jprint_match struct checks for two NULL 
pointers were changed to the same pointer checked twice.


lcn2:
Nice progress @xexyl

xexyl:
> Nice progress @xexyl

Thank you! I will do more later today or tomorrow .. not sure which. Best luck with the server upgrades and do stay safe but have a good time with the event!

PR #703, opened 2023-06-23
Fix -Y usage syntax in jprint usage string

xexyl:
It originally was thought to be -Y type:value but it was changed to be -Y type and then the other part was just a name_arg which is an arg to the program and not part of the option arg.


lcn2:
No compiler warnings were seen in the GitHub CodeQL tool compile run.

xexyl:
> No compiler warnings were seen in the GitHub CodeQL tool compile run.

Great!

PR #704, opened 2023-06-23
Remove unmatched .RE in jsemtblgen.8

xexyl:


lcn2:
Good catch @xexyl

xexyl:
> Good catch @xexyl

Well to be fair it was just running `checknr` on it. But that's another thing. That tool exits with zero whether or not there's an issue. I was thinking about modifying the one you put on GitHub and making a pull request so that it can exit non-zero if there are issues. I haven't looked at the source though.

xexyl:
See commit https://github.com/lcn2/checknr/pull/1/commits/0db081823a554ebc72d6b9e5a7a407765b740863 for a change to the tool so that we can more easily find these problems in make test etc.

I installed it this way so that it will run prior to the one from MacPorts so now this problem should not occur - assuming one remembers to run `make test`!

PR #713, opened 2023-06-24
Fix use after free in jprint_sanity_chks()

xexyl:
This caused a segfault if the check tool does not exist. It might be 
better if the char * is not in the struct jprint but since the program 
is aborted at this point the memory will be freed by the OS anyway so 
it's probably okay.


xexyl:
..now I really am going for the day!

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

You are most welcome my friend! Hope you're asleep by now.

PR #715, opened 2023-06-25
Slight formatting updates in jprint usage string

xexyl:
In particular when it comes to aliases the formatting was changed 
slightly to be more consistent and to make it more clear what the 
aliases were for (if only slightly more clear).


lcn2:
We are not really sure if:

```
something for: this,that,other
```

is better or worse than:

```
something for 'this,that,other'
```

We can see both ways for now, and it probably comes down to preference and style.

xexyl:
> We are not really sure if:
> 
> ```
> something for: this,that,other
> ```
> 
> is better or worse than:
> 
> ```
> something for 'this,that,other'
> ```
> 
> We can see both ways for now, and it probably comes down to preference and style.

Probably preference and style yes.

But for this:

```
... alias for 'int,float,exp'
```

it's more consistent with the style of the other lines in the options (specifically in case). As for `'`s I think it's good as it makes it very clear it's that. Also if one were to quote it the shell would not do anything to it and it would work just as well.

PR #719, opened 2023-06-25
Move most code from jprint.c to jprint_util.c

xexyl:
Since most of the functions in jprint.c are utility in nature I have 
moved them to jprint_util.c. This required moving the structs that were
in jprint.h to jprint_util.h which I'm not particularly fond of doing 
but it seems worth doing anyway to keep things a bit more organised. Now
the only functions in jprint.c are main(), jprint_sanity_chks() (see 
below for why) and usage().

The reason for having the jprint_sanity_chks() function in jprint.c and
not jprint_util.c is because it uses usage() which is the same name as 
other tools' usage function so it has to stay in jprint.c. Of course the
function could be renamed jprint_usage() but it feels like it's not so 
much a utility anyway and the sanity checks feels like not much of a 
utility either but rather a core component of the tool.

There are some functions that should probably remain in jprint.c but 
those can be moved back later if that is decided to be the case.


lcn2:
The CodeQL warning ⚠️ is back (or it was there all along and we didn't notice it until now) in `jprint_util.c`:

image

xexyl:
> The CodeQL warning ⚠️ is back (or it was there all along and we didn't notice it until now) in `jprint_util.c`:
> 
> image

Nope ... I just forgot to do the proper set up to prevent it. I'll fix it soon. Thanks for the notice.

xexyl:
Does https://github.com/ioccc-src/mkiocccentry/pull/720/commits/1dd2c8048ae74b35aed6f9833a42ce80e24c1e7f fix it ? It should anyway but if not I'll test it in linux too where I know it was a problem.

lcn2:
> 1dd2c80

It might, however since we can only test it during a GitHub action such as considering a pull request, it is hard to say without such a GitHub action to process.

xexyl:
> > [1dd2c80](https://github.com/ioccc-src/mkiocccentry/commit/1dd2c8048ae74b35aed6f9833a42ce80e24c1e7f)
> 
> It might, however since we can only test it during a GitHub action such as considering a pull request, it is hard to say without such a GitHub action to process.

I'm surprised it didn't do it but compiling under linux (and one that has the warning) it works fine.

As far as being able to run the GitHub tests commit https://github.com/ioccc-src/mkiocccentry/pull/721/commits/e7883a988b398613d7997a89af2abd714685ff03 should trigger that. Fixes a bug or maybe bugs with `-L` and `-I`. That's the subject of a todo item which is how I discovered it and why I fixed it. I'll bring it up in reply to that comment in the other thread.

PR #726, opened 2023-06-27
Partially fix jprint -Y str

xexyl:
A bug in searching for already existing matches was also fixed.

The function add_jprint_match() now takes a char *name_str and char
*value_str and these are used instead of pattern->pattern. The pattern
can now be NULL. If it is then the new jprint->matches list will be
acted on and otherwise pattern->matches list.

Testing of the call to add_jprint_match() where calls are does still
need to be tested more thoroughly (in particular with the name_str and
value_str args) but with the caveat that the value is the same as the
name in the current matches list this cannot be tested as easily.

The purpose of the change to the function is to help with printing the
entire file as there are no patterns (or name_args) when this happens
but until now there was no way to have a match without a pattern (but
for patterns see below). Now a new function or functions can be added so
that goes through the tree but adds every node to the jprint->matches
list.

As the concept of patterns as it stands is not correct but currently is
all that it can be (until the new soon to be added functions are added
to the jparse library) this will do but the struct and list might very
well be removed. However matches are more likely to stay.

This fix to -Y str is not perfect in that in some cases it can end up
matching other types but that depends on the JSON file contents and the
options. It might be said that -Y int is also somewhat improved but
again depending on the json file (h2g2.json which has also been updated)
this cannot be tested entirely.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,17 @@ come with another commit (if it proves useful; as the concept of `name_arg`s is
 currently not correct it might be that this won't be useful at all and that the
 struct is even removed).
 
+`jprint -Y str` partially works now. It's not perfect in that in some cases it
+can end up matching other types but that depends on the JSON file and options.
+The function `add_jprint_match()` can now have a NULL pattern and also NULL
+`pattern->pattern`. If `pattern == NULL` then the new `jprint->matches` is
+iterated through and set up. This change is to facilitate adding matches without
+patterns for when printing the entire file but this part has not been
+implemented yet. More needs to be tested when calling this function but other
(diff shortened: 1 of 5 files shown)


xexyl:
I believe that this commit is great progress in the sense of being able to print the entire file in the correct way and also possibly for getting the match _and_and name when they're different into the matches list rather than it being one or the other. But for printing the file entirely with the rules of options new functions will have to be added.

I will have to go afk soon but I hope to continue this when I get back. I might be able to do a bit more before I go afk but I might get that stuff out of the way too. Anyway that's what I think about this commit.

xexyl:
Something that I did not mention is it appears that `-Y int` might also have ben improved as far as matches hit.

PR #728, opened 2023-06-27
Fix to jprint -Q handling

xexyl:
Make sure to check that the match type is JTYPE_STRING before printing
quotes. Whether the match struct needs a type for the name and value or
just the value is TBD later but currently it just has one - the matched 
type.


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

I'm happy to help! I am hoping to do a bit more today before I do something else but I'm not sure of that one way or another.

PR #730, opened 2023-06-27
Fix adding of value/name matches for num/str

xexyl:
For strings and numbers (but see below) the adding of names versus 
values has been corrected. This required a lot of new code and more
modularity is called for but this can be done later. However now one can
get the name or value or both of these types and this goes for the -Y 
type option too.

The caveat is that there might be cases that have not been handled as 
for each type of name there has to be a check for types of values too
and there might be some I missed for numbers (and there might be the
same problem for strings but I am not sure yet). 

Others were not done yet but I wanted this in as it's a
significant accomplishment that will help move forward in the tool. For
that reason I have updated the jprint version again to 0.0.31 2023-06-27.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,12 @@ and printing syntax or quotes are requested but that the type of the match is a
 string (`JTYPE_STRING`) as well. Get rid of string bool in `jprint_match`
 struct as it's not needed with the `type`.
 
+Improvements in printing some types wrt quotes. Now the function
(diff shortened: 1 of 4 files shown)


lcn2:
thanks @xexyl

xexyl:
> thanks @xexyl 

Happy to help! I feel like this is great progress. I will do more tomorrow on this hopefully tackling the other types. 

Bedtime in about two hours from now so I will reply to anything tomorrow. 

Good night!

PR #733, opened 2023-06-28
Add jprint -R option to disable sub-tree searching

xexyl:
Add jprint -R option to disable sub-tree searching

New version of jprint - 0.0.32 2023-06-28.

The purpose of this option will be to allow one to find matches that are
not under a hierarchy of name_args. Although it was originally a 
misunderstanding it was a misunderstanding also in part because I feel 
that there is a great value in having it search in a non recursive way 
(in the sense of two name_args as described elsewhere) and it seems like
a waste of the code I already put in. Since there will be no ambiguity 
there should not be any problem.

The default will be to search in a recursive way but for now that is 
disabled as that feature is not yet implemented.

Fixed a potential memory leak that was discovered at the same time as 
well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,17 @@
 Add new good JSON test file 42.json under
 `jparse/test_jparse/test_JSON/good/42.json`.
 
+New `jprint` version "0.0.32 2023-06-28". Add option `-R` to disable recursive
+sub-tree searching. The temporary default is false which disables recursive
+searching but once recursive sub-tree searching is supported the default will be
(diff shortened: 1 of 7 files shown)


lcn2:
The addition of `-R` seems premature AND there are other things to consider in this regard.  We request that you hold off until other issues can be resolved.

xexyl:
> The addition of `-R` seems premature AND there are other things to consider in this regard. We request that you hold off until other issues can be resolved.

It's currently inactive though. I can remove it if you prefer.

PR #743, opened 2023-06-29
Add checks for matches of nulls and bools

xexyl:
Just like for strings and numbers the basic matching now is in for nulls
and bools. This is only for basic types - inside an array is not handled 
yet for instance.


lcn2:
Just so that the pull requests are not left hanging, we will merge what is here.

xexyl:
> Just so that the pull requests are not left hanging, we will merge what is here.

Thank you. Especially since at least one of them also updated the bug report script. 

Another has the new options you mentioned. 

Anyway as discussed I will hold off for now. Might get some things done with the other repo but not today. Just had a quick moment to reply to this. 

Hope you have a great rest of your day and I look forward to seeing what design you come up with!

PR #746, opened 2023-07-02
Updated CHANGES.md

xexyl:
New release 1.0.27 2023-07-02.


xexyl:
Until more is done in the planning of jprint this will likely be the last commit for now especially to do with json things unless I find a bug or some problem elsewhere.

I hope to in a bit work on the other repo but first I need a rest. I'll do more today or if not hopefully tomorrow though as I said elsewhere tomorrow I do have possible downtime. It might be that at that point though it would have already been past the time where I would work on the repos.

Good day and look forward to what you come up with for jprint!

Until then .. BCNU!

PR #747, opened 2023-07-02
Add more test json files - good and bad

xexyl:
The good one is a longer file with earthquake data much of which is
similar lines which seems like a good test case for jprint. For this 
theme probably no more need to be added but it seems like a fun one and 
will be useful (it's also minified which is good for that reason).

The bad one is the same but it's in both bad and bad_loc - with the
proper error file for the latter directory.


lcn2:
Having address a previous pull request (that we caused), we were able to complete the previous pull request.

In doing so, we noted down how to do the process, which went a lot faster this time.

Now we can complete this pull request: thanks @xexyl

xexyl:
> Having address a previous pull request (that we caused), we were able to complete the previous pull request.
> 
Thanks!

> In doing so, we noted down how to do the process, which went a lot faster this time.
> 
Good idea. What kind of problem was it (if you get a chance to reply .. I know there's a lot of things to do and more important things at that!) ?

> Now we can complete this pull request: thanks @xexyl

Thank you as well and you're welcome!

xexyl:
> > Having address a previous pull request (that we caused), we were able to complete the previous pull request.
> 
> Thanks!
> 
> > In doing so, we noted down how to do the process, which went a lot faster this time.
> 
> Good idea. What kind of problem was it (if you get a chance to reply .. I know there's a lot of things to do and more important things at that!) ?
> 
> > Now we can complete this pull request: thanks @xexyl
> 
> Thank you as well and you're welcome!

Oh I think the problem was the change in json file names etc. ?

lcn2:
> Good idea. What kind of problem was it (if you get a chance to reply .. I know there's a lot of things to do and more important things at that!) ?

Conflicting changes were made  to a file.

xexyl:
> > Good idea. What kind of problem was it (if you get a chance to reply .. I know there's a lot of things to do and more important things at that!) ?
> 
> Conflicting changes were made  to a file. 

I had that same problem today due to forgetting to pull changes. That was in the other repo and it happened because I thought I had pulled merges but that was this repo not the other one. 

I don't know how long I will be before I can work more on this repo or the other for that matter but I do think that I should be able to do a bit more this morning. 

Tomorrow is probably worse in this way but if all goes well Wednesday will be fine.

PR #751, opened 2023-07-13
Fix JSON_NUMBER regexp - resolve #750

### LINKED ISSUE #750: Bug report: jparse fails to correctly parse certain exp numbers
## Describe the bug

A **clear** and **concise** description of what the bug is.

## To Reproduce

Steps to reproduce the behavior (adjust steps as needed):

0. cd parse
1. ./jparse -v 0 -J 0 -- test_jparse/test_JSON/good/two-e0.json

## Expected behavior

The jparse tool should exit 0 with the message:

```
valid JSON
```

## Relevant Output

```
Warning: jparse_lex: at line 1 column 2: invalid token: 0x65 = <e>
syntax error node type JTYPE_NUMBER in file test_jparse/test_JSON/good/two-e0.json at line 1 column 2: <e>
ERROR[1]: ./jparse: invalid JSON
```

## bug report attachment

We request that you **PLEASE** run the command:

```sh
./bug_report.sh
```

and attach the bug report file that `./bug_report.sh` generated.
The bug report file will be a filename in the same directory with a
filename of the form: `bug-report.YYYYMMDD.HHMMSS.txt` and it will tell you the
name as well.

vvvvvvvvvvvvvvvvvvvvvvvvv --- Drag the file into the area below, please!

[bug-report.20230712.174127.txt](https://github.com/ioccc-src/mkiocccentry/files/12034303/bug-report.20230712.174127.txt)

^^^^^^^^^^^^^^^^^^^^^^^^^ --- Drag the file into the area above, please!

## Please provide the following information:

 - Did you attach the above requested bug report file?  yes
 - What is the airspeed Velocity of an Unladen Swallow? n/a

## Additional context

The so-called JSON spec allows for JSON numbers of the form:

[1-9][0-9]*[eE]...

For example, 2e0 is valid JSON.

xexyl:
New json parser and jparse version 1.0.10 2023-06-29.

A problem was reported on 12 July 2023 where the string 1e0 was reported as invalid JSON. This is because the regexp in the jparse.l only allowed (as far as exponent goes) a fraction followed by an exponent rather than that OR an integer followed by an exponent.

Another couple problems had to be resolved as well here to make make test work (at the top level directory) okay. First of all a googolplex is too big so the file googolplex.json was removed from the good directory. As we don't have a directory for general JSON that is although valid JSON cannot be parsed this file was removed instead of moved to another directory. Perhaps we need that functionality in the test suite too but that can be decided later on (as a reminder we do have it for the auth.json and info.json files). For reference a googolplex is
1e10000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000.

The googol.json (yes this is where Google gets their name :-) ) file was kept as it's not too big for C types.

Finally the test_ioccc/test_JSON/general.json/bad/info.timestamp-0.json is valid JSON so it was moved to the good subdirectory. I'm not sure how this was never triggered a problem unless it was just recently added.


lcn2:
Thanks @xexyl 

Interesting call on googolplex.json.  It is valid JSON, technically, however as you point out it cannot be represented as a number.

Is there a point of having a `struct json_number` for which `is_e_notation` is true but all of the "foo_sized" are false and more importantly the `converted` would be false.  The `as_str` could point at the string, and `as_str_len` would have a value > 0, so the only thing one could do is to print the string.

The code using `jparse/json_parse.h` is consistent in that if `converted` is false, then you don't have something to work with in the parse tree.  Changing code to add some sort of, `parsed` boolean so that if  `converted` is false and you have is the `as_str` string, could be done but is it worth it?

xexyl:
> Thanks @xexyl
> 
> Interesting call on googolplex.json. It is valid JSON, technically, however as you point out it cannot be represented as a number.

Unfortunate circumstance yes. We might want to have the test script do what I did for the chkentry tool? That is have it valid json but not valid according to some rule in this case because it can't be converted? I'm not sure.
> 
> Is there a point of having a `struct json_number` for which `is_e_notation` is true but all of the "foo_sized" are false and more importantly the `converted` would be false. The `as_str` could point at the string, and `as_str_len` would have a value > 0, so the only thing one could do is to print the string.

But would it be printed when it can't be parsed?
> 
> The code using `jparse/json_parse.h` is consistent in that if `converted` is false, then you don't have something to work with in the parse tree. Changing code to add some sort of, `parsed` boolean so that if `converted` is false and you have is the `as_str` string, could be done but is it worth it?

What would be gained by this? If you can't think of anything I don't see the point - at least not now. I might be too tired to understand what you're getting at fully though. I'm pretty sure that's the case unfortunately. Whether it'd be more clear later or not I don't know so what are you getting at exactly please?

I'm trying to finish up what I am working on but I might be able to reply to things here and there today. If not I'm sure I can do more tomorrow.

lcn2:
> > Thanks @xexyl
> > Interesting call on googolplex.json. It is valid JSON, technically, however as you point out it cannot be represented as a number.
> 
> Unfortunate circumstance yes. We might want to have the test script do what I did for the chkentry tool? That is have it valid json but not valid according to some rule in this case because it can't be converted? I'm not sure.
> 
> > Is there a point of having a `struct json_number` for which `is_e_notation` is true but all of the "foo_sized" are false and more importantly the `converted` would be false. The `as_str` could point at the string, and `as_str_len` would have a value > 0, so the only thing one could do is to print the string.
> 
> But would it be printed when it can't be parsed?
> 
> > The code using `jparse/json_parse.h` is consistent in that if `converted` is false, then you don't have something to work with in the parse tree. Changing code to add some sort of, `parsed` boolean so that if `converted` is false and you have is the `as_str` string, could be done but is it worth it?
> 
> What would be gained by this? If you can't think of anything I don't see the point - at least not now.

There are use cases where such big numbers are useful.

The `jparse` code could use [GNU MP](https://gmplib.org) as that code is released under GNU GPL v2.   On the other hand, that would be an extra complication for most code that doesn't want big numbers. Those applications that do want big numbers would already be using their own big number code, so perhaps it would be better to **NOT** pick given big number library in case the end point is using some different big number library. 

Passing big numbers across the wire in JSON has merit for applications that can use them, or for those applications that don't care about the numeric value.  For code that receives such an unconvertible number, the code could still process the string.  And programs that use big numbers could be given the string and do with it what they wish.

If we did this, then we would add a `parsed` boolean above the `converted` boolean to all of the `struct json_foo` in `jparse/json_parse.h`.  Then code in this repo that currently used a `converted` boolean would be examined.  In most code points that would change to checking the `parsed` boolean.  The other code points where a numerical value is needed, the `converted` boolean would need to be checked.  Use of the function `sem_node_valid_converted()` would need to be checked, and an equivalent `sem_node_valid_parsed()` would need to be used in most places.

xexyl:
> > > Thanks @xexyl
> > > Interesting call on googolplex.json. It is valid JSON, technically, however as you point out it cannot be represented as a number.
> > 
> > 
> > Unfortunate circumstance yes. We might want to have the test script do what I did for the chkentry tool? That is have it valid json but not valid according to some rule in this case because it can't be converted? I'm not sure.
> > > Is there a point of having a `struct json_number` for which `is_e_notation` is true but all of the "foo_sized" are false and more importantly the `converted` would be false. The `as_str` could point at the string, and `as_str_len` would have a value > 0, so the only thing one could do is to print the string.
> > 
> > 
> > But would it be printed when it can't be parsed?
> > > The code using `jparse/json_parse.h` is consistent in that if `converted` is false, then you don't have something to work with in the parse tree. Changing code to add some sort of, `parsed` boolean so that if `converted` is false and you have is the `as_str` string, could be done but is it worth it?
> > 
> > 
> > What would be gained by this? If you can't think of anything I don't see the point - at least not now.
> 
> There are use cases where such big numbers are useful.
> 
> The `jparse` code could use [GNU MP](https://gmplib.org) as that code is released under GNU GPL v2. On the other hand, that would be an extra complication for most code that doesn't want big numbers. Those applications that do want big numbers would already be using their own big number code, so perhaps it would be better to **NOT** pick given big number library in case the end point is using some different big number library.

I brought this up before and it was thought this was not a good idea due to then tying the parser to another dependency.

> 
> Passing big numbers across the wire in JSON has merit for applications that can use them, or for those applications that don't care about the numeric value. For code that receives such an unconvertible number, the code could still process the string. And programs that use big numbers could be given the string and do with it what they wish.
> 
That's true.

> If we did this, then we would add a `parsed` boolean above the `converted` boolean to all of the `struct json_foo` in `jparse/json_parse.h`. Then code in this repo that currently used a `converted` boolean would be examined. In most code points that would change to checking the `parsed` boolean. The other code points where a numerical value is needed, the `converted` boolean would need to be checked. Use of the function `sem_node_valid_converted()` would need to be checked, and an equivalent `sem_node_valid_parsed()` would need to be used in most places.

If you want to do that you can do that as you seem to have ideas on how to go about it and what it might entail.

lcn2:
> If you want to do that you can do that as you seem to have ideas on how to go about it and what it might entail.

We opened issue #752 to discuss this matter.  It is possible that it could be closed as "wont fix", but probably if the work to do this proves too ugly to use.  We don't think this will be the case.  We think this enhancement would be of benefit to the JSON parser.

xexyl:
> > If you want to do that you can do that as you seem to have ideas on how to go about it and what it might entail.
> 
> We opened issue #752 to discuss this matter. It is possible that it could be closed as "wont fix", but probably if the work to do this proves too ugly to use. We don't think this will be the case. We think this enhancement would be of benefit to the JSON parser.

Replied. I don't know how to go about it though. It seems like you have some ideas so perhaps in this case you'd like to address it? In the meantime I can look at the other document to progress in the new tools.

xexyl:
> > > If you want to do that you can do that as you seem to have ideas on how to go about it and what it might entail.
> > 
> > 
> > We opened issue #752 to discuss this matter. It is possible that it could be closed as "wont fix", but probably if the work to do this proves too ugly to use. We don't think this will be the case. We think this enhancement would be of benefit to the JSON parser.
> 
> Replied. I don't know how to go about it though. It seems like you have some ideas so perhaps in this case you'd like to address it? In the meantime I can look at the other document to progress in the new tools.

.. well possibly not look at it today but you know what I mean.

Issue #752, opened 2023-07-13
Enhancement: add parsed boolean to the JSON parser


lcn2:
The following valid JSON:

```json
1e1000000
```

is not correctly parsed by the JSON parser.  This is because as of "JSON parser version: 1.0.10 2023-06-29" there is an assumption that if a JSON node cannot be converted into a hardware value, it is not valid JSON.  This assumption is not correct in the case of JSON numbers.

In the case of `struct json_number`, there is a case where JSON can be parsed but not converted into a machine value.  Passing big numbers across the wire in JSON has merit for applications that can use them, or for those applications that don't care about the numeric value. For code that receives such an unconvertible number, the code could still process the string, or such application might be using a multi-precision library allowing the application to convert the JSON number string into a form that it can use.

We do **NOT** propose to add the conversion of big numbers via some multi-precision library.  Such a big number conversion can be said to be beyond the scope of a JSON parser.

We propose to add a boolean `parsed` to all "struct json_foo" in `jparse/json_parse.h` indicating if the given JSON node represents a valid parsed JSON.  The `converted` boolean would continue its current meaning.  In all cases, if `parsed` is set to false, `converted` will be set to false. 

Existing code that tests the `converted` boolean would be examined to determine if it should instead use `parsed`.  Such code points would need to be examined "by hand".

Functions such as  `json_process_decimal()` and `json_process_floating()` would continue to work as the do now in converting values.

The `is_decimal()` function works for big number integers of any size now, so that would continue be used as it is. A new function `is_floating_notation()` would be added to do similar character checking for floating strings and `is_e_notation()` for e-notation strings: both will work for big numbers of any size. 

Code that calls those functions such as `json_conv_number()` would still set `converted` boolean to false it they were not able to convert to a machine numerical value as they do now. The `parsed` boolean value would be based on the returns from `is_decimal()` and the new `is_floating_notation()` and the use `is_e_notation()`. Thus the `parsed` boolean would be set independently of it the  `json_process_decimal()` and `json_process_floating()` were able to set `converted` to true or false.

A `sem_node_valid_parsed()` function, along the lines of `sem_node_valid_converted()` would be added.  Code that called `sem_node_valid_converted()` would be examined to determine if it should call `sem_node_valid_parsed()` instead.

JSON debug code that printed information about a JSON number would be enhanced to be able to handle the case where a `struct json_number` that had `parsed` as true but `converted` was false.

[...]

PR #754, opened 2023-07-14
Add code for test of floating / e notations

xexyl:
These functions are probably not complete and might not be strictly 
correct AND/OR what is needed. Much of the tests do the same as the JSON
checks but these functions have fewer tests as C is less strict in some 
things when it comes to floating point than JSON. The JSON parser does 
all the checks including the additional checks.

Add check for more than one dot in JSON number strings. Not strictly 
necessary as the lexer would report an invalid token but this is for 
sanity. This check is also in the new functions.

This does NOT fully resolve issue #752!

Fix issues in jnum test code where '.' was not always detected (that is
the is_floating) and also the e notation (is_e_notation) was not always
detected. Rebuilt jnum_test.c for this to work right.

Running make prep passes.

XXX comments still exist and need to be looked at but again this issue
is not fully resolved and the rest might not be completed by me as the
other parts I am less clear on.


lcn2:
Thanks and best wishes on successful resolution of this issue @xexyl

xexyl:
> Thanks and best wishes on successful resolution of this issue @xexyl

Welcome. There are some things that I'm entirely unclear about but we can discuss those and see if it's something I can do or not. It'd be good if I could but we'll see.

PR #761, opened 2023-07-14
Remove reference to jprint in jparse/README.md

xexyl:
Until the other tools can be further decided upon nothing else was added
in place of jprint. I will do that later.


lcn2:
Thanks

PR #763, opened 2023-07-15
Add new JSON node helper macros

xexyl:
Three new macros have been added to check JSON nodes' converted and 
parsed booleans.

The CONVERTED_PARSED_JSON_NODE macro checks that both are true.
The CONVERTED_JSON_NODE macro checks that only converted is true.
The PARSED_JSON_NODE macro checks that only parsed is true.

These macros are now used in jparse/json_util.c but other files will be
checked and if necessary updated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.31 2023-07-15
+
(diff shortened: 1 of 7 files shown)


xexyl:
.. really time for a rest now. I'll check the other files when I'm back, however long that might be.

PR #764, opened 2023-07-15
Resolve most or all of issue #752

xexyl:
More use of macros added in commit
1e4240bf00fd3abe7e695c0eb302bd682dae12ea.
 
Renamed sem_node_valid_converted() to sem_node_valid() and update it to
use the macro CONVERTED_PARSED_JSON_NODE (for non-numbers) and 
VALID_JSON_NODE for numbers. For numbers either converted or parsed can
be true: if neither are true it is an error. For other types both 
converted and parsed have to be true.

Unless I am mistaken this commit fully resolves the issue #752. If this
is so I will then update the jparse and json parser version to 1.1.0 but
in case this is not done yet I haven't yet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.31 2023-07-15
+
+New JSON parser and jparse version "1.0.12 2023-07-15".
+
+Add more checks to JSON node `parsed` member so that if converted is false but
+parsed is true it's not an error (in the cases where parsed is allowed to be
+different it is not an error: if they're not allowed to be different we warn
+about it if they are different).
+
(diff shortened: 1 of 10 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl 

Sure. Should I update the version or is there more that has to be done?

lcn2:
> > Thanks @xexyl
> 
> Sure. Should I update the version or is there more that has to be done?

Looks good, please do so, thanks!

xexyl:
> > > Thanks @xexyl
> > 
> > 
> > Sure. Should I update the version or is there more that has to be done?
> 
> Looks good, please do so, thanks!

Done!

PR #768, opened 2023-07-16
Add placeholder JSON tools source and header files

xexyl:
An important key term is TEMPORARY PLACEHOLDER. The files are actually
copies of what were once jprint related but for each tool jprint / 
JPRINT were changed to their new tool counterparts. This is done this 
way because some of the code is common and this way they are in the
repo. A lot of code will be removed but some will be kept. Much will 
also be changed. Option parsing in particular is useful but this will
also have some changes.

As it will take more work to get them in order I have kept the other
things in too for now but this GREATLY simplifies starting the tools
which seems like a worthy thing to do.  A very nice way that this
simplifies starting the tools is that the parsing of the JSON file is
already done. Along with this some of the options being the same and 
some of the structs being the same (though they will have to be modified
and some will be removed) it will help get started.

Respective test scripts for each tool have also been added which are 
also updated copies of the old script for jprint. These are not run yet
because there is no point in running them yet and it would only slow
down make test.

With the exception of the scripts the required changes will be done soon
\- hopefully sometime in the coming week (real weeks start on Monday) -
but what this means is that TEMPORARILY the tools function as the
incomplete no longer useful jprint. By required changes I mean that the
unnecessary code will be removed and option parsing / usage strings will
be updated (whether the option parsing is completed in the next week is
indeterminate).

Updated jparse/Makefile and run make depend.

Update .gitignore: fixed references of jprint and add the other tools.

This is certainly not perfect, is incomplete and has incorrect options 
and concepts but again this will simplify the starting of the tools.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,10 +34,14 @@ jparse.tab.c
 jparse.tab.h
 jparse.xml
 jparse/jparse
-jparse/jprint
+jparse/jfmt
+jparse/jnamval
+jparse/jval
 jparse/lex.yy.c
 jparse_test.log
-jprint_test.log
+jfmt_test.log
+jnamval.log
+jval.log
 jsemtblgen
 jstr_test.out
 jstr_test2.out
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,36 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.32 2023-07-16
+
+Add **TEMPORARY PLACEHOLDER** source and header files for `jfmt`, `jval` and
+`jnamval` tools. The key term that is important is _TEMPORARY PLACEHOLDER_. The
+files are actually copies of what were once jprint related but for each tool
+jprint / JPRINT were changed to their new tool counterparts. This is done this
+way because some of the code is common and this way they are in the repo. A lot
+of code will be removed but some will be kept. Much will also be changed. Option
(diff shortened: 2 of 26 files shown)


xexyl:
I have to go afk now .. might get started more on the tools later but this is far from certain. Either way although it might seem like some to be pointless it is VERY VERY USEFUL this way. It works very well in my head and will greatly simplify starting the tools which is a very good thing.

xexyl:
Something I forgot to mention btw: I set the versions to 0.0.0 as not even the usage strings are accurate: well some options maybe but not all. When I update the options and usage string I will update the version of the respective tool.

But for now I must go. Hope you're having safe travels / a safe good time!

lcn2:
Is a beginning @xexyl 👍

xexyl:
> Is a beginning @xexyl 👍

A good one too though the other pull request that I made after is of course better - thanks!

I hope to do more in at least a bit: might see if I can rest a bit first. Doubt that I can sleep but might try resting anyway. 

I would have started earlier but I hadn't changed the Notifications app settings on the new iPad that I got a couple months ago or whenever it was. It was really annoying having notifications not showing if screen was locked as typically it is locked. But then I looked at other settings too. 

Then I did that on the iPhone that I got last year - the first one since the iPhone 8 Plus. 

But I didn't finish it as I saw how much time had passed and since most likely it isn't necessary for now. 

Hope you're having a safe and fun mini holiday!

PR #775, opened 2023-07-18
Proper parsing of jval -S and -n options

xexyl:
Updated version to 0.0.2 2023-07-18.

Instead of trying to do the parsing of the string (which might have
invalid JSON format or can't be converted in a JSON valid way) or number
(which might not be valid, can't be converted but can be parsed) etc. we
now have a struct json_number * and a struct json_string * in the
compare struct. Now the json_conv_(string|number) functions are used.
     
Added debug call for string to verify it worked okay. This has not been
done for numbers as there is much more to do than to print just any
field. Perhaps a general function should be added to the JSON API (if
this does not already exist - and it might) but that can be addressed
later.
 
Sequenced exit codes.

With this commit all option parsing for jval is complete! Note that 
there is no code to actually do the comparison: that has to come later 
after option parsing of all tools is done and then the man pages are 
written and then discussion is had over all of that.

The next step is either test code for the -S / -n options to jval (which
shouldn't be too hard though for numbers depending on how it's done it 
might be more involved and possibly not yet worth it) or else work on 
jnamval option parsing. After that man pages can be written and we can 
go from there.


xexyl:
.. and now it's really time for a rest ... will hopefully do more later!

PR #788, opened 2023-07-21
Make jval/jnamval exit 7 on number conv error

xexyl:
If a number cannot be represented by a C type exit 7. This means for
jnamval the old 7 (no matches found) is now 8 so that they are the same.

Use macros to verify converted / parsed json conversions both for
strings and numbers in both jval and jnamval (for -n and -S options).
This is partly related to the exit code change so it's in the same
commit.     
            
Add json log (currently debug level none so it's always printed but this
will change just like all the other JSON_DBG_NONE references in these
tools will) for what number to compare.
                
Updated man pages for above mentioned exit code changes.
                
Updated usage strings.


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

You are most welcome! Thanks, btw, for resolving the conflict the other day that arose from you modifying the `json_util_README.md` file.

I see you noted this pull request: I'll look at it. The example output you suggested is helpful. Thanks!

PR #790, opened 2023-07-22
New mkiocccentry version - add -y yes option

xexyl:
This option overrides the yes_or_no() function in MOST cases. It does
NOT override it in the case of overwriting the answers file or in the
case of if you're a previous winner. There might be a few other cases.

This resolves a bug in make test where one would get:

    yes: standard output: Broken pipe

though this did not occur when running mkiocccentry_test.sh directly.

Updated mkiocccentry version to 1.0.1 2023-07-22 and updated the
necessary json files in the test_ioccc/test_JSON directories. This was
done by:

    find . -type f -name '*.json' -execdir sed -i -e \
    's/"mkiocccentry_version" : "1.0 2023-02-04"/"mkiocccentry_version" : "1.0.1 2023-07-22"/g' -e \
    's/"mkiocccentry_version":"1.0 2023-02-04"/"mkiocccentry_version":"1.0.1 2023-07-22"/g' '{}' \;

The vermod.sh script was not used as it has a flaw in that it would 
update the version of any tool that matched the version. That should 
probably be fixed but that can happen another time .. or not.


xexyl:
I forgot about my sgit tool! That would have simplified matters. Oh well. It's all good anyway. Hope to do more later on the json related issues but that's far from certain I'm afraid.

Issue #785, opened 2023-07-20
Bug report: jparse -J 3 does not correctly print JSON number values


lcn2:
## Describe the bug

When `-J 3` is in effect, JSON numbers are printed as "false" instead of the string containing their value

## To Reproduce

Steps to reproduce the behavior (adjust steps as needed):

0. ./jparse/jparse -J 3 jparse/test_jparse/test_JSON/good/two-e0.json
1. observe the output:

```

JSON tree[3]:	lvl: 0	type: JTYPE_NUMBER	{converted,parsed}: false
in parse_json(): JSON debug[1]: valid JSON
valid JSON
```

## Expected behavior

The JSON number as a string (i.e., `as_str`) is printed instead of "false".

## Relevant Output

Consider the "type: JTYPE_NUMBER" line output from commands as well:

```sh
./jparse/jparse -J 3 jparse/test_jparse/test_JSON/good/42.json
./jparse/jparse -J 3 jparse/test_jparse/test_JSON/good/top_level_array.json
```

## bug report attachment

We request that you **PLEASE** run the command:

```sh
./bug_report.sh
```

and attach the bug report file that `./bug_report.sh` generated.
The bug report file will be a filename in the same directory with a
filename of the form: `bug-report.YYYYMMDD.HHMMSS.txt` and it will tell you the
name as well.

vvvvvvvvvvvvvvvvvvvvvvvvv --- Drag the file into the area below, please!

[bug-report.20230720.133326.txt](https://github.com/ioccc-src/mkiocccentry/files/12113854/bug-report.20230720.133326.txt)

^^^^^^^^^^^^^^^^^^^^^^^^^ --- Drag the file into the area above, please!

## Please provide the following information:

 - Did you attach the above requested bug report file?  yes
 - What is the airspeed Velocity of an Unladen Swallow? see http://style.org/unladenswallow/

## Additional context

The reason to print the `as_str` as the "value" for a NUMBER is that one may avoid issues related to the machine hardware representation, or lack of representation of the "value".

### Suggested enhancement

The text inside "{...}" for "JTYPE_NUMBER" is needlessly verbose and simply refers to two of the several `struct json_number` booleans that pertain to a JSON NUMBER: all be it two of the less likely to be needed booleans associated with `struct json_number`.

The `-J 3` output is intended to be for debugging purposes and is intended for general use.  Such `-J` output is intended for those who are familiar with JSON parser structure internals.

[...]

PR #792, opened 2023-07-23
Make !VALID_JSON_NODE condition not an error

xexyl:
In the parse_json_foo() functions it is NOT an error if both converted and parsed are false. This was brought up on GitHub numerous times (in issue #752) so I have made this change. I had originally updated it to use the macro that I added (one of a number of them, see below) as it previously was an error if !converted but given the purpose of issue #752 and what's been discussed in issue #785 it appears that that check should not be there at all.

Given that the macros are helpful I will recite (yes I am hijacking the definition :-) ) them:

    #define VALID_JSON_NODE(item) ((item)->converted || (item)->parsed)
    #define CONVERTED_PARSED_JSON_NODE(item) ((item)->converted && (item)->parsed)
    #define CONVERTED_JSON_NODE(item) ((item)->converted && !(item)->parsed)
    #define PARSED_JSON_NODE(item) ((item)->parsed && !(item)->converted)

Thus VALID_JSON_NODE reports where either of the booleans converted or parsed is true; CONVERTED_PARSED_JSON_NODE reports that both are true; CONVERTED_JSON_NODE reports that converted is true but parsed is not; and PARSED_JSON_NODE reports that parsed is true but converted is false.

New version of json parser and parser: 1.1.1 2023-07-23.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -26,6 +26,12 @@ and `is_integer` booleans, also respectively.
 Improved JSON debug messages from `json_process_decimal()` and
 `json_process_floating()`.
 
+In the `parse_json_foo()` structs make it so that if `!converted && !parsed`
+it's not an error. These should I believe never actually happen anyway but on
+GitHub it was suggested this condition should not be considered an error.
+
+Updated json parser version and jparse version string to "1.1.2 2023-07-23".
+
 
 ## Release 1.0.37 2023-07-22
 
--- a/jparse/jparse.h
+++ b/jparse/jparse.h
@@ -52,7 +52,7 @@
 /*
  * official jparse version
  */
-#define JPARSE_VERSION "1.1.1 2023-07-21"		/* format: major.minor YYYY-MM-DD */
+#define JPARSE_VERSION "1.1.1 2023-07-23"		/* format: major.minor YYYY-MM-DD */
 
 /*
  * definitions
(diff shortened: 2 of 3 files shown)


xexyl:
... actually not sure this is correct in every case like strings. Converting to draft. Will fix after a rest.

xexyl:
In particular for anything but numbers this has to be incorrect as otherwise valid JSON is reported for cases it is clearly not. Maybe a third boolean has to be added. I'll elaborate later but it might be that's just a thought in a tired head.

lcn2:
We don't think this change is quite correct.

See [comment 1646807312](https://github.com/ioccc-src/mkiocccentry/issues/785#issuecomment-1646807312).

xexyl:
Closing this per a comment I made in the other thread that makes this not work. I think more discussion needs to be had in order for this to be addressed after all.

xexyl:
Oh you closed it already. I thought it was incorrect too.

xexyl:
> We don't think this change is quite correct.
> 
> See [comment 1646807312](https://github.com/ioccc-src/mkiocccentry/issues/785#issuecomment-1646807312) and see [comment 1646809829](https://github.com/ioccc-src/mkiocccentry/issues/785#issuecomment-1646809829).

I don't either.

PR #793, opened 2023-07-24
Improve JSON validity convenience macros

xexyl:
They now check that the node != NULL. They look now like:

    #define PARSED_JSON_NODE(item) ((item) != NULL && ((item)->parsed == true))
    #define CONVERTED_PARSED_JSON_NODE(item) ((item) != NULL && (((item)->parsed == true) && ((item)->converted == true)))
    #define VALID_JSON_NODE(item) ((item) != NULL && (((item)->parsed == true) || ((item)->converted == true)))

Although the explicit check of == true for booleans is not my style I kept them in as looking at the macro without looking at the structs it might not be as clear that it's a boolean but this way it's clear.

### DIFF
--- a/jparse/json_parse.h
+++ b/jparse/json_parse.h
@@ -35,9 +35,9 @@
 /*
  * convenience macros
  */
-#define PARSED_JSON_NODE(item) ((item)->parsed == true)
-#define CONVERTED_PARSED_JSON_NODE(item) (((item)->parsed == true) && ((item)->converted == true))
-#define VALID_JSON_NODE(item) (((item)->parsed == true) || ((item)->converted == true))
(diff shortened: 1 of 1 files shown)


lcn2:
True @xexyl 😀

xexyl:
> True @xexyl 😀

```c
#include <stdbool.h>
#include <stdio.h>
int main(void)
{
    bool is_false;

    is_false = (((!!!false - true) + true) == true) == (!!!!true);

    printf("%d\n", is_false);

    return is_false == !!!!true;
}
```
```sh
$ make bool ; ./bool ; echo $?
cc   -L/opt/local/lib  bool.c   -o bool
1
1
```

PR #795, opened 2023-07-24
Improve parsed,converted flags in debug output

xexyl:
As parsed and converted booleans are not strictly part of the numbers or
strings but rather just booleans indicating if it's validly parsed and 
converted these boolean flags ('p' and 'c') now are suffixed with a ':'.
If both are true then it will look like 'p,c:' but if just parsed is
true, for example, it will look like 'p:'.

Originally I had the thought of comma separating all flags but this 
would be problematic as for each one it would have to check if there is
another flag to print and that would greatly complicate things as there
are a lot of flags to inspect.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,26 @@ seems like it might be a good place to do that.
 Fixed `warning: incompatible pointer to integer conversion passing 'struct json
 *' to parameter of type 'enum item_type'` in `json_util.c`.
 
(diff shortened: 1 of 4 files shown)


xexyl:
.. forgot to mention I updated the version strings ..

lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help. Will do more tomorrow if there's something to do. Good day!

PR #796, opened 2023-07-24
jval and jnamval -S and -n options now a list

xexyl:
As the options can be specified more than once we need a list of the 
options parsed. I am still not clear on when it's OR or AND but that can
be figured out later.

Unfortunately the -n mode does not work correctly in that checking for 
numbers being converted seems to be broken, seemingly from a change in
the convenience macros. I will have to address this another time. The -S 
option is tested meaning added to list and freeing the list works fine.
It is possible that I made a mistake in the -n processing even though
they're mostly the same.

Some details might change later on but with this I believe all options
are parsed completely. This of course could also change but I believe at
this time it's done.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,13 +32,16 @@ For strings it's likewise just the rest is string data, not number data:
 
 Updated json parser and jparse version strings to "1.1.2 2023-07-24".
 
+New definition of recently removed macro `CONVERTED_JSON_NODE`:
+
+```c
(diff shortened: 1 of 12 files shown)


lcn2:
Thanks @xexyl

xexyl:
> Thanks @xexyl

Happy to help. Off now - zoom meeting soon and then done for the day at computer most likely.

Good day!

PR #797, opened 2023-07-25
Move common json tool variables to common struct

xexyl:
The struct jfmt, struct jval and struct jnamval now have a single 'struct json_util common' which has the variables that are common to all three tools in json_util.h. Code moved to json_util.c.

The structs (jfmt|jval|jnamval)_number are now json_util_number in json_util.h and the associated code has been moved to json_util.c.

The code for the common -L option to the three tools has also been moved to the json_util.c file.

Any other common (to all three tools) options not mentioned here have also been moved to json_util unless some were missed by accident. I don't believe that is the case but if there are any they will be found and moved accordingly.

The options common to just jval and jnamval have NOT been moved to common structs just yet. That will happen in another commit.

Now why json_util.[ch] instead of some new file? Because they are json utilities and the document that describes them is json_util_README.md not something like jutil_README.md.

New version for these tools: 0.0.4 2023-07-25.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,27 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 1.0.40 2023-07-25
+
+New version for `jfmt`, `jval` and `jnamval`:  "0.0.4 2023-07-25". Moved common
+variables from the structs for each tool (`struct jfmt`, `struct jval` and
+`struct jnamval`) to `json_util.h` and prefix with `json_util` rather than the
+`(jfmt|jval|jnamval)_`. More work needs to be done here as `jval` and `jnamval`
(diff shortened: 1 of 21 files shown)


lcn2:
Yea 🍾

xexyl:
> Yea 🍾

After taking a screenshot and zooming way in I believe that's a bottle so here's a funny story when I turned 30: at dinner and my brother was there along with some others. I looked over halfway through the dinner (teppanyaki) and I saw that at least half of his beer was left. I looked at him and said 'what kind of alcoholic are you?' Shortly after that he finished his drink.

I believe he doesn't really drink much except on special occasions now. I never have and never will. But that's one of numerous funny stories at teppanyaki.

I'll do more tomorrow. Hope you have a good day!

PR #798, opened 2023-07-25
Update jval common code to be in json_util

xexyl:
Although the code for both jval and jnamval that is common to each other
are in json_util now only jval uses that code for now. Some code is the
same but has some minor functional differences and that code has not
been moved over. It might be that it can be later on and at least the
match functions can be later on but the parsing maybe not. This is not
really a problem.

jnamval has options that are not in jval and so in addition to moving
the common code to json_util (which again is not yet done) the code for
that also remains (and will always remain) in jnamval.[ch].

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,27 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 1.0.40 2023-07-25
+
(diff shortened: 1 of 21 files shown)


lcn2:
👍 @xexyl

xexyl:
> 👍 @xexyl

Happy to help .. I'll do `jnamval` tomorrow in the same way as this.

PR #807, opened 2023-07-29
fprintf() -> fpr() in jval and jnamval

xexyl:
In jfmt, jval and jnamval when printing the file contents (which is done
in temporary way) use fpr() not fprintf() now that the tty issue is
resolved.

If jval -c/-C or jnamval -c/-C is used without an arg is a command line
error.  This is done in the sanity check function of each tool.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@ All "print function call"s are now followed by a call to
 `chk_stdio_printf_err()` to properly detect if there was a
 failure in the "print function call".
 
+Add `print_test` execution to `make test` in `jparse/test_jparse/Makefile`.
+
 New versions of `jfmt`, `jval` and `jnamval` with some minor bug fixes and
 enhancements: `"0.0.6 2023-07-28"`, `"0.0.7 2023-07-28"` and `"0.0.6
 2023-07-28"`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,10 +51,8 @@ implementation details until we have discussed output, how `jfmt` should format
 code, search routines are added and numerous other things. This was just an
 aside, one might say.
 
-For `jval` and `jnamval` use of `-c` or `-C` without an arg is considered an
-error and this is in the sanity check function of each tool. Added `XXX` comment
-to the print count function of each tool that points out that currently the
-count is not correct. This will be removed in time.
+Added `XXX` comment to the print count function of each tool that points out
+that currently the count is not correct. This will be removed in time.
 
 Update `jval(1)` man page - add exit code 8 to list of exit codes.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,5 @@
 # Major changes to the IOCCC entry toolkit
 
-
 ## Release 1.0.42 2023-07-29
 
 Add `jparse/test_jparse/print_test` test tool to test if various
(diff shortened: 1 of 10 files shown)


lcn2:
> If jval -c/-C or jnamval -c/-C is used without an arg is a command line
> error. This is done in the sanity check function of each tool.

Without an arg:

```sh
jval foo.json
```

Matches all "JSON values" and prints them.  As the `json_util_README.md` file states:

```
The jval command, without any "args" on the command line, will cause the printing of any JSON values that are not otherwise restricted by a -t type or -l lvl option.
```

So by adding `-c` one is asking `jval` to count all "JSON values" in the file instead.  Thus use of `-c` without an "arg" is **NOT** an error.

See "jval example 6" for more details such as using `jval -C` without any "args" as well.

Regarding `jnamval` without args.  A similar thing applies.  As the spec states:

```
By default, jnamval prints JSON member values, just like the jval utility only prints JSON values.
```

So without args, `jnamvsl` matches and prints all JSON members.

Think of it this way: without any args, both commands match everything.  In the case of `jval`, all "JSON values" are matched.  In the case of `jnamval`, all "JSOM members" are matches.

So without args, the `-c` will count everything.  Without args, the `-C` will print the count of each match value.

So the description of this pull request is not correct.

Do you want us to reject this pull request so that you can reissue the other part as a new pull request:

```
In jfmt, jval and jnamval when printing the file contents (which is done
in temporary way) use fpr() not fprintf() now that the tty issue is
resolved.
```

Or do you want to issue a new pull request that reverts the part about:

```
If jval -c/-C or jnamval -c/-C is used without an arg is a command line
error. This is done in the sanity check function of each tool.
```

Let us know how you want to proceed: Reject this pull request or wait for a later pull request that corrects it?

xexyl:
> > If jval -c/-C or jnamval -c/-C is used without an arg is a command line
> > error. This is done in the sanity check function of each tool.
> 
> Without an arg:
> 
> ```shell
> jval foo.json
> ```
> 
> Matches all "JSON values" and prints them. As the `json_util_README.md` file states:
> 
> ```
> The jval command, without any "args" on the command line, will cause the printing of any JSON values that are not otherwise restricted by a -t type or -l lvl option.
> ```
> 
> So by adding `-c` one is asking `jval` to count all "JSON values" in the file instead. Thus use of `-c` without an "arg" is **NOT** an error.
> 
> See "jval example 6" for more details such as using `jval -C` without any "args" as well.
> 
> Regarding `jnamval` without args. A similar thing applies. As the spec states:
> 
> ```
> By default, jnamval prints JSON member values, just like the jval utility only prints JSON values.
> ```
> 
> So without args, `jnamvsl` matches and prints all JSON members.
> 
> Think of it this way: without any args, both commands match everything. In the case of `jval`, all "JSON values" are matched. In the case of `jnamval`, all "JSOM members" are matches.
> 
> So without args, the `-c` will count everything. Without args, the `-C` will print the count of each match value.
> 
> So the description of this pull request is not correct.
> 
> Do you want us to reject this pull request so that you can reissue the other part as a new pull request:
> 
> ```
> In jfmt, jval and jnamval when printing the file contents (which is done
> in temporary way) use fpr() not fprintf() now that the tty issue is
> resolved.
> ```
> 
> Or do you want to issue a new pull request that reverts the part about:
> 
> ```
> If jval -c/-C or jnamval -c/-C is used without an arg is a command line
> error. This is done in the sanity check function of each tool.
> ```
> 
> Let us know how you want to proceed: Reject this pull request or wait for a later pull request that corrects it?

Since it has other things too I'll correct it. I wasn't even sure of this. It did seem like it _could_ be in conflict but I wasn't sure about that either.

xexyl:
Fixed.

lcn2:
> Fixed.

Fixed as in this pull request was modified and so we should process it OR fixed as in there is another pull request to correct it and we should wait for that later pull request, @xexyl?

xexyl:
> > Fixed.
> 
> Fixed as in this pull request was modified and so we should process it OR fixed as in there is another pull request to correct it and we should wait for that later pull request, @xexyl?

This one. Commit https://github.com/ioccc-src/mkiocccentry/pull/807/commits/d12d2e45b8cf4c9847796c028ccc48e1d7b5628f.

xexyl:
And thanks for the reminder. I remember the spec details you referred to now. Anyway yes it's this pull request in the commit noted above.

I'll look at other things tomorrow like with make test and so on: though I asked a question about that one.

lcn2:
Thanks for doing the pull request update @xexyl

PR #808, opened 2023-07-30
Improve, typo fix and bug fix print_test.c

xexyl:
Now checks the file descriptor versions of the printf functions.

Added 'not_reached();' after an error function was called.

There might have been other improvements or fixes.

New version 1.0.1 2023-07-30.


xexyl:
Doing other things and I don't think I'll do more here today except maybe reply to comments later on if you do. We'll see I guess.

lcn2:
Thanks @xexyl

PR #811, opened 2023-08-01
Move print_test call to jparse_test.sh

xexyl:
New version of jparse_test.sh script 1.0.3 2023-08-01. The script now
runs print_test and directs it to the log file to prevent clutter. The
original thought was to have a new script run_print_test.sh but this
seemed unnecessary complication and less time efficient as all it would
have to do is run the test tool and check exit status but jparse_test.sh
would call it. Since the jparse_test.sh is already set up and it only
needed an additional function it is not necessary to have a new script.

If the -p print_test flag is used it runs the print_test checker. It
does it this way because other calls to the script would repeat the test
unnecessarily and thus slow it down. Specifically it is not run when
running the script from test_ioccc/.


xexyl:
> Thank you @xexyl

You are most welcome! I hope that soon I can update the man pages to the tools for the `-F` option, the man page for this script and add the man page for `print_test` tool but we'll see.

PR #813, opened 2023-08-02
Move parse_verbosity(3) to dbg facility

xexyl:
Since the debug facility is the one that uses a verbosity_level that I added here I have moved the function from jparse/util.c to dbg/dbg.c.

The dbg.3 man page has been updated with this and also some other improvements. There is a recursive reference to the man page for parse_verbosity(3) in the man page just for fun.

The dbg related changes have already been added to the dbg(3) repo as a pull request.

The dbg_example.c file that I wrote for this repo and the dbg(3) repo has also been updated with a comment change, referring to not strtol(3) but rather the parse_verbosity(3) function.

The dyn_array/dyn_test.c no longer has the parse_verbosity() function as this is in the dbg(3) facility which the dyn_array facility uses.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,21 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.46 2023-08-02
+
+Move `parse_verbosity(3)` to dbg/dbg.c as it is the dbg code that uses the
+`verbosity_level` variable. Ran `make test` to verify that this works okay.
+
+Updated `dbg.3` man page to add this function to the man page. Also improved the
+man page in some clarification of other functions (call semantics for example).
+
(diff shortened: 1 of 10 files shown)


xexyl:
Next for the `dyn_array(3)` repo to do this as it requires the `dbg(3)` repo anyway... at least: the code here does. If it does not include it in the dyn_array repo then something else will have to be done. I'll see and if it is a problem I'll reply to this comment. Else I'll open a new pull request: either way after that it'll be time for a rest.

I hope to work on the man pages of the three tools and add one for print_test later on.

xexyl:
The above problem is solved in another pull request and synchronised to the dyn_array facility in an open pull request in that repo too.

Should the macro be `DBG_USE` or something else ? I see an argument for both ...

PR #818, opened 2023-08-03
Improve location tool - make more generally useful

xexyl:
New version of location is 1.0.1 2023-08-03.

New option -s searches by substring rather than exact match.

New option -a shows all matches. Only useful with -s as otherwise it's an exact match.

To implement -a we have two new re-entrant functions:

    char const *lookup_location_name_r(char const *code, size_t *idx, struct location **location, bool substrings);
    char const *lookup_location_code_r(char const *location_name, size_t *idx, struct location **location, bool substrings);

These are the same as lookup_location_name() and lookup_location_code() except that they allow for calling them additional times to get the next match. If idx != NULL set *idx to the position + 1 for the next call. If location != NULL then set *location to the matching location. Then in location's main() (see location_main.c) it prints (if code/name != NULL) both the return value and then the name/code, depending on whether -N is used. When *idx >= the sizeof the table or no more matches are found then the functions will return NULL, signalling to location's main() that there are no more matches. This also means that instead of calling err() on NULL return value it just exits 1 if -a is used and no matches are found.

Note that without -N -s and -a are not very useful as the function that does country code checks explicitly checks for two characters so there's no real substring match that can be done. However the function was added to have matching functions.

Use of -s requires at least one arg.

Use of -a with an exact country name (if -N) is not very useful as it will only find an exact match (obviously). Nevertheless the only error condition is if -s is used without an arg.

With -a if -v 1 show text in the form of:

    XX ==> Anonymous location
    DE ==> Germany

or more generally:

    foo ==> bar

and

    Anonymous location ==> XX
    DE ==> Germany
    ...

Example command and output:

    ./soup/location -a -s -N -v 1 germ
    German Democratic Republic ==> DD
    Germany ==> DE

The joke here, of course, is that 'Germ'any is a country full of 'germ ridden people' (obviously not but it's a fun pun of mine :-) ).

Note that chkentry does NOT use these new features as the checks must be exact in that tool. The rationale of these changes is to help users of mkiocccentry find the right country code, even if they know their own (as they can also find country codes / names for other countries).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,65 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.47 2023-08-03
+
+New version of `location` tool: `"1.0.1 2023-08-03"`.
+
+Improved tool to allow for `-a` (show all) and `-s` (substring search). Uses new
+re-entrant versions of the functions `lookup_location_name()` and
+`lookup_location_code()`. These new functions take a `size_t *idx` which is set
+to the _next_ element in the table (if != NULL) so that the next call can skip
+continue _after_ the previous element. They also take a `struct location **`
+which stores the _found_ location. This is useful for `-v 1` as one can then see
+something like:
+
+
+```sh
+./location -a -s -N -v 1 'united'
+United Arab Emirates ==> AE
+United Kingdom of Great Britain and Northern Ireland (the) ==> GB
+United States Miscellaneous Pacific Islands ==> PU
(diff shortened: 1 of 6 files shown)


xexyl:
I forgot to mention that the man page was also updated ...

lcn2:
Thanks and going back to sleep, @xexyl

xexyl:
> Thanks and going back to sleep, @xexyl

Sleep tight! And happy to help.

PR #822, opened 2023-08-05
Add bad location test JSON files, update .gitignore

xexyl:
Update .gitignore: add '/archive': a directory in my working repo of old files that although useful to me should not be added to the repo. Since archived files are not really something that should be in the repo anyway adding '/archive' to the .gitignore seems reasonable.

Add some bad test JSON files in the
'jparse/test_jparse/test_JSON/bad_loc' directory, updated Makefile rule Added appropriate 'err' files.  Updated Makefile rule 'rebuild_jparse_err_files' to not show the command being run for each file to make the output clean (and to show what it looks like in the README.md file - which has been updated due to slight changes).


lcn2:
So as to not confuse the IOCCC directory "archive" that is used with the "archive" of this repo, please rename "archive" to something else. 

Yes, these are two different repos, but still. An archive should be about preserving things, not fit hiding things. 

For example, there is a `.gitignore` exclusion on "NOTES" (which should be "/NOTES/" in `.gitignore`) that can contain various notes and stuff to keep around.

xexyl:
Good points. Let me see ... I have a better idea. I'll just put it in `.git/info/exclude` like I did with some other files. One moment.

xexyl:
See https://github.com/ioccc-src/mkiocccentry/pull/826/commits/8e09cff1d44d638e77b5eeb4cbcbc3d897380229 for the removal of the path. Thanks for pointing it out.

PR #823, opened 2023-08-05
Add -I to jnamval (same as from jfmt)

xexyl:
New version of jnamval is 0.0.11 2023-08-05.

Add -I option with the same semantics as jfmt. This option does not 
exist in jval and it is not clear if it is supposed to. It would be very
easy to add as it's common code but it might not be necessary. This is 
TBD later.

Updated jnamval.1 to add the -I option.

Fix typo in jnamval usage string that could cause confusion - it
referred to a non-existent option. It was supposed to be an option arg
and it now is correct.

Typos were discovered in jfmt.1, jval.1 and jnamval.1 when updating
jnamval.1 and these were also fixed.


xexyl:
QUESTION for you Landon: does `jval` need `-I` also ? It's common code so it would be very easy to add but I believe it does not need it? If that's the case I could of course make it only common to `jfmt` and `jnamval` though that seems like unnecessary work.

lcn2:
> QUESTION for you Landon: does `jval` need `-I` also ? It's common code so it would be very easy to add but I believe it does not need it? If that's the case I could of course make it only common to `jfmt` and `jnamval` though that seems like unnecessary work.
> 
> 

We cannot think of a reason who `jval` needs `-I stuff`.

xexyl:
> > QUESTION for you Landon: does `jval` need `-I` also ? It's common code so it would be very easy to add but I believe it does not need it? If that's the case I could of course make it only common to `jfmt` and `jnamval` though that seems like unnecessary work.
> 
> We cannot think of a reason who `jval` needs `-I stuff`.

Good. Just wanted to be sure.

PR #824, opened 2023-08-05
Add jnamval -P types option - types to print

xexyl:
The list is the same as the -t option as the -t option types has all the
types. The macros of the type bits are used and the same function is
used as both are for types. The only difference in use is that one is 
what types to match and the other is what types to print.

New functions for checking what to print were added however to be clear 
in code use. This is not a problem with parsing the options however. The
jnamval_print_type functions are the same as the jnamval_match_type
functions: just the names have been changed (or added). Realistically
they could just be #defined to the _match functions but this is not 
done. A comment should probably be added to the functions to indicate
this but that can come later.

Man page updated with new option.


xexyl:
See update 0 in [comment 1666485500](https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1666485500) as there has to be some changes after all. What is committed is fine but it needs some fixes as well since you want the `-t` list simplified. 

I hope to address that later today.

PR #825, opened 2023-08-05
Fix jnamval -t and -P lists

xexyl:
The types list allowed by -t has been fixed. As it's now the same in
both jval and jnamval it's in json_util.[ch] not in the files for both
jval and jnamval.

There was a fix in one of the parse functions where it did not allow for
a comma separated list but only one item which was incorrect.

Many name changes in macros and function names. An unfortunate
complication with the jnamval -p and -P options is that since both are
for print types there is a possible confusion with names of macros as
there are two sets of JNAMVAL_PRINT_ macros but whether one of the sets
should be renamed is not yet determined. If they are the function names
might also want to change as both -p and -P have similar function names
for checking what options are used. Previously the -p option referred to
matching of name, value, both or json but it has to do with printing, 
not matching. The -t option is matching types. Maybe the -p needs to be
changed to a different letter but this can be determined in discussion.

Updated jnamval.1 man page.


xexyl:
I will be making a comment in the issue as there is a confusing thing with `-p` and `-P` that maybe can be remedied. Since they both refer to printing it can cause some confusion about what variables are for what options. I think this is unfortunate but since they both are printing related the names of the variables and functions are reasonable.

xexyl:
Please see [comment 1666514586](https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1666514586) for important points to consider with the problem that both `-p` and `-P` are specific to `jnamval` and both say what to print but in a different way. One is a style I guess and maybe it could be changed to refer to style, the `-p`? I'll update the comment and say that ... off to do other things now.

lcn2:
> Please see [comment 1666514586](https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1666514586) for important points to consider with the problem that both `-p` and `-P` are specific to `jnamval` and both say what to print but in a different way. One is a style I guess and maybe it could be changed to refer to style, the `-p`? I'll update the comment and say that ... off to do other things now.

We have some ideas and will discuss them in regards to [comment 1666514586](https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1666514586).

xexyl:
> > Please see [comment 1666514586](https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1666514586) for important points to consider with the problem that both `-p` and `-P` are specific to `jnamval` and both say what to print but in a different way. One is a style I guess and maybe it could be changed to refer to style, the `-p`? I'll update the comment and say that ... off to do other things now.
> 
> We have some ideas and will discuss them in regards to [comment 1666514586](https://github.com/ioccc-src/mkiocccentry/issues/523#issuecomment-1666514586).

Thanks .. will react to them tomorrow morning if I can't later on today - which is entirely possible.

PR #827, opened 2023-08-05
Fix mkiocccentry -y message

xexyl:
Missing ',' at end of lines in para() made for long lines and merged words.

Moved the call above the warning about ignoring warnings as that warning says no more warnings will be shown so the -y warning should be shown first.


xexyl:
Will do more tomorrow .. off to do other things now.

lcn2:
Good catch, @xexyl

xexyl:
> Good catch, @xexyl

Thanks .. can easily happen. It wasn't seen earlier because of all the text shown with `make test`.

PR #828, opened 2023-08-06
Update jfmt, jval and jnamval -F option

xexyl:
The option now looks like:

    -F fmt	Change the JSON format style (def: use default)

		tty	    When output is to a TTY, use colour, otherwise use simple
		simple	    Each line has one JSON level determined by '[]'s and '{}'s
		colour	    Simple plus ANSI colour syntax highlighting
		color	    Alias for colour (colour excluding you :-) )
		1line	    One line output
		nows	    One line output, no extra whitespace

Updated man pages for this option as well as a typo fix in jnamval.1 (an option letter was incorrect making it so that the example had no explanation).

Updated tmp/TODO.md - something that has to be done after the specs of these tools are finalised is to make sure common options are in the same order ahead of the other options both in usage strings and man pages.

The -F option is not yet added to the json_util_README.md file or if it is it's not updated. That will happen another time.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,34 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 1.0.50 2023-08-06
+
+New version of `jfmt`, `jval` and `jnamval`: `"0.0.12 2023-08-06"`. `jfmt` and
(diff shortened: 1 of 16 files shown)


lcn2:
thanks @xexyl

xexyl:
> thanks @xexyl

Happy to help.

PR #831, opened 2023-08-07
Change jnamval -P to -r, restrict types

xexyl:
This is because it can be confusing to have -p for printing and -P for printing especially with the names of macros used. The macros were renamed, the option was changed to -r and the usage string updated. Clearly 'p' had to come after 'r' so the usage string was changed in order as well.

The change has been updated in json_util_README.md which also did not even have the -P option (now -r).

The new version is 0.0.13 2023-08-07.


xexyl:
Off to do other things now .. will do more tomorrow. Good day!

PR #837, opened 2023-09-13
Add 'ne' to -S/-n help for jval/jnamval

xexyl:
Minor fix in jval and jnamval - new version 0.0.15 2023-09-13.

The 'ne' operator for -S and -n already was parsed (by foresight that it would be needed, maybe) but it was not in the help. Also I decided to change the order of the enum so that JSON_CMP_OP_NE comes right after JSON_CMP_OP_EQ rather than at the end.

Man pages updated - added 'ne' to -S and -n options.


xexyl:
This was simple enough to do I felt before 'I'm back'.

I'll update you on my condition next in the other thread and then be doing other things for the rest of the day.

PR #838, opened 2023-10-19
Make a kitchen! :-)

xexyl:


lcn2:
🚰👩‍🍳@xexyl

xexyl:
> 🚰👩‍🍳@xexyl

The only problem is it doesn't say what kind of soup. I should have added a type of soup. Still our inside joke never specified it either so it doesn't matter and perhaps that's how it should be - make it cryptic for the IOCCC. Yes I like that.

PR #840, opened 2023-11-06
Forbid certain filenames in submissions

xexyl:
The following files are no longer allowed:

- README.md
- prog.orig
- prog
- index.html
- inventory.html
- prog.orig.c

The mkiocccentry program will detect this early on and tell users this rather than making them input details about their entry only to have to start over simply because of an invalid filename.

The function test_extra_file() for chkentry has also been updated.

New version of mkiocccentry 1.0.2 2023-11-06. Updated json files for new version of mkiocccentry.

No extra test files were added but this might be a good idea to test the chkentry checks as well.


xexyl:
Hmm ... see a possible problem. Checking. I did run `make prep` and no problem but pulling now I see one. Will fix if necessary.

lcn2:
That was fast 💨 @xexyl thanks 🙏 

Too bad it has to be done. However we just know that someone might have submitted their source code as `README.md` and made it difficult for us to do a writeup of their win. 🥇 😉

xexyl:
Ah .. I forgot to run `make all` before the test it seems.

xexyl:
> That was fast 💨 @xexyl thanks 🙏
> 
Welcome.

> Too bad it has to be done. However we just know that someone might have submitted their source code as `README.md` and made it difficult for us to do a writeup of their win. 🥇 😉

Yes indeed though that would be kind of funny. Would require the `-x c` option too.

I'm working on the txzchk fix but I'm tired enough where I might have to delay that until tomorrow :(

lcn2:
> Ah .. I forgot to run `make all` before the test it seems. 

Feel free to revise with another pull request.  As this code has not been used, the version number can be the same, @xexyl

xexyl:
> > Ah .. I forgot to run `make all` before the test it seems.
> 
> Feel free to revise with another pull request. As this code has not been used, the version number can be the same, @xexyl

I ended up changing the version anyway. That would have been useful to not have to though. It was just forgetting to run `make all` that caused the problem though. Thanks.

lcn2:
> Yes indeed though that would be kind of funny. Would require the -x c option too.

C code as markdown could be fun for someone, just not as `README.md` 🤓

xexyl:
> > Yes indeed though that would be kind of funny. Would require the -x c option too.
> 
> C code as markdown could be fun for someone, just not as `README.md` 🤓

It could indeed! Or as html. Or as I know has been done graphics.

PR #841, opened 2023-11-06
Forbid certain filenames in tarballs (txzchk)

xexyl:


xexyl:
I forgot to add the list of files but since the previous commit lists files it probably can be discerned.

I think we might want to add some test cases for both this update and the one with mkiocccentry. However I'm too tired to do that today. If you wish to do that so I can focus on the other repo that'd be welcome. For the txzchk test script see the README.md file in that directory for how to install the error code files.

Otherwise I can get to it another time.

lcn2:
> I think we might want to add some test cases for both this update and the one with mkiocccentry. However I'm too tired to do that today

Adding those test cases sometime would be a good way to check if your new code works as you expect.

xexyl:
> > I think we might want to add some test cases for both this update and the one with mkiocccentry. However I'm too tired to do that today
> 
> Adding those test cases sometime would be a good way to check if your new code works as you expect.

Of course. I would normally do it already but I'm so tired right now that I can't do that much :(

It _should_ be fine but please remind me if I don't get to it because it would be good to have it. It's what I built the script for after all!

xexyl:
I'm doing the tarball one now .. will then commit and then that's it for the day. Perhaps you'd like to make one for chkentry ? I did test the mkiocccentry of course: if any of those files are specified then it exits early. So the only one that won't be done by me now is the chkentry one.

xexyl:
> I'm doing the tarball one now .. will then commit and then that's it for the day. Perhaps you'd like to make one for chkentry ? I did test the mkiocccentry of course: if any of those files are specified then it exits early. So the only one that won't be done by me now is the chkentry one.

Done: https://github.com/ioccc-src/mkiocccentry/pull/842/commits/5f7b1ea8d713bb4a89c784e83c642c412d2557e0.

PR #842, opened 2023-11-06
Add test case for forbidden tarball files

xexyl:
The new fake tarball has files with each new forbidden file:

- README.md
- index.html
- inventory.html
- prog
- prog.orig
- prog.orig.c


xexyl:
There you go! I must stop for the day now though. If you'd like to add a bad json file for chkentry that'd be helpful though it should be good. If not I'll correct it.

xexyl:
Okay did the chkentry one too. But you're welcome to split it into multiple files if you like: I had all the filenames listed in the same json file as I need to get going for the day.

PR #843, opened 2023-11-06
Add chkentry test case for invalid filenames

xexyl:


lcn2:
🧪 🤓 @xexyl

xexyl:
> 🧪 🤓 @xexyl

Yes. Testing is vitally important, even in chemistry. After all, don't we all know the famous rhyme that explains the importance of it?

Little Willie was a chemist,\
Little Willie is no more.\
What he thought was H<sub>2</sub>O\
Was H<sub>2</sub>SO<sub>4</sub>!

... poor little guy...

xexyl:
There's another great chemistry joke:
--
Two chemists walk into a bar/pub. The first one says: I'll have a glass of H<sub>2</sub>O. The second chemist says I'll have a glass of H<sub>2</sub>O too.

The second chemist dies.
--

Not that there aren't many other great chemistry jokes...

xexyl:
> 🧪 🤓 @xexyl

... and it's a great emoji: thanks for doing it! I didn't know they have a test tube one. It was clear to me it was that due to the nature of the pull request and having a love for chemistry: though not very knowledgeable any more. In high school after my year of it the next year I spent some time preparing the lab for that year's class. That was fun.

But that was a very long time ago now and I've forgotten much of it. I noticed they also have a syringe emoji which is great: 💉!

xexyl:
I have a minor improvement - an additional file that should be forbidden .. doing it now. Also improving fix to mkiocccentry: make use of the defines and not the literal strings of the filenames (the txzchk update used the macros).

xexyl:
Almost done .. will have extra json files to test and it will be more modular too plus simpler. No need to explicitly check each filename in a separate if only to print the macro of the filename: can just print the `char *` itself. This is done slightly differently in txzchk but both use the `test_extra_file()` function along with the posix safe plus one: that's used separately because of different error messages.

xexyl:
I am aware that this repo needs still the print_test.1 man page written. I started it a while back but I'll finish it another day: when we're back at this repo.

xexyl:
> Almost done .. will have extra json files to test and it will be more modular too plus simpler. No need to explicitly check each filename in a separate if only to print the macro of the filename: can just print the `char *` itself. This is done slightly differently in txzchk but both use the `test_extra_file()` function along with the posix safe plus one: that's used separately because of different error messages.

Oh dear .. that won't work for txzchk. Would need a separate function. That's because the tarball requires certain files that are not allowed as extra files but txzchk does not act on extra files. So I had to fix that. Will take a bit longer but almost there I think.

xexyl:
Nope .. there's something really funny going on. I'll undo the changes another time and forbid another file I thought of: the prog.alt (without the .c). I'll just add the json files for test cases and then the other changes can happen another time. Assuming that the really odd problem is resolved then. if not I'll have to do this all another time.

xexyl:
https://github.com/ioccc-src/mkiocccentry/pull/845/commits/2d4bb86ae27a3fa3c7679c38ae6360d7639d49b1 and https://github.com/ioccc-src/mkiocccentry/pull/845/commits/c6d71f3e8794799ba4d3da9220b0bade2026658a take care of these issues including the extra filename added to the list of disallowed filenames.

PR #845, opened 2023-11-07
Add bad manifest filenames tests

xexyl:
Each forbidden filename in the manifest now has a separate file to check
it including prog.alt. The -34.json file was also updated to have 
prog.alt as an extra filename though it might be better to also check 
this for other files too if this is not done (been too long and too many
other changes in the other repo to recall if this is done or not).

That means that under test_ioccc/test_JSON/info.json/bad individual 
files that have an extra file with the names of:

- README.md
- index.html
- inventory.html
- prog
- prog.alt
- prog.orig
- prog.orig.c

have been added. Unless other files have to be added later to the list 
of disallowed filenames or there needs to be a more specific check in 
mkiocccentry for filenames of the required files this should complete 
the disallowed and required filenames checks.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,29 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.0.55 2023-11-07
+
+Add another forbidden file name in submissions: `prog.alt` (prog.alt.c is
+allowed).
+
+Modularise the checks for invalid filenames in entries. For instance in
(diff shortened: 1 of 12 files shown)


lcn2:
Thanks @xexyl 

Going back to sleep 🛌 🥱 😴

xexyl:
> Thanks @xexyl
> 
> Going back to sleep 🛌 🥱 😴

Hope you got some sleep!

PR #847, opened 2023-11-25
Update .gitignore

xexyl:
The makefile.local file should not be just in / but rather anywhere as it can work for any other directory with a Makefile.

Added vim swap file globs.

Added an additional .DS_Store related glob.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 *.[ao]
 *.dSYM/
 *~
-.*.swp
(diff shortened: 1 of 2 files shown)


lcn2:
Fair enough 😄 @xexyl

xexyl:
> Fair enough 😄 @xexyl

I had those files in my fork so it seemed like a good idea.

PR #856, opened 2024-03-03
Fix jparse/test_jparse/Makefile

xexyl:


xexyl:
When I first did the other changes I forgot to pull. So I had to tar up the changes and then extract them. That way I didn't have to do it all over again. But that means the old file for print_test still existed. That probably is what happened here.

lcn2:
Trying ...

lcn2:
Looking good so far ... train still here .. they station forgot to pump the waste water and add new drinking water and there person who has access to that station went into town .. so we have more time :)

xexyl:
> Looking good so far ... train still here .. they station forgot to pump the waste water and add new drinking water and there person who has access to that station went into town .. so we have more time :)

That's convenient! Hopefully it works fine now. I forgot the issue of make targets and with all the other changes I must have missed the change in the Makefile.

Just reading a bit before bed. You know what I'm reading too! :-)

PR #858, opened 2024-03-05
Improve rebuilding txzchk test error files

xexyl:
The script test_ioccc/test_txzchk.sh now has a new option: -B. This rebuilds the test error files. The script prompts the user to verify that everything is correct. If it does not get a YES (case sensitive!) it will exit 3 (invalid command line). If this is the case the rest of the rule is not run so no notice about adding and committing the new or updated files.

The rule in the Makefile now is called (the previous one still exists though): rebuild_txzchk_test_errors.

The test_ioccc/test_txzchk/README.md file has been updated for this (there was also an error in it but that text has been removed as the make rule tells the user what to do which is what the incorrect text was only of course the text is now corrected).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -632,14 +632,6 @@ release: test_ioccc/prep.sh
 #
 rebuild_txzchk_test_errors force_expectations_for_txzchk_test:
 	@./test_ioccc/txzchk_test.sh -B || exit 1
-	@echo
-	@echo "Now run ./test_ioccc/txzchk_test.sh and verify it works!" 1>&2
-	@echo
-	@echo "Assuming all tests pass you must then do:" 1>&2
-	@echo
-	@echo "    git add test_ioccc/test_txzchk/bad/*.txt test_ioccc/test_txzchk/bad/*.err" 1>&2
-	@echo
-	@echo "and then commit them (and if necessary open a pull request)." 1>&2
 
 # make parser
 #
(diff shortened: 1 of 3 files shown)


xexyl:
I also noticed something: these test scripts do not have man pages. Would you like me to add these? I presume the priority is pretty low as it would delay work in the other repo.

I will look at the other comments here and in the other repo later on today .. I hope. if not then tomorrow I should be able to. For now though it's another break. Two days awake too early and I won't be able to rest again but hopefully will be a fine day.

Hope you're sleeping well / had a nice sleep!

xexyl:
> > I also noticed something: these test scripts do not have man pages. Would you like me to add these? I presume the priority is pretty low as it would delay work in the other repo.
> 
> 
> 
> Low priority add, yes.  Good idea @xexyl 
> 
Figured it would be low priority. 
> 
> 
> > I will look at the other comments here and in the other repo later on today .. I hope. if not then tomorrow I should be able to. For now though it's another break. Two days awake too early and I won't be able to rest again but hopefully will be a fine day.
> 
> > 
> 
> > Hope you're sleeping well / had a nice sleep!
> 
> 
> 
> Well thank you, hope you had one as well.

Woke up too early again but I am okay at least.

lcn2:
Thanks @xexyl  .. and it looks like the so-called GitHub checks are working again.

xexyl:
> Thanks @xexyl .. 

Welcome.

> and it looks like the so-called GitHub checks are working again.

... that's good!

PR #863, opened 2024-05-16
More fixes wrt terminology + update FAQ/man pages

xexyl:
This is a big update because as well as fixing some terms I updated the man pages (in the BUGS section, see below) and the FAQ (wrt terms but also briefly mentioning three files that seem like they are out of place so might need a bit of explaining though they should remain brief). At the same time I found a typo to do with the term 'entry'. In many files there was the comment for an include of:

    util - entry common utility functions for the IOCCC toolkit

but the word 'entry' is not needed anyway (I had noticed it before but it did not seem worth worrying about but since the term 'entry' now is being considered I went ahead and removed it). This means quite a few files have been modified, more than would otherwise be needed.

As for the three files in the FAQ they are in particular: soup/foo.h, soup/foo.c and soup/oebxergfB.h. They are out of place in that they are obfuscated and the rest of the repo is not (except perhaps for the parts of the JSON parser that are generated by tools :-) ). It seemed like they should be noted explicitly. These files should not be modified except by the judges or me as the code corresponds to a private tool I wrote at the request of Landon. As for oebxergfB.h that file is generated from the tool in its entirety. The text of the FAQ entries, however, can be improved certainly (though no more elaboration should be done, I think, as that would ruin the point of how the way the files are implemented), but these were done quickly.

The man pages for mkiocccentry and chkentry have had their BUGS section updated to note that just because a word exists in the code (or data) that seem to be incorrect terminology does not mean that it is. For instance if you look at the JSON semantics tables code you will see 'entries' and this has to do with table entries, not IOCCC entries. The word 'entries' also is in a comment in some or all of the three files above and they should not be updated either.

No code functionality was modified in this commit but nonetheless I ran make prep to make sure everything is okay and it is.


lcn2:
Thank you .. that was a useful cleanup @xexyl

xexyl:
> Thank you .. that was a useful cleanup @xexyl

You are most welcome.

PR #866, opened 2024-05-19
Fix chkentry.1 and mkiocccentry.1

xexyl:
Though intended well it was decided that certain references are perhaps not a good idea to have after all.


lcn2:
🙈👍@xexyl

PR #869, opened 2024-05-21
Add some credits to location code

xexyl:
To match what was done in chkentry I have also done it in the location tool code in the same way that was done in the chkentry code and header files.

Also in CHANGES.md I removed a reference to '@xexyl' (a fun pun is attributed to me still, however). This is because otherwise every time a release is made GitHub would tell me I was mentioned which is both confusing (as it's buried deeply) and unnecessary (plus if ever I am mentioned I might otherwise miss it). This problem is actually how I discovered the missing credits mentioned in this commit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -427,7 +427,7 @@ German Democratic Republic ==> DD
 Germany ==> DE
 ```
 
-I, that is Cody/@xexyl, observe that Germany is a country full of 'germ ridden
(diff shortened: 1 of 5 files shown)


xexyl:
I hope to return to the other repo tomorrow but I will be doing other things the rest of the day. If I don't return tomorrow I imagine the next day or if not then at the very latest the following day.

PR #870, opened 2024-05-22
Documentation fixes/additions for location(1)

xexyl:
Updated LOCATION_VERSION to "1.0.4 2024-05-22" from "1.0.3 2024-03-03".

Fixed usage string in location_main.c so that it shows that one can specify more than one location.

The location(1) tool was missing from the README.md and this seems like a good tool to have there. There might be other tools missing too but that can be determined and if necessary fixed another time.

The man page had a typo as well: in SYNOPSIS it had lookup when it would probably be more correct to say 'look up': the former is the facility and the retrieval of information but one would 'look up information' or in this case 'look up location information' or to be more specific 'look up ISO 3166 codes, location names by name (substring, full) or print the entire table' (this was expanded as well as it was missing some useful information). Also added some clarity to the -s and -a options in the DESCRIPTION section.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,18 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 1.1.1 2024-05-22
+
+Changed `LOCATION_VERSION` to `"1.0.4 2024-05-22"`.
+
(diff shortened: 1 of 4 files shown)


lcn2:
Thanks

xexyl:
> Thanks

You are most welcome!

PR #878, opened 2024-06-26
Update CHANGES.md and version of txzchk

xexyl:
The new version of txzchk is 1.0.4 2024-05-16. The adding #include <sys/wait.h> was necessary as in some systems like NetBSD the WEXISTATUS() does not exist without it which I had neglected by accident to address. Thanks to @SirWumpus this is now fixed.

In some old test files (that are supposed to fail) the txzchk_version had not been updated and this was not noticed until now because they were supposed to fail and unlike with txzchk test script that checks for specific errors (I had to do it this way for very specific reasons). The few references to an older version have been updated to the most recent version.

Ran make prep without any problems.


lcn2 (approved):
Thank you 🙏 @xexyl

xexyl:
> Thank you 🙏 @xexyl

Most welcome!

PR #883, opened 2024-06-27
Change C standard to c17

xexyl:
It was stated that this would happen sometime in 2023 and we're almost at July in 2024. One of the problem systems that required c11 will be reaching EOL (thanks to Red Hat) on 30 June 2024 so it seems to not be a problem. Rocky linux appears to work with c17 (although I only tested the top level Makefile for that but it should work fine).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,23 +145,21 @@ MAKE_CD_Q= --no-print-directory
 
 # C source standards being used
 #
-# This repo supports c11 and later.
+# This repo supports c17 and later.
 #
-# NOTE: The use of -std=gnu11 is because there are a few older systems
-#	in late 2021 that do not have compilers that (yet) support gnu17.
-#	While there may be even more out of date systems that do not
-#	support gnu11, we have to draw the line somewhere.
+# NOTE: at one point we used -std=gnu11 because there were a few older systems
+#       in late 2021 that did not have compilers that (yet) supported gnu17.
+#       While there may be even more out of date systems that do not support
+#       gnu11, we have to draw the line somewhere. Besides, one of those systems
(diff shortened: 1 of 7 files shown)


SirWumpus:
> Does this change cause a problem with NetBSD though?

I doubt it.  Haven't tested it yet though.  Need to check which is the first version of gcc that support `-std=c17`.  That'll be the kicker.

xexyl:
> > Does this change cause a problem with NetBSD though?
> 
> I doubt it. Haven't tested it yet though. Need to check which is the first version of gcc that support `-std=c17`. That'll be the kicker.

Supposedly it's: GCC 8.1.0 and LLVM Clang 7.0.0.

lcn2 (approved):
This change is consistent with the recent changes to [IOCCC guidelines](https://ioccc-src.github.io/temp-test-ioccc/next/guidelines.html). Thanks @xexyl

xexyl:
> This change is consistent with the recent changes to [IOCCC guidelines](https://ioccc-src.github.io/temp-test-ioccc/next/guidelines.html). Thanks @xexyl

You are most welcome. Have to be afk a while .. then have other things I must do after the electricians. I don't have to be there the whole time but I will possibly turn laptop off during their time here .. not sure.

xexyl:
> This change is consistent with the recent changes to [IOCCC guidelines](https://ioccc-src.github.io/temp-test-ioccc/next/guidelines.html). Thanks @xexyl

I didn't even see that - yet!

I was hoping I could look at those files today but I'm not sure I can unfortunately. I need to go afk for a while but I am far from certain if I can work on them today or if I can if it's a good idea. At least the power situation is worked out: unless of course the city finds a problem with the work. But I'm sure it's fine.

Issue #873, opened 2024-06-24
Bug report: iocccsize Abort trap: 6


lcn2:
## Describe the bug

The `iocccsize(1)` tool may crash when processing its own binary.

## To Reproduce

0. make iocccsize
1. ./ioccsize iocccsize

## Expected behavior

The `iocccsize(1)` will print an integer and warnings about rule violations.

## Relevant Output

```sh
$ ./iocccsize iocccsize
Abort trap: 6
$ ./iocccsize < iocccsize
Abort trap: 6
```

## bug report attachment

vvvvvvvvvvvvvvvvvvvvvvvvv --- Drag the file into the area below, please!

[bug-report.20240624.163152.txt](https://github.com/user-attachments/files/15962290/bug-report.20240624.163152.txt)

^^^^^^^^^^^^^^^^^^^^^^^^^ --- Drag the file into the area above, please!

## Please provide the following information:

 - Did you attach the above requested bug report file?  yes
 - What is the airspeed Velocity of an Unladen Swallow? unknown

## Additional context

n/a

lcn2:
FYI:

```
(lldb) run iocccsize
Process 84830 launched: '/Users/chongo/bench/ioccc/mkiocccentry/iocccsize' (arm64)
Process 84830 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x0000000192cfea60 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`:
->  0x192cfea60 <+8>:  b.lo   0x192cfea80               ; <+40>
    0x192cfea64 <+12>: pacibsp
    0x192cfea68 <+16>: stp    x29, x30, [sp, #-0x10]!
    0x192cfea6c <+20>: mov    x29, sp
Target 0: (iocccsize) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x0000000192cfea60 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x0000000192d36c20 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x0000000192c43ac4 libsystem_c.dylib`__abort + 136
    frame #3: 0x0000000192c3525c libsystem_c.dylib`__stack_chk_fail + 96
[...]

PR #885, opened 2024-06-27
GH-873 Fix potential keyword buffer overflow off by 1 byte causing stack corruption.

SirWumpus:
My initial attempt to address this issue.  I'm not yet able to duplicate the failure.  This change still makes sense regardless so worth merging and testing on those platforms with compiler options that failed.

### DIFF
--- a/soup/rule_count.c
+++ b/soup/rule_count.c
@@ -243,7 +243,7 @@ RuleCount
 rule_count(FILE *fp_in)
 {
 	size_t wordi = 0;
-	char word[WORD_BUFFER_SIZE+1];
+	char word[WORD_BUFFER_SIZE];
 	RuleCount counts = { 0, 0, 0, false, false, false, false, false };
 	int ch, next_ch, quote = NO_STRING, escape = 0, is_comment = NO_COMMENT;
 
@@ -431,11 +431,20 @@ rule_count(FILE *fp_in)
 
 		/* Collect next word not in a string or comment. */
 		if (IS_CODE && (isalnum(ch) || ch == '_' || ch == '#')) {
+			word[wordi++] = (char) ch;
 			if (sizeof (word) <= wordi) {
+				/* ISO C11 section 5.2.4.1 Translation limits, identifiers
+				 * can have 63 significant initial characters, which can be
+				 * multibyte.  The C keywords are all ASCII, longest is 14
+				 * bytes.
+				 *
+				 * We only care about the C keywords and not the identifiers,
+				 * so the buffer can overflow regularly as long words or
+				 * identifiers are ignored.
+				 */
 				counts.wordbuf_warning = true;
 				wordi = 0;
 			}
-			word[wordi++] = (char) ch;
 			word[wordi] = '\0';
 		}
 
--- a/iocccsize.h
+++ b/iocccsize.h
@@ -41,7 +41,7 @@ typedef unsigned char bool;
 #endif
 
 #ifndef WORD_BUFFER_SIZE
-#define WORD_BUFFER_SIZE	256
+#define WORD_BUFFER_SIZE	16
 #endif
 
 /*
--- a/test_ioccc/iocccsize_test.sh
+++ b/test_ioccc/iocccsize_test.sh
@@ -7,6 +7,10 @@
 # Public Domain 1992, 2015, 2019 by Anthony Howe.  All rights released.
 #
 # IOCCC mods in 2019,2021,2022 by chongo (Landon Curt Noll) ^oo^
+#
+# ***
+# *** NOTE when editing this file it contains UTF8 multibyte characters.
+# ***
 
 # setup
 #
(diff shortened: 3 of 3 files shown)


xexyl:
I am not saying you're wrong btw: just wanted to point out that once there were literal characters and it became a problem. 

Out of curiosity and **merely curiosity** is there a reason why you want it to be the literal characters?

SirWumpus:
The \x codes originally used for the utf8.c test are NOT correct UTF8 multibyte values for those characters, which were mb2 codes.  The new extra tests use Japanese which are mb3 characters.  The whole point is to test with UTF8 files, if you convert the characters in strings to \x, then you defeat the point of the test.

The new tests have C identifiers with multibyte characters, which C11 allows.  As mentioned in the comments for iocccsize, a developer could write native language comments or strings or identifiers (eww), so we need to test that behaviour.  BTW the hello-jp*.c files do successfully compile; I checked to be sure.

SirWumpus:
> ... it was broken in some systems

On those systems, was the environment set to use UT8, eg. `LANG=en_CA.UTF-8`?  I had issues with Unicode until I made sure the default `LANG` allowed UTF8.

xexyl:
> The \x codes originally used for the utf8.c test are NOT correct UTF8 multibyte values for those characters, which were mb2 codes. The new extra tests use Japanese which are mb3 characters. The whole point is to test with UTF8 files, if you convert the characters in strings to \x, then you defeat the point of the test.
>  

That's fair enough: like I said i was not saying you were wrong - I just wanted to point out that literal characters did prove a problem. But see below as I don't think it was this script anyway. Though shellcheck in some systems had a problem. I imagine that the rocky linux version does not though and even if it does it's not the best example system to test on.

> The new tests have C identifiers with multibyte characters, which C11 allows. As mentioned in the comments for iocccsize, a developer could write native language comments or strings or identifiers (eww), so we need to test that behaviour. BTW the hello-jp*.c files do  successfully compile; I checked to be sure.

It wasn't about the compilation of the files. It was the scripts. BUT I'm pretty sure it was the mkiocccentry test script anyway and I'm sure it's fine. Even if the tests cause problems in rocky linux (if they do - not sure yet) it's as noted above, not a good test case anyway.

Above all I was just merely curious in the rationality. I hope you know that but in case you did not I hope you do now.

I hope also that this commit fixes the problem! Then you can relax about it and the bug will be fixed too. I don't remember which system caused a problem and I'm not sure what optimisation levels were a problem so I can't test that.

xexyl:
> > ... it was broken in some systems
> 
> On those systems, was the environment set to use UT8, eg. `LANG=en_CA.UTF-8`? I had issues with Unicode until I made sure the default `LANG` allowed UTF8.

Can't say now .. thanks to Red Hat :( .. but I don't mind much now since rocky linux works well and has been up since last year. Anaconda installer is annoying with md+raid but I got it working. Now if I could remember how I did that exactly .. would probably just reinstall with fedora: new drive anyway (bad sectors though I noticed that apparently some of the other disks also seem to have a bad sector). Everything's on backup anyway.

Anyway I'm sure it's fine. And again not saying you're wrong - I just wanted to say there was at one point a problem. Besides you said those characters were not accurate translations so that's a problem too. Was just wondering.

I hope this fixes the problem so you can not worry about it - I know you've been thinking about it and pondering it for a couple days now and you deserve a break!

soup/rule_count.c
```diff
@@ -243,7 +243,7 @@ RuleCount
 rule_count(FILE *fp_in)
 {
 	size_t wordi = 0;
-	char word[WORD_BUFFER_SIZE+1];
```
lcn2, line 246: Not sure that the removal of this paranoia byte is a good idea, but we are willing to give it a try

lcn2 (approved):
Let's give this a try @SirWumpus

soup/rule_count.c
SirWumpus: The paranoia byte doesn't help you since the test is against the size of the word buffer, ie. `sizeof (word) <= wordi` not the macro constant.  The byte would've help if `WORD_BUFFER_SIZE <= wordi` had been the expression.

xexyl:
Not to cause you more grief, @SirWumpus, but an interesting thing I just noticed with `iocccsize`. If you run it under valgrind it seems to just stop. Haven't tried it with it on a non-binary file but running:

```sh
$ valgrind ./iocccsize  iocccsize
==1815818== Memcheck, a memory error detector
==1815818== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==1815818== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==1815818== Command: ./iocccsize iocccsize
==1815818== 
^C^C^C^C
Killed
```

.. and the killed, as I am sure you see, was because I `kill -9` on the PID in another terminal.

Not that I think it should matter much but since we're talking about memory faults (or something along those lines if not) ...

PR #886, opened 2024-06-27
Fix limit_ioccc.sh not being generated at times

xexyl:
If one were to do make clobber iocccsize it would not build soup/limit_ioccc.sh. It would, however, build the soup.a library, and since since it's a useful file to be built, and since tools rely on soup.a, the soup.a rule now makes limit_ioccc.sh.

Doing it this way prevents having to add something like:

	${MAKE} -C soup limit_ioccc.sh

to all the rules in the Makefile and now one can do:

	make clobber iocccsize
	./test_ioccc/iocccsize_test.sh

and have it run okay. Without something like this (that can probably be improved upon / cleaned up) one would have to build everything in order for the file to be built.

### DIFF
--- a/soup/Makefile
+++ b/soup/Makefile
@@ -390,6 +390,7 @@ soup.a: ${LIB_OBJS}
 	${RM} -f $@
 	${AR} -r -u -v $@ $^
 	${RANLIB} $@
+	${MAKE} limit_ioccc.sh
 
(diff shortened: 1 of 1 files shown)


xexyl:
Feel free to clean up or suggest another place this can be done if you wish. .. though I will be away for now and not sure how much more I will be able to do anything here at all today.

xexyl:
Closing this though .. I hate typos and I have to fix it. One moment. Yes I know it's funny.

PR #887, opened 2024-06-27
Fix limit_ioccc.sh not being generated at times

xexyl:
If one were to do make clobber iocccsize it would not build soup/limit_ioccc.sh. It would, however, build the soup.a library, and since it's a useful file to be built, and since tools rely on soup.a, the soup.a rule now makes limit_ioccc.sh.

Doing it this way prevents having to add something like:

		${MAKE} -C soup limit_ioccc.sh

to all the rules in the Makefile and now one can do:

		make clobber iocccsize
		./test_ioccc/iocccsize_test.sh

and have it run okay. Without something like this (that can probably be improved upon / cleaned up) one would have to build everything in order for the file to be built (which maybe indicates a better procedure could be devised).

### DIFF
--- a/soup/Makefile
+++ b/soup/Makefile
@@ -390,6 +390,7 @@ soup.a: ${LIB_OBJS}
 	${RM} -f $@
 	${AR} -r -u -v $@ $^
 	${RANLIB} $@
+	${MAKE} limit_ioccc.sh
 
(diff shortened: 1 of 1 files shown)


xexyl:
Again feel free to clean up or suggest another way to go about it. It does make me wonder that since it previously would only be built if one built everything that maybe it could be devised better. How that might be though I don't have an idea at this time.

I don't know that I like it at the soup.a library but since these tools rely on soup.a and thus will be building them and since it's in soup it seems like a place it can be.

lcn2 (approved):
🙏 @xexyl

PR #888, opened 2024-06-27
GH-873 Fix formatting of memset() commit 6b27db.

SirWumpus:


xexyl:
You know this might seem ridiculous to some - and funny (both haha funny and otherwise) - but I totally get it. I would have to do the same thing. It looks nicer now.

lcn2 (commented):
Thanks 🙏 @SirWumpus

Issue #894, opened 2024-07-03
`make prep` does not handle NetBSD `gmake`


SirWumpus:
1. The `gmake prep` generates errors WRT the use of `make`.  It appears the `test_ioccc/prep.sh` is not passed `-m${MAKE}` option and assumes `make`, instead of `$MAKE`.

2. With a some changes (I will push) it will run to completion, but reports these errors
One or more tests failed:

        make_action 25: gmake -f ./Makefile shellcheck: non-zero exit code: 2
        make_action 26: gmake -f ./Makefile seqcexit: non-zero exit code: 2
        make_action 27: gmake -f ./Makefile picky: non-zero exit code: 2

Where the 3 above tools are not installed.

@landon please assign to me as this is a subtask of #893.

xexyl:
Looks like you don't have those tools installed. 

Do you? If not that's probably why. But I can't check for the error codes until tomorrow .. bedtime. 

The Makefile will tell you where to get those tools btw.

xexyl:
Ah you said that already. 

So that's another concern of whether it will work on a workflow. But I don't know. 

I can look at it more tomorrow. If you give more details on the problem maybe that will help too. 

But maybe that's all nonsense: can barely see. Anyway fwiw the Makefile tells where to get the tools. 

Some of them might not be strictly necessary but I don't know. 

Anyway good night!

lcn2:
BTW: Platforms such as **NetBSD** are not the target platforms for `make prep` nor `make release`.  We do **NOT** possess those platforms, nor do we plan to acquire them to be able to support them under  `make prep` nor `make release`.

We have nothing against **NetBSD**, nor **OpenBSD**, nor classic BSD.  As mentioned before, Landon was an early BSD developer when BSD was at Berkeley.

**NOTE**: This does not mean we don't want `make clobber all` to work on as many platforms that is reasonable.  We would want `make clobber all` to work on **NetBSD**, should doing so not negatively impact other platforms.  

If you wish to consider creating a pull request that would help **NetBSD**, and that does not negatively impact our existing support platforms, feel free to suggest one.

lcn2:
One approach we may wish to consider is to better specify the system requirements for running `make prep` and `make release`.  Perhaps in an FAQ?

While there may be workarounds for systems with a shell or make that is less functional than we require: there is also a limit to how far we need to go to accommodate various platforms.

[...]

PR #895, opened 2024-07-03
GH-893 Fix root Makefile to pass`-m${MAKE}` to `prep.sh`; fix

SirWumpus:
`prep.sh` to reference `$MAKE` in place of `make`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: C/C++ CI
 
 on:
   push:
-    branches: [ "master", "gh_893" ]
+    branches: [ "master" ]
   pull_request:
     branches: [ "master" ]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: make
+      run: make
+    - name: make prep
+      run: make prep
+    - name: make test
+      run: make test
+    - name: make clobber
+      run: make clobber
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,14 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+    - name: install shellcheck
+      run: yum install shellcheck
+    - name: install seqcexit
+      run: git clone https://github.com/lcn2/seqcexit
+      run: cd seqcexit && make install
+    - name: install picky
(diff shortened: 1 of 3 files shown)


xexyl:
FWIW: I tested this and it works ... but is it necessary to do `export` on the `MAKE` variable in the parsing of options? It doesn't matter really but it is kind of inconsistent with the rest of the script (and `MAKE` is already exported).

What did it solve on your end btw? (Since we haven't had the problem.) I guess it's so you can use gmake on your system ?

xexyl:
Ah .. I see that is the case, for NetBSD or other systems which have to rely on `gmake`.

xexyl:
Anyway works well (as expected but thought it might be nice to have confirmation in macOS also .. sure it's fine in linux also).

xexyl:
I just did another couple tests, fwiw:

```sh
make MAKE=/opt/local/libexec/gnubin/make prep
```

and also:

```sh
./test_ioccc/prep.sh -m /opt/local/libexec/gnubin/make 
```

though obviously the former is the preferred way as it's cleaner: no flooding the screen of compilation etc. and you get a nice summary (that was something I was asked to do and I'm glad I was: it's MUCH nicer and much more helpful!).

I would add a space after the `-m` of course to be consistent with the rest of the command but that is less important (though I might do it anyway :-) ).

#894

xexyl:
.. oddly adding `-m ${MAKE}` to the bug_report rules does not however work ... even though I added the option `-m` recently. But since I made the script (thanks to @SirWumpus) have the search for `gmake` first that should not be a problem .. I think.

lcn2 (commented):
Thank you @SirWumpus for this pull request.

We recommend that we try this and see how it goes.

We can experiment with forcing a "deliberate failure" and see how it goes,
before reverting the that change and restoring to working order.

On a slightly related matter,
see [comment-2206988540](https://github.com/ioccc-src/temp-test-ioccc/issues/2537#issuecomment-2206988540),
and see [comment-2206996221](https://github.com/ioccc-src/temp-test-ioccc/issues/2537#issuecomment-2206996221).

We suggest we try this one in this repo and given the results think about issue #2537 for the other repo.

xexyl:
> Thank you @SirWumpus for this pull request.
> 
> We recommend that we try this and see how it goes.
> 
> We can experiment with forcing a "deliberate failure" and see how it goes, before reverting the that change and restoring to working order.
> 
> On a slightly related matter, see [comment-2206988540](https://github.com/ioccc-src/temp-test-ioccc/issues/2537#issuecomment-2206988540), and see [comment-2206996221](https://github.com/ioccc-src/temp-test-ioccc/issues/2537#issuecomment-2206996221).
> 
> We suggest we try this one in this repo and given the results think about issue #2537 for the other repo.

I was going to suggest just making a quick formatting change (extra line or something) for testing it but I'm glad you thought of the other way, a failure. That does seem like a good idea just to make sure it doesn't cause a problem.

Thanks @SirWumpus for the work!

PR #902, opened 2024-07-10
Bug fix txzchk -q option

xexyl:
Although it was quiet in some cases it still showed warnings. That was a problem even if it wasn't for the fact that showing output with quiet mode requested makes no sense.

An important point, however, is that if -w, -q will now be disabled. That option is important for the test suite and that always applied but since -q did not block warnings it was not detected.

I (re)discovered these problems (I had known about it previously but due to the test suite needing warnings it was not fixed) when working on the guidelines as I know txzchk best and there were some things that I felt could be improved or better explained.

Updated TXZCHK_VERSION to "1.0.7 2024-07-10".
Updated TXZCHK_TEST_VERSION to "1.0.3 2024-07-10".

The txzchk_version in all the test json files have been updated (at least those which are supposed to have the current version, which might or might not be all of them).

Ran make release to rebuild parser (though that was not necessary as jparse was not touched) and to test everything.


xexyl:
Now I can fix the guidelines.

lcn2 (changes requested):
We see the point being made in this pull request, however ...

The problem with `txzchk` may be that while it prints fun stuff, and such fun stuff gets in the way of it being used as a tool that only reported problems, or in the face of nothing wrong, **prints nothing**.

### Begin historical/histericl side note

The `-q` reminded us of a P1003.2 POSIX ballot objection the standard committee received to the description of the `grep` command where the person wanted to `grep(1)` to write to stderr when grep failed to find a match, something along the following line:

```
The following pattern was not found: __pattern__
```

It took a great deal of argument to prevent that change being forced onto the POSIX standard, and forcing people to use `grep 2>/dev/null` to try to get close to the `grep(1)` we know and use today!

### End historical/histericl side note

In our opinion, `txzchk`'s job is mainly to report on any problems found in a compressed submission tarball.  In the face of nothing wrong, print nothing.  Let the exit code (exit 0 or exit non-zero) present information to a script.  It should have been a "print out any minor problems and any serious problems" and in a "_grep-like_" way, if nothing is wrong, to print nothing.

In our opinion, the fact that `-q` was needed at all is the perhaps an mis-feature that could be addressed in a better way.  In our opinion, code such as the following, is problematic:

```c
...

    /*
     * environment sanity checks
     */
    if (!quiet) {
	para("", "Performing sanity checks on your environment ...", NULL);
    }

    txzchk_sanity_chks(tar, fnamchk);
    if (!quiet) {
	para("... environment looks OK", NULL);
    }

    /*
     * check the tarball
     */
    if (!quiet) {
	para("", "Performing checks on tarball ...", NULL);
    }

    tarball.total_feathers = check_tarball(tar, fnamchk);
    if (!quiet && !tarball.total_feathers) {
	para("No feathers stuck in tarball.", NULL);
    }
...
```

BTW: The above code fragment is not intended to be an exhaustive list of such code.  We did not do a complete scan of the source code.

Messages, however fun they might seem, such as "Performing sanity checks on your environment ..." get in the way of how the tool will be used: **a means to report minor and major problems with a submission's compressed tarball".  In the submit server, those messages have to be filtered out so that they don't turn up complicating the feedback given to the submitter.  Messages such as "No feathers stuck in tarball." should only be printed under a new `-e` flag:

```
    -e     Print entertaining and fun messages (def: do not)
```

BTW: The `-e` is just a guess for an option letter.  See **UODATE 0** below.  For now, think of `-e` as an "_example_".

The server shouldn't have to use `-q` to stop such messages from being issued back to the submit server uploader.

If a tool such as `mkioccentry` wants such messages to show up in the output (and here fun messages about feathers are fine), then the `mkicccentry` should launch `txzchk -e`.  

Returning to the example code fragment above, perhaps the code should be be rewritten long these lines:


```c

...

    /*
     * environment sanity checks
     */
    if (entertain) {
	para("", "Performing sanity checks on your environment ...", NULL);
    }

    txzchk_sanity_chks(tar, fnamchk);
    if (entertain) {
	para("... environment looks OK", NULL);
    }

    /*
     * check the tarball
     */
    if (entertain) {
	para("", "Performing checks on tarball ...", NULL);
    }

    tarball.total_feathers = check_tarball(tar, fnamchk);
    if (entertain) {
        if (!tarball.total_feathers) {
	    para("No feathers stuck in tarball.", NULL);
        }
    }
...

```

The so-called `-e` would change the `entertain` boolean from a default of `false` to `true`.

The `mkiocccentry(1)` is fundamentally a human interactive command.  Yes, you added a very nice feature that allows `mkiocccentry(1)` to be driven as a batch / non-interactive command.  That mode is very helpful for the test suite. 

When `mkiocccentry` calls `txzchk`, it should call it with the `-e` option.

Perhaps the `-q` for `txzchk(1)` should be deprecated or do nothing or removed from the command line?

We are going to let this pull request sit in the queue a bit longer: neither accepted nor rejected.   Perhaps the above comments could be considered and potentially implemented (and thus the pull request might amended).

**NOTE**: The above paragraph means that subsequent pull requests for this repo that assume this pull request will be held up until this pull request is accepted, accepted as amended, or rejected.  That shouldn't be much of a problem as this repo is in "_sort of a code freeze-like mode_" anyway.

## UPDATE 0

The choice of `-e` above was just a wild guess at a new flag, and is based on the word "entertainment".  If the `-e` is a poor choice due to other command options in this repo, then by all means please select a **BETTER OPTION** letter!

**ALSO NOTE**: The use of the term "entertainment" is **NOT** intended to disparage the fun messages that were added to `txzchk(1)` such as puns about feathers.   We find those messages fun and hope they stay in the code.   We just want them to be off by default when `txzchk(1)` is used on the command line unless someone gives the `-e` (or whatever we call the option).

We also suggest that the `txzchk(1)` man page **emphatically** recommend that the `-e`  be used by people so as to not **miss the punny entertainment**.  :-)    And as mentioned above, when the `mkiocccentry` command calls `txzchk(1)` it should call it with that flag so that humans see the punny messages.

lcn2:
In regards to [comment-2221034715](https://github.com/ioccc-src/temp-test-ioccc/issues/5#issuecomment-2221034715):

> Something I just noticed (when looking at the code to update the guidelines) is this:
>

```c
    if (p == NULL) {
        warn("txzchk", "%s: NULL pointer encountered trying to parse line", tarball_path);
        msg("skipping to next line");
        ++tarball.total_feathers;
        return;
    }

```

> The reason this was deemed not an error but a warning (and still an issue) is to list every problem found. But the real question is: am I right in saying that you would prefer that in the case quiet is true that the `warn()` and `msg()` functions are not called? Instead just the total feathers would be increased.
>
> Please let me know and I'll look through the code for other cases like this and fix them all.
>
> There are quite a few cases of this btw.

You raise some valid points in the above.  

**BTW**: The question of what should be an error and what should be a warning is a valid question **AND** is somewhat independent if the `-q` vs. `-e` issue raised in [https://github.com/ioccc-src/mkiocccentry/pull/902#pullrequestreview-2171162232](https://github.com/ioccc-src/mkiocccentry/pull/902#https://github.com/ioccc-src/mkiocccentry/pull/902#pullrequestreview-2171162232).  That is, we need to consider the issues of what is a true `warn()` call **AND** we need to address the quiet vs entertainment messages printing defaults.

For those problems that are true problems, they need to be counted and printed accordingly.  For those things are are not true problems: where `msg()` might be called, the call to printing such messages might be best of they are muted by default unless `-e` is given.

xexyl:
As for the `quiet` and showing checking sanity checks etc.: that was because the other tools have that very option as well.

I can indeed change it to your `-e` idea: I like that. But there is something you're forgetting too.

The tool has to print something at least part of the time: `mkiocccentry(1)` **USES** `txzchk(1)` to print out the contents of the tarball. Without `-q` it is shown and with it it is not. So it's not necessarily good to get rid of `-q` outright.

Still some of it could be changed to be shown with `-e`. What I'll do when more awake is change all that can be changed to use that option and the rest not.

As for the version update the change is that it no longer shows warnings if quiet.

xexyl:
Another reason for quiet:

```c
    /*
     * filename must use only POSIX portable filename and + chars plus /
     */     
    if (!posix_plus_safe(file->filename, false, true, false)) {
        ++tarball.total_feathers; /* report it once and consider it only one feather */
        ++tarball.unsafe_chars;
        if (!quiet) {
            warn(__func__, "%s: file does not match regexp ^[/0-9a-z][/0-9a-z._+-]*$: %s",
                tarball_path, file->filename);
        }
    }

```

When should it show the warning the `quiet` option is removed? The user would not know the problem if the quiet option is removed and the message is never shown.

.. and of course I certainly will keep those fun messages in but that's another matter entirely. This comment has a question that has to be answered before I can resolve this.

xexyl:
(Still waking up ..) Looking at your comment it might seem like you wish in those cases for `msg()` to be used instead. Is that so? That can be done but even so `-q` has to stay in here as I said as `mkiocccentry(1)` uses `txzchk(1)` to print the contents of the tarball.

So there are some problems with this request as I see it (see it meaning in my still sleepy eyes). I'll try reading again in a while but I'm not sure if that'll help or not. It might be that I have to wait until you get back to me in which case, depending on when you do, it might not be addressed until tomorrow.

xexyl:
Oh one more thing for now. I actually would like to get rid of the welcome message (though now that I think about it with an `-e` flag it could be modified to have even more fun) in general. The only reason I added it is because you have the same thing in mkiocccentry (for example).

With `-e` I could make it something much more fun, though. But what mkiocccentry would have I do not know (I like the idea that mkiocccentry would call `txzchk` with `-e` though).

xexyl:
> In regards to [comment-2221034715](https://github.com/ioccc-src/temp-test-ioccc/issues/5#issuecomment-2221034715):
> 
> > Something I just noticed (when looking at the code to update the guidelines) is this:
> 
> ```c
>     if (p == NULL) {
>         warn("txzchk", "%s: NULL pointer encountered trying to parse line", tarball_path);
>         msg("skipping to next line");
>         ++tarball.total_feathers;
>         return;
>     }
> ```
> 
>  > The reason this was deemed not an error but a warning (and still an issue) is to list every problem found. But the real question is: am I right in saying that you would prefer that in the case quiet is true that the `warn()` and `msg()` functions are not called? Instead just the total feathers would be increased.
> > Please let me know and I'll look through the code for other cases like this and fix them all.
> > There are quite a few cases of this btw.
> 
> You raise some valid points in the above.
> 
> **BTW**: The question of when and how errors and warnings should be printed is a valid question that should be addressed **AND** is somewhat independent if the `-q` vs. `-e` (or whatever flag letter we choose to add as the "_punny entrainment_" flag) issue raised in [pullrequestreview-2171162232](https://github.com/ioccc-src/mkiocccentry/pull/902#https://github.com/ioccc-src/mkiocccentry/pull/902#pullrequestreview-2171162232). For example, we need to consider **BOTH** the issues when to call `warn()` **AND** we need to address the quiet vs entertainment messages printing defaults.
> 
> For those problems that are true problems (errors and warnings), they need to be counted and printed accordingly. For those things are are not true problems: where `msg()` might be called, the call to printing such messages might be best of they are muted by default unless the so-called "punny entrainment flag" is given.
> 

From the above paragraph: it sounds like you think it should always print warnings / issues with the tarball. But that then means that if there are problems one cannot simply check the exit code. A recent issue you had me make it so that it works this way even. So this seems to be a change in what you wanted or needed. 

Of course the message of a NULL pointer (on a line) could always be printed rather than just skipped (silently) but the other ones it remains unclear what to do as it is contrary to what you recently had me do.

It is possible that the showing of the tarball contents could be changed to not be shown unless `-e` is used but it seems like it's not the best option: how is it entertaining? I don't know.

xexyl:
> The server shouldn't have to use -q to stop such messages from being issued back to the submit server uploader.

Perhaps it should always show the tarball contents too, to verify that the user also sees what is being uploaded? I think that might be a good idea. But it still might be nice to have the ability to silence that.

It might be that the contents could be framed in an entertaining way which would mean quiet is not needed. I already changed some of the other (the welcome ones) messages to be more entertaining (which do not worry I did not take that as disparaging: they were meant to be entertaining and I know you laughed at them too). I wonder ...

Pondering this .. if I could solve this then the only question would be how or when to show problems in the tarball, whether it should always do this or not.

SirWumpus:
As for an option name suggested by @lcn2 for extra output, eg. `-e` I would argue `-vv` (very verbose) (pseudo) diagnostic might better be suited.  I agree with @lcn2 about how the tool should be silent on success and useful unburden diagnostic output otherwise.  Verbosity levels might be a better way.  Sane and boring being the default, -vv, -vvv getting more bizarre.  However I would suggest reserving -v for sane detailed diagnostic messages, before having fun.  `iocccsize` supports a `-v level` (instead of multiple -vvv's) mainly for my own debugging.  However I could in future make, eg. -v 99, inject odd ball comments.

xexyl:
> As for an option name suggested by @lcn2 for extra output, eg. `-e` I would argue `-vv` (very verbose) (pseudo) diagnostic might better be suited. I agree with @lcn2 about how the tool should be silent on success and useful unburden diagnostic output otherwise. Verbosity levels might be a better way. Sane and boring being the default, -vv, -vvv getting more bizarre. However I would suggest reserving -v for sane detailed diagnostic messages, before having fun. `iocccsize` supports a `-v level` (instead of multiple -vvv's) mainly for my own debugging. However I could in future make, eg. -v 99, inject odd ball comments.

This would complicate parsing esp when it already requires a level (and so do all the other tools).

I'm actually working this out .. just one thing (I think) has to be resolved. Anyway verbosity and the `-e` option is not at all related. The use of `-q` is actually good, as it stands, as it does allow one to not show anything. HOWEVER, if it is by default quiet it could be better: perhaps `quiet` is true by default but if `-e` is used it disables quiet mode (and enables the puns). The only other question is when to show warnings or issues: it seems like an issue with the tarball is not necessarily something to show for entertainment but one might not always wish to see them either. 

That could be for another option: perhaps verbosity even. That is actually even suggested in the test suite.

Hmm ... these things might work ... thanks for making me think it even more!

xexyl:
Ah yes .. I already have an option to always show warnings: it has to be there too, for the test suite. So I can make it quiet by default but if the person needs to see the warnings they can use that option. The `-e` option will allow for fun too. I thin that'll work.

SirWumpus:
As much as we enjoy puns, gibberish, humour, anecdotal stories, etc. in these discussions, sometimes they get in the way of `The Signal`.  The `mkiocccentry` tool suite should first and foremost be about the signal and less noise.

I don't know about yourself, but I recall many a time when submitting entries how stressful the process could actually be, from getting an accurate size count to actually pressing "send" (and hope the mail didn't bounce).  Adding superfluous output clouds the signal at a critical point.  

`gcc` error and warning messages have become multiline essays on what is wrong with some code; there used to be a time you could count errors and warnings by single lines.  And `gcc` is trying to be helpful!  Imagine if `gcc` interjected random jokes and jibes between the error and warning messages how hard it would be to actually fix something.

Another point to make about superfluous messages is you have to maintain the code going forward.

xexyl:
> As much as we enjoy puns, gibberish, humour, anecdotal stories, etc. in these discussions, sometimes they get in the way of `The Signal`. The `mkiocccentry` tool suite should first and foremost be about the signal and less noise.
> 

And that's what it does. There won't be much .. and the only important message (which is probably what you would see) from txzchk, assuming all is good, is that you see the contents of the tarball: that has to be done.

> I don't know about yourself, but I recall many a time when submitting entries how stressful the process could actually be, from getting an accurate size count to actually pressing "send" (and hope the mail didn't bounce). Adding superfluous output clouds the signal at a critical point.

It was not for me no.

> 
> `gcc` error and warning messages have become multiline essays on what is wrong with some code; there used to be a time you could count errors and warnings by single lines. And `gcc` is trying to be helpful! Imagine if `gcc` interjected random jokes and jibes between the error and warning messages how hard it would be to actually fix something.
> 

That's another matter entirely. I particularly find the 'misleading indentation' one annoying and pointless; it's nothing but style policing at best.

> Another point to make about superfluous messages is you have to maintain the code going forward.

That's not a problem at all and not related to the above. Keep in mind that these will only be shown if the option to show them is used; it won't be on by default.

xexyl:
As far as whether to keep the `quiet` boolean in. I will keep it in but it defaults to true. The only way to disable it is to enable showing warnings (`-w`). I could remove the boolean outright and instead just check for `show_warnings` (and I might do that anyway) but it's the same effect now.

The other option (and something else fun) to entertain has been played with some (and more work has to be done) and will provide a bit more enjoyment, at least for those who wish to have it. 

I will have to update the man page of course and the guidelines but I can do all this later on ... need to break away for a bit I think.

xexyl:
> As far as whether to keep the `quiet` boolean in. I will keep it in but it defaults to true. The only way to disable it is to enable showing warnings (`-w`). I could remove the boolean outright and instead just check for `show_warnings` (and I might do that anyway) but it's the same effect now.
> 
> The other option (and something else fun) to entertain has been played with some (and more work has to be done) and will provide a bit more enjoyment, at least for those who wish to have it.
> 
> I will have to update the man page of course and the guidelines but I can do all this later on ... need to break away for a bit I think.

... actually I can get rid of it. In one case it will be based on verbosity level and in the others whether or not the `-w` option is used.

xexyl:
I have it done except the man page needs to be updated .. that I will do later. Must leave for a bit. But to give a summary (quick one):

- Removed `-q`: it's always quiet and shows no warnings unless `-w` is used (or a command line error or some error condition of course). Unsure if `mkiocccentry` should use `-w` if it doesn't but I guess it should? Yes I think it better so I'll update that.
- Added the `-e` option with another one that I'll let you discover (it is only used if `-e` is used).

But I found just now a strange problem. Might need to have a call to `fflush(stdout);` somewhere added. Will look at it when I'm back.

xexyl:
Actually there IS another bug here ... I'll report it as an issue later on. As for txzchk showing to mkiocccentry the tarball listing or not it seems like it might not have in the past. I have to check this and make any fixes. It turns out that maybe this issue of whether to have the quiet option or not is a good thing.

But this will happen some hours from now probably ... must go afk a while now!

xexyl:
I have looked at your comments again and I think that it's good, what I have, except that I have to update the man page. It will require updating the guidelines but I'll probably not do that until after this is merged.

I have found another problem that I think might be strictly to do with the test suite (as in `make test`) where messages are not shown in the right order which I guess I should report as a bug but that can happen later too, after I do the commit that makes the fixes.

I think you'll appreciate the new messages too.

xexyl:
> As much as we enjoy puns, gibberish, humour, anecdotal stories, etc. in these discussions, sometimes they get in the way of `The Signal`. The `mkiocccentry` tool suite should first and foremost be about the signal and less noise.
> 

Now that I'm more awake and I saw this again the words `The Signal` reminds me of the [Great Signal](https://www.glyphweb.com/arda/g/greatsignal.html). No .. I can't help but list another reference to that book! It's all the more something because I feel like a lot of people do not comprehend what is going on then. It is somewhat unclear in part so that might be why but ... the Great Signal was important in the [War of the Ring](https://www.glyphweb.com/arda/w/warofthering.html).

Of course my comments I hope should allay your concerns. It was requested that the mkiocccentry tool use the `-e` option but not all the messages will be seen and those that are shown should not be in the way of packaging a submission for the contest. 

A commit will be made soon and I hope it resolves the requests here.

xexyl:
Okay Landon .. does https://github.com/ioccc-src/mkiocccentry/pull/902/commits/b37e9a1cbca865f6ef0aad1335e0781e7434ca5f take care of the problem? I think it should and it offers a bit more fun (thanks for the suggestion!). The man page does not give examples of these fun things though. Should they? I don't know but I was trying to get this done.

Once this is merged I can update the guidelines. If you still have a request or requests I will take care of those first.

I will be going afk soon for a little bit at least. I hope to hear back from you in the other repo and then I can move forward there too. But what else I get done today I really do not know. It kind of depends on hearing back from you. I have one thing to try, maybe, that might make it not necessary to hear back from you over there and it might be I can just worry about that thing later (as in go back to it) but that will be worked out later on.

lcn2:
> As for the `quiet` and showing checking sanity checks etc.: that was because the other tools have that very option as well.
> 
> 
> 
> I can indeed change it to your `-e` idea: I like that. But there is something you're forgetting too.
> 
> 
> 
> The tool has to print something at least part of the time: `mkiocccentry(1)` **USES** `txzchk(1)` to print out the contents of the tarball. Without `-q` it is shown and with it it is not. So it's not necessarily good to get rid of `-q` outright.
> 
> 
> 
> Still some of it could be changed to be shown with `-e`. What I'll do when more awake is change all that can be changed to use that option and the rest not.
> 
> 
> 
> As for the version update the change is that it no longer shows warnings if quiet.
> 
> 
> 
> 

You raise some very interesting points that we seem to have forgotten with respect to how `mkiocccentry(1)` **USES** `txzchk(1)`.  Thanks for the reminder!

Perhaps in addition to some "human entertainment command line option" we need to add a "for use by mkiocccetry" command line for the purpose of "print(in) out the contents of the tarball".  Maybe that flag is `-q` and doesn't need a new option?  Maybe a new command line option is needed so that we don't have to explain why `-q` isn't quiet.

Thinking out load here: an UPDATE to the comment may be in order.  Stay tuned.

lcn2:
> From the above paragraph: it sounds like you think it should always print warnings / issues with the tarball. But that then means that if there are problems one cannot simply check the exit code. A recent issue you had me make it so that it works this way even. So this seems to be a change in what you wanted or needed.
> 
> 

We are not sure if "it should always print warnings / issues with the tarball": perhaps not now that you mention this issue.  What do you think?

lcn2:
> As for an option name suggested by @lcn2 for extra output, eg. `-e` I would argue `-vv` (very verbose) (pseudo) diagnostic might better be suited.  I agree with @lcn2 about how the tool should be silent on success and useful unburden diagnostic output otherwise.  Verbosity levels might be a better way.  Sane and boring being the default, -vv, -vvv getting more bizarre.  However I would suggest reserving -v for sane detailed diagnostic messages, before having fun.  `iocccsize` supports a `-v level` (instead of multiple -vvv's) mainly for my own debugging.  However I could in future make, eg. -v 99, inject odd ball comments.

The mkiocccentry tool set (as well as bin tools in the [temp-test-ioccc repo](https://github.com/ioccc-src/temp-test-ioccc) all use the "-v level" model, sorry (tm Canada 🇨🇦).

lcn2:
> As much as we enjoy puns, gibberish, humour, anecdotal stories, etc. in these discussions, sometimes they get in the way of `The Signal`.  The `mkiocccentry` tool suite should first and foremost be about the signal and less noise.
> 
> 
> 
> I don't know about yourself, but I recall many a time when submitting entries how stressful the process could actually be, from getting an accurate size count to actually pressing "send" (and hope the mail didn't bounce).  Adding superfluous output clouds the signal at a critical point.  
> 
> 
> 
> `gcc` error and warning messages have become multiline essays on what is wrong with some code; there used to be a time you could count errors and warnings by single lines.  And `gcc` is trying to be helpful!  Imagine if `gcc` interjected random jokes and jibes between the error and warning messages how hard it would be to actually fix something.
> 
> 
> 
> Another point to make about superfluous messages is you have to maintain the code going forward.

We believe that @SirWumpus has raised some very important points in the quote we have added above.  The "should first and foremost be about the signal and less noise" is well taken.

Adding a `-e` (or whatever the "entertainment command line option" becomes) to `txzchk` would mean that such messages would **NOT** appear by default.  Someone would have to want to see them by adding the "entertainment command line option" to their command line.

As far as what `mkiocccentry` prints, he raises a fair point.   Perhaps the `mkiocccentry` also needs a `-e` (or whatever the "entertainment command line option" becomes) too?  Running `mkiocccentry` by default would not mention feathers.

BTW: The reference to "tarball" and feathers, while fun, might not be well understood by all international users, which is another reason why we might not want to print them by default.

So perhaps by default, a `mkiocccentry` when it passes to `txzchk` the "the `txzchk` tool is being run by `mkioccentry` command line option), a message of the form:

```
No warnings or errors detected in ...
```

Would be best by default?

lcn2:
> Hmmm .. a downside of having to use `-v` to show tarball listing is that it means that other lines will also be shown when running `mkiocccentry`. For instance, debug output.
> 
> 
> 
> That would be a good reason to keep `-q` in. The other issue is resolved by adding a `-v 1` to the invocation of `txzchk`.

This is where a "this tool be being run on behalf of mkiocccenrry" command line option might help.

xexyl:
> > As much as we enjoy puns, gibberish, humour, anecdotal stories, etc. in these discussions, sometimes they get in the way of `The Signal`.  The `mkiocccentry` tool suite should first and foremost be about the signal and less noise.
> > I don't know about yourself, but I recall many a time when submitting entries how stressful the process could actually be, from getting an accurate size count to actually pressing "send" (and hope the mail didn't bounce).  Adding superfluous output clouds the signal at a critical point.
> > `gcc` error and warning messages have become multiline essays on what is wrong with some code; there used to be a time you could count errors and warnings by single lines.  And `gcc` is trying to be helpful!  Imagine if `gcc` interjected random jokes and jibes between the error and warning messages how hard it would be to actually fix something.
> > Another point to make about superfluous messages is you have to maintain the code going forward.
> 
> We believe that @SirWumpus has raised some very important points in the quote we have added above. The "should first and foremost be about the signal and less noise" is well taken.
> 
> Adding a `-e` (or whatever the "entertainment command line option" becomes) to `txzchk` would mean that such messages would **NOT** appear by default. Someone would have to want to see them by adding the "entertainment command line option" to their command line.
> 
> As far as what `mkiocccentry` prints, he raises a fair point. Perhaps the `mkiocccentry` also needs a `-e` (or whatever the "entertainment command line option" becomes) too? Running `mkiocccentry` by default would not mention feathers.
> 
> BTW: The reference to "tarball" and feathers, while fun, might not be well understood by all international users, which is another reason why we might not want to print them by default.
> 
> So perhaps by default, a `mkiocccentry` when it passes to `txzchk` the "the `txzchk` tool is being run by `mkioccentry` command line option), a message of the form:
> 
> ```
> No warnings or errors detected in ...
> ```
> 
> Would be best by default?

I actually thought of that. For now I'll do a commit that disables the `-e` in mkiocccentry. Then later I can look at the option to add it back. That way this can be merged and complete.

xexyl:
Doing make prep. Should be fine. If so I'll do a commit and then that should be it. Maybe the `mkiocccentry` should NOT have it either. Perhaps it should be only fun for those who wish to invoke the command directly? I can see that appeal too.

xexyl:
> > Hmmm .. a downside of having to use `-v` to show tarball listing is that it means that other lines will also be shown when running `mkiocccentry`. For instance, debug output.
> > That would be a good reason to keep `-q` in. The other issue is resolved by adding a `-v 1` to the invocation of `txzchk`.
> 
> This is where a "this tool be being run on behalf of mkiocccenrry" command line option might help.
> 
> ## UPDATE 0a
> > Now another important point is .. should the other tools have `-q` removed too? This is why I added the option to `txzchk` in the first place, to match the other tools and the `dbg` repo/library.
> > Please advise and I'll be happy to remove it - if not.
> 
> We suggest adding:
> 
> * an entrainment command line option to `txzchk`
> * an entrainment command line option to `mkiocccentry` that added on to `txzchk`
> * a command line option to `txzchk` that indicates it is being called by `mkiocccentry` and to output accordingly
> 
> At this point the `-q` might not be needed and might only cause confusion about its purpose.
> 
> Comments welcome.

Actually `-q` was removed this morning. I just removed `-e` from `mkiocccentry(1)`. If `make prep` works fine (if not it'll only be need of fixing error files) then I'll do a commit. We can worry about `mkiocccentry` another time. I want this issue resolved.

xexyl:
All good. Will make a commit now.

xexyl:
There. Done. We can worry about the other tool having a like option another time. I am trying to get some documentation fixes (in guidelines) done and then leave which is why I want to not worry about doing the mkiocccentry part now (plus I'm not sure if it should have it or not .. I do like the idea but since it popped into my head that it might not I'll let the idea sit a bit).

Anyway -q no longer exists, warnings are shown if there are any (i.e. invalid tarball) like you wanted (or that's how I understood it .. please advise if not the case).

lcn2 (approved):
Thank you @xexyl and thanks for your patience as we reviewed and improved the PR.

xexyl:
> Thank you @xexyl and thanks for your patience as we reviewed and improved the PR.

No problem at all.

xexyl:
> Thank you @xexyl and thanks for your patience as we reviewed and improved the PR.

As far as the way mkiocccentry uses it I solved it with -v 1. 

It can certainly be changed back if you want but this probably works fine?

As far as an entertainment option I can do that. Do you have a suggested letter if `e` isn't available?

I am unfortunately unable to give your comments a proper look but I will do so tomorrow. 

This was just me passing through.

lcn2:
> It can certainly be changed back if you want but this probably works fine?

No need, as works as it is now.

xexyl:
> > It can certainly be changed back if you want but this probably works fine?
> 
> 
> 
> No need, as works as it is now.

So the question is should the mkiocccentry code have an option to enable the entertainment option?

Have to leave again. Back tomorrow though I have an appointment tomorrow. Doctor.

lcn2:
> > > It can certainly be changed back if you want but this probably works fine?
> > 
> > 
> > No need, as works as it is now.
> 
> So the question is should the mkiocccentry code have an option to enable the entertainment option?
> 
> Have to leave again. Back tomorrow though I have an appointment tomorrow. Doctor. Particularly ENT. Which I will bring up tomorrow probably.

It would be easier to test without, and more fun with for those that need entertaining.  Perhaps this not a priority?

xexyl:
But another thing is ... you said maybe the warnings should not be shown all the time, or so I think you said that. What do you wish? Now any warnings will be shown but is that how it should be? Maybe a verbosity level or some option has to be used and it would just exit 1 if there are any issues (rather than show warnings and exit 1)?

xexyl:
> > > > > It can certainly be changed back if you want but this probably works fine?
> > > > 
> > > > 
> > > > No need, as works as it is now.
> > > 
> > > 
> > > So the question is should the mkiocccentry code have an option to enable the entertainment option?
> > > Have to leave again. Back tomorrow though I have an appointment tomorrow. Doctor. Particularly ENT. Which I will bring up tomorrow probably.
> > 
> > 
> > It would be easier to test without, and more fun with for those that need entertaining. Perhaps this not a priority?
> 
> Maybe not .. but it could be added too as it would not take much time. I wonder if the letter is available for the tool ...
> 
> It is! I'll do a pull request for it. Seems like a good idea.

Doing a make prep but I have to go afk a bit now .. back later.

xexyl:
> > From the above paragraph: it sounds like you think it should always print warnings / issues with the tarball. But that then means that if there are problems one cannot simply check the exit code. A recent issue you had me make it so that it works this way even. So this seems to be a change in what you wanted or needed.
> 
> We are not sure if "it should always print warnings / issues with the tarball": perhaps not now that you mention this issue. What do you think?

I don't know. But if it did not print warnings all the time (which is how I had it) I'd have to add back the `-q` option. Which might not be a bad thing. It would be better I think than forcing one to use `-v 1`. 

Actually now i think on it: I think it's probably good to always show it and this is why: it's safer for users. They might not think to check the exit code. Okay if they use mkiocccentry as they're supposed to the problem would not occur but still .... this way they will always get warnings if there is some problem.

Thoughts ?

xexyl:
Commit https://github.com/ioccc-src/mkiocccentry/pull/908/commits/534c9bc7912c2b0a275b9d0166983b584f1fe4ca takes care of the `-e` in mkiocccentry. I think you'll be pleased with it.

I might have another minor thing to do but not sure .. the question is should the guidelines be updated for `-e` in `mkiocccentry`? Currently it only says that if one checks the source or man page they might find something fun in `txzchk`. I think that might be a nice way to go about it as it becomes more of a surprise (though the changelog is a hint that it might be something to try) ... the test script, to simplify and speed things up, always runs it. The fact we are the ones who use it most do appreciate it (and it shows other output - much more output) that people would also not understand that it hardly matters to have some fun (and it's a good thing for people who do understand it). Besides the guidelines do refer to 'feathers stuck in the tarball' as being issues. :-)

lcn2:
> Actually now i think on it: I think it's probably good to always show it and this is why: it's safer for users. They might not think to check the exit code. Okay if they use mkiocccentry as they're supposed to the problem would not occur but still .... this way they will always get warnings if there is some problem.
> 
> Thoughts ?

Showing warnings to users is a good idea, yes.

As far as `mkiocccentry` goes, it could show the same warnings to users and bail of `txzchk` exits non-zero as well.

PR #907, opened 2024-07-12
Fix (hopefully) workflow of make release

xexyl:
The problem, it would appear, is that in trying to be done for the day, I forgot to commit the Makefile, which would mean the rule could not be run since it was not there.


xexyl:
Well that might explain a lot ... will be going afk shortly and not sure when I'll be back but I do hope to do more today, most likely at the other repo.

lcn2 (approved):
Thanks for the `git add Makefile` @xexyl 😁

PR #910, opened 2024-07-13
Fix git clone picky

xexyl:
The repo it was changed to was actually a fork of my repo so it seems better to have it as my repo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
     - name: install checknr
       run: cd checknr && sudo make install
     - name: clone picky
-      run: git clone https://github.com/lcn2/picky
+      run: git clone https://github.com/xexyl/picky
     - name: install picky
       run: cd picky && sudo make install
     - name: make


xexyl:
Correct me if I'm wrong. I can change it back but it seems this is better.

lcn2 (commented):
That wasn't an error: we had been using our modified version of the code for some time now. We noticed the odd install mechanism and realized that our version of the code wasn't compatible (although after the pull request one could was it was). But the the same reason that we pulled in the `iocccentry` code, we did so with picky. 

BTW: We had been working with the original author before their code disappeared (see the man page): so having it around was nothing new.

xexyl:
> That wasn't an error: we had been using our modified version of the code for some time now. We noticed the odd install mechanism and realized that our version of the code wasn't compatible (although after the pull request one could was it was). But the the same reason that we pulled in the `iocccsize(1)` code, we did so with picky.

It does seem a bit odd to use a fork of a copy but as you will. That does necessitate some changes here I think as well as on the mkiocccentry end but I'll get to those later.

What install mechanism though? Did I not use `install`? I thought I did ... though maybe not.


> 
> BTW: We had been working with the original author before their code disappeared (see the man page): so having it around was nothing new.

I knew that and it's something that crossed my mind with the pull request too.

> 
> ## UPDATE 0
> In other words, we don't always need the most current code: we need a set of consistent code that has been tested and known to work together.
> 

Fair enough.

> We don't want, for example, the latest idea someone might have regarding the `iocccsize(1)` tool.
> 

Out of curiosity what do you mean by 'latest idea someone might have regarding ...' ?

> Once your _jparse repo_ running, this repo will not directly reference it. There will be that internal `jparse.clone` tree was the ability to sync with it (via those name diff and make clone rules), however what IOCCC contestants will use is a version of the copy: not the repo itself.
> 
O course .. kind of like with the dynamic array API and the debug API.

> Once the **Great Fork Merge** happens and you continue to develop your code (say by picking up our new `jval` tool, or your new `jprint` tool), the IOCCC won't instantly update. Instead integration of an updated jparse repo, say in regards to those "XXX" spots for bin tools, would only be considered after IOCCC28.
> 

I would expect that too. I would want that also.


> Why? Because it isn't needed right now. Unless there we found some "super code bug" with `jparse` (and related tool friends) we wouldn't change the code during a contest.
> 

That seems unlikely even but I would agree with the idea too.

> So the model for the mkiocccentry repo is to provide a single stop for picking up what someone needs submit to the IOCCC. It is also to provide a well tested set of code that is known to work together. And while a bug may be discovered in the future among that set of code, the behavior will be at least consistent.

I agree with this of course.

> 
> BTW: The reason why we put all those version strings into the JSON files, and why we have to do those annoying test updates every time a version changed, is because when we look back at a given JSON file for some reason we can know the code base context it was formed under.

Of course - and other reasons besides I am sure.

> 
> Say we discover some obscure bug that caused `fnamchk(1)` to approve a filename that was otherwise bogus. Knowing the `fnamchk(1)` version that was used would allow us to be "more forgiving" to a submission that used a bogus filename because we would see it used the buggy code base.

That seems unlikely too but of course it could happen.

But with these things in mind there are two other things:

0. You have not finished the job :-) there are references to the old repo. I was going to have you do it as I want to get the FAQ done and then maybe do other things today but since it's a quick fix with sgit I'll do it in a moment.
1. Maybe you should delete your fork and recreate the repo and then I can clone that instead. Of course I don't mind if you keep it the way it is as I can always merge (like I did yesterday) but it might be something you want to consider ... if you do please let me know.

See commit https://github.com/ioccc-src/mkiocccentry/pull/911/commits/23637f89d868026552fe4a5c368d43c5cbbbc7fa.

PR #912, opened 2024-07-14
Actually add picky link fix to Makefiles

xexyl:
Not sure how this did not get through as I could have sworn I did 'git commit -va' but apparently not.


lcn2 (approved):
Good catch @xexyl thanks 🙏

PR #914, opened 2024-07-14
Fix a few error messages in jparse/json_sem.c

xexyl:
In some of the functions sem_member_value_*() (those that validate JSON parse node types) the error message (if the type was wrong) reported incorrectly the type expected and in particular all of them suggested the type should be JTYPE_BOOL but none of the functions were for booleans (though all were for numbers I think).


lcn2 (approved):
Interesting bug and nice 👍 fix @xexyl

PR #923, opened 2024-08-02
Add JSON reference files for chkentry

xexyl:
This commit is to begin the process of updating chkentry(1) to validate .entry.json and author_handle.json files. The reference files use Landon as the winner/author as not only do I not want to be conceited :-) but he's a judge and it works well as a reference. A few notes of interest here.

The author_handle.reference.json file has the current email address of the Judges (which changes on a yearly basis). It was not clear if I should have null instead but this can be decided later.

Where needed the entry_ids and author_handles etc. match in the two files.

More can be done with this as more is discussed. The other data files that will be needed for this hacking of chkentry can be added later as more of the issue is discussed but this is a nice starting point.


xexyl:
When one of us opens the issue we can then look at these files and also discuss the other ones (like the patch files) and decide how to proceed. But I wanted to get these in now.

Afk a while .. when back have other things to do here too and then I think I'll either open the issue or else try working on the next year in the other repo. I have a feeling it's going to be a slow day though. Anyway we'll see. Back in a while!

lcn2 (changes requested):
Fun idea, however we request that instead of using what looks like a plausible email address `q.2024@ioccc.org`, we request modify this pull request to use something that isn't a valid email address. 

Here are a few ideas for what to change all references of `q.2024@ioccc.org` into an invalid address such as:

- lcn2@example.org
- lcn2@ioccc.example.org
- q.2024@example.org
- q.2024@ioccc.example.org

xexyl:
> 👍

I also, as you will have seen, added structs for them - or starting points - and starting points for the code. I won't be doing anything more here today - certainly not this repo if at all - but I will most likely look at 2001 tomorrow and I will aim to finish it though as you know for another week to week and a half I'll be slower.

PR #935, opened 2024-08-23
Minor fixes in Makefile.example

xexyl:
Typo fix and change 'entry' to 'submission' (in two places).

### DIFF
--- a/Makefile.example
+++ b/Makefile.example
@@ -1,12 +1,12 @@
 #!/usr/bin/env make
 #
 # XXX - Please remove these 3 XXX-ed comment lines - XXX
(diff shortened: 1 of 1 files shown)


lcn2 (approved):
Oopsy 😬

xexyl:
> Oopsy 😬
> 

Oopsy 🌼(daisy)?

> Thanks for the catch, @xexyl

You're welcome. Personally, though, in this case I'd rather **pick** flowers, not catch them: because if I have to catch flowers it means they're running away from me which means either I'm in a world that has gone mad or I've lost my bloody mind.

PR #939, opened 2024-08-24
Update .gitignore and test GitHub PR issue

xexyl:
There is a problem with GitHub where if one pull request is closed then
accepting ones that come later will merge the closed pull requests. It
is my belief that this only happens if the maintainer closes it; if the
submitter closes it it does not seem to be a problem.

This pull request is the same as a pull request I just closed except the
line order has been swapped. This is not a perfect test and it entirely
depends on the order GitHub does merging (one might think that it would
be the order merged but perhaps it's not that - we do not know as it
should not be happening in the first place).

The diff of THIS commit is:

    -.*.sw[a-zA-Z0-9]
     .sw[a-zA-Z0-9]
    +.*.sw[a-zA-Z0-9]

which means that if all is okay (after this is merged), the glob:

    .sw[a-zA-Z0-9]

should come before the glob:

    .*.sw[a-zA-Z0-9]

In other words the .gitignore file should have the lines in this order:

    .sw[a-zA-Z0-9]
    .*.sw[a-zA-Z0-9]

and if I described it wrong above it is because it is an off day. I
believe, however, that the order will be as described, whereas it MIGHT
have been different if the other pull request had been closed by Landon
instead of me.

If it is not clear, the glob that was added is:

    .sw[a-zA-Z0-9]

These globs might (unsure) be better to have a '*' at the end but this
can be done later (it is extremely unlikely that anyone would have that
many vi(m) swap files and I have never bothered to check the vim - let
alone even the vi - code to determine its - or their - algorithm in
generating swap file names).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
 *.[ao]
 *.dSYM/
 *~
-.*.sw[a-zA-Z0-9]
 .sw[a-zA-Z0-9]
+.*.sw[a-zA-Z0-9]
 .DS_Store
 ._.DS_Store
 /.build.log.*
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.dSYM/
 *~
 .*.sw[a-zA-Z0-9]
+.sw[a-zA-Z0-9]
 .DS_Store
 ._.DS_Store
 /.build.log.*


xexyl:
Here's a very interesting thing that I noticed. If I have NO pull requests open the log and summary of the commit is automatically filled in here. If however there is already a pull request it is not: it's just the summary and incorrect summary (based on the branch name).

In those cases I copy paste the log (including summary) prior to hitting `:q` in vim in the commit.

But this time, despite closing the previous pull request (thus making it where there is no open pull request), it did not populate the form.

And this is not because I did not delete the branch on GitHub, or seemingly this is not it, because I have many branches as GitHub (quite annoyingly) does not have a way to delete all but the master branch (which with the amount of pull requests I have made is a huge burden .. have to click on each one and I don't always get a chance to open the pull requests after they are merged or closed in order to delete the branch).

lcn2:
We believe that and you opened up an 2nd unrelated PR request to make some edit of some other file, we would not be able to independently complete (close/merge or close/reject) the 2 open PRs independently: such as the closing of the later open PR would also close the former open PR.

lcn2:
Two independent (in terms of files changed) consecutive PRs force a chaining order.  This may be a GH mis-feature.

We wonder what will happen after the **Great Fork Merge**.  Say someone wants to edit a file in a PR.  Assume that their edit is bogus and should be rejected.  Say someone issues a to edit a file in a subsequent PR.  Assume that this 2nd PR is good and should be accepted.  We believe we will have to reject the former PR and accept the later PR in that order.  But will accepting the 2nd PR force the 1st PR to be accepted?  Perhaps yes IF the same person issued both PRs?

What may be an issue here is the commit from the master branch from which each PR branches.   Two independent PRs from two different people, each starting with same commit number might produce independent PRs.    Two independent PRs from the same person, because they are consecutive, might happen where the 1st PR started is based on the commit from the master branch, and then the 2nd PR starts from the commit of the 1st PR (not the commit number from the master branch).

xexyl:
It was an experiment to see what happens when I close it.

PR #948, opened 2024-08-28
Fix hostchk.sh

xexyl:
With the recent updates to the inclusion of the different repos and the MKIOCCCENTRY_SRC a problem in linux for hostchk.sh had to be resolved. This might also be a problem in macOS but since the libraries were already installed this is not detected (it seems that in linux if the other libraries were installed this also was not a problem).

There still exists an issue with building, however, as the make dependencies have system paths (cc -MM will solve this so it might be the way independ works but this is not clear to me at this time). Nonetheless the hostchk.sh should at least now work.

New version of hostchk.sh is 1.0.2 2024-08-28.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 1.5.3 2024-08-28
+
+Fix `hostchk.sh` that was broken with the recent change of `MKIOCCCENTRY_SRC`.
+New version for hostchk.sh: `"1.0.2 2024-08-28"`.
+
+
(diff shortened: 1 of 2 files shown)


xexyl:
As noted build will still fail but I believe this to be a problem with the make dependencies. Maybe that tool you wrote can skip over system paths like `cc -MM` does? Or perhaps it's because of how `#ifdef..#endif` works?

lcn2:
We had to resolve the conflict and edit the grep regex by hand .. hope this works.

lcn2:
Stay tuned for a fix to Makefiles for dbg, dyn_array, and perhaps jparse too.

PR #950, opened 2024-08-29
Rename library filenames for external repos

xexyl:
Rename dbg.a, dyn_array.a and jparse.a to libdbg.a, libdyn_array.a and libjparse.a, respectively. These respective repos have had the respective change done as well (well the jparse one has not been committed yet as it needs the other libraries in order to work).

This is important for the jparse repo (not the jparse/ subdirectory here though this has also been updated) to be usable outside this repo without having to have ../dbg and ../dyn_array directories containing the correct respective libraries. The jparse repo
(https://github.com/xexyl/jparse) still has to be updated (and there might be some additional changes here as well) but that cannot happen until the dbg and dyn_array repos have also had this done (or rather merged).

The Makefile does NOT use -ldbg, -ldyn_array or -ljparse (yet?). Obviously if it does it'll use the -I option.

Updated MKIOCCCENTRY_REPO_VERSION to match this release version (this means it jumped from 1.5.2 to 1.5.4 as it was not updated when 1.5.2 2024-08-28 was updated to 1.5.3 2024-08-29).

Ran make prep to test these changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,18 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.5.4 2024-08-29
+
+Rename libraries for easier linking in other applications. In particular the
+[jparse repo](https://github.com/xexyl/jparse/) (which currently requires some
+changes in order to work on its own but this release is part of that) needs the
+`dbg` and `dyn_array` libraries but in order to link these in properly one must
+name them `libdbg.a` and `libdyn_array.a` (i.e. add the `lib`) so that one can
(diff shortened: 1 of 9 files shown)


lcn2 (approved):
This PR looks like a good fix.  We will see how it plays out on other distributions .. nevertheless we think you are in the correct track.  Thank you very much @xexyl ‼️

lcn2:
We tested the mkiocccentry as modified up to this PR and `make release` works!  Well done @xexyl.  👏

We will later rename `MKIOCCCENTRY_SRC` to `LOCAL_DIR` where needed.  That is not critical but eating dinner so, so BCNU later.

PR #953, opened 2024-09-01
Improve Makefiles (install, uninstall rules)

xexyl:
Add to Makefiles the 'PREFIX' variable to allow for installing to a different location than the default /usr/local.

Added some uninstall rules to the Makefile in a number of subdirectories. The external repos that are synced to this repo have not had any direct changes as those repos first need those rules (which will be done soon). This is also why the top level Makefile does not run 'make uninstall' in those subdirectories. Once these are all done the top level Makefile can have a simpler uninstall rule.

Along the lines of silencing (non-verbose) install in the install rule one may now do the same with the uninstall (instead of 'INSTALL_V' it is 'RM_V'). This is only done to be congruent with the install rule.

Updated the MKIOCCCENTRY_REPO_VERSION to "1.5.7 2024-09-01".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,23 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.5.8 2024-09-01
+
+Add to Makefiles the `PREFIX` variable to allow for installing to a different
+location than the default `/usr/local`.
+
+Added some uninstall rules to the Makefile in a number of subdirectories. The
+external repos that are synced to this repo have not had any direct changes as
+those repos first need those rules (which will be done soon). This is also why
+the top level Makefile does not run `make uninstall` in those subdirectories.
+Once these are all done the top level Makefile can have a simpler uninstall
+rule.
+
+Along the lines of silencing (non-verbose) `install` in the `install` rule one
+may now do the same with the uninstall (instead of `INSTALL_V` it is `RM_V`).
+This is only done to be congruent with the `install` rule.
+
+Updated the `MKIOCCCENTRY_REPO_VERSION` to `"1.5.7 2024-09-01"`.
+
 ## Release 1.5.7 2024-08-31
 
 Synced `jparse` subdirectory from the [jparse
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,6 +1,6 @@
 # Frequently Asked Questions about the `mkiocccentry` repo
 
-This is FAQ version **28.0.0 2024-07-31**.
+This is FAQ version **28.0.1 2024-09-01**.
 
 <div id="toc"></div>
 
@@ -10,7 +10,7 @@ This is FAQ version **28.0.0 2024-07-31**.
 
 1. [How do I compile the mkiocccentry tools?](#compiling)
 
-2. [Do I need to install this code to use it?](#install)
+2. [How may I install the tools if I wish to and is it required?](#install)
 
 3. [What can I do if my system's tar(1) does not support the correct options?](#tar)
 
@@ -23,6 +23,8 @@ This is FAQ version **28.0.0 2024-07-31**.
 7. [Why do these tools sometimes use the incorrect IOCCC terms?](#terms)
 
 8. [How do I participate in the IOCCC?](#ioccc)
+
+9. [How do I uninstall the toolkit?](#uninstalling)
 
 
 <div id="download"></div>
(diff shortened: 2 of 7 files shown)


xexyl:
Now for the dbg and dyn_array repos. I do have to hurry t hough as I have other things I need to do that I was not able to finish yesterday. Once these are merged we can sync the subdirectories from the repos.

I observe again that the check failed here due to the library not being found due to no `-L` options. As that seems to be something you want to take care of I won't deal with it.

lcn2:
Did you notice that the checks failed, @xexyl?

lcn2 (approved):
This looks good at first glance, @xexyl

lcn2 (changes requested):
Please attempt to address the check failures, @xexyl, before we consider this PR.

We will hold off on this PR (and this future PRs will be delayed), until this matter is addressed.    We also recommend addressing the suggestions we made in the related [dbg PR](https://github.com/lcn2/dbg/pull/22#pullrequestreview-2274496211) and the related [dyn_array PR](https://github.com/lcn2/dyn_array/pull/6#pullrequestreview-2274496137) as well: this way we can process all 3 at the same time.

xexyl:
Yes I am afraid it's needed, that option. I will play with this tomorrow morning. 

Thanks.

lcn2:
> Yes I am afraid it's needed, that option. I will play with this tomorrow morning.
> 
> Thanks.

Thank you for looking into this, @xexyl 

What we ask is that whatever this mkiocccentry repo needs for using the local libraries `dbg/libdbg.a` and `dyn_array/libdyn_array.a` **NOT** impact the case where someone has installed the [dbg repo](https://github.com/lcn2/dbg) and/or [dyn_array repo](https://github.com/lcn2/dyn_array).  Those 2 repos should continue to install, under the control of `PREFIX` (that defaults to `/usr/local/include`) such that existing tools (including those well outside the scope of the IOCCC and jparse) will not see any changes.

xexyl:
I understand that. I will look at this soon. Hopefully it will be easy to sort out.

xexyl:
Hmm .... the `LD_SPECIAL` idea does not work possibly due to spaces but not sure of that yet.

xexyl:
In fact .... `C_SPECIAL` does not work if it requires more than one flag. So this might not be a good idea .. not sure. Maybe have to explicitly pass in `-L` to each line? Not sure.

xexyl:
BTW if setting `LDFLAGS` to the right `-L` options:

```
clang -std=gnu17 -O3 -g3 -pedantic -Wall -Wextra -Wno-char-subscripts -DINTERNAL_INCLUDE -L../dyn_array -L../dbg mkiocccentry.c -c
clang: warning: argument unused during compilation: '-L../dyn_array' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L../dbg' [-Wunused-command-line-argument]
```

so that isn't working either. And in the case of `${MAKE} -C` it's not going to work either I think. if the Makefiles themselves have in the `${CC}` line it would work presumably but other than that ...

xexyl:
Now I can do the make depend change and you can try and resolve this afterwards - might be a good idea to move things along - but I leave that up to you.

lcn2:
> This is what happens:
> 
> 
> 
> ```
> 
> make: invalid option -- .
> 
> make: invalid option -- .
> 
> make: invalid option -- /
> 
> make: invalid option -- g
> 
> 
> 
> 
> 
> ```
> 
> 
> 
> That's through passing the `LD_SPECIAL` to the make rule. If `C_SPECIAL` has more than one arg similar. Let's say `-Dfoo`:
> 
> 
> 
> ```
> 
> make: invalid option -- D
> 
> 
> 
> ```
> 
> 
> 
> Same problem with quoting it. So this is a bad idea as it stands. Which means these fixes cannot be applied. I'll remove `LD_SPECIAL` from jparse since it seems to not be a good idea.

One has to quote stuff that has a space, and when dealing with a make invocation put the make variable assignment at the end. 

We will process these PRs as is and then we will fix, if and as needed afterwards.

xexyl:
> > This is what happens:
> > ```
> > 
> > make: invalid option -- .
> > 
> > make: invalid option -- .
> > 
> > make: invalid option -- /
> > 
> > make: invalid option -- g
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That's through passing the `LD_SPECIAL` to the make rule. If `C_SPECIAL` has more than one arg similar. Let's say `-Dfoo`:
> > ```
> > 
> > make: invalid option -- D
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Same problem with quoting it. So this is a bad idea as it stands. Which means these fixes cannot be applied. I'll remove `LD_SPECIAL` from jparse since it seems to not be a good idea.
> 
> One has to quote stuff that has a space, and when dealing with a make invocation put the make variable assignment at the end.
> 
> We will process these PRs as is and then we will fix, if and as needed afterwards.

I was pretty sure I did that but maybe not .... thanks for taking care of it.

PR #954, opened 2024-09-04
Add to -V option json parser version

xexyl:
As the json parser is now in its own repo
(https://github.com/xexyl/jparse) it seems like a good idea for the tools here to tell us what version is linked in in the case that there is a divergence (which can happen though it is not necessarily likely).

At the same time, because the parser is now in its own repo, the version of both the tool jparse (which is not in -V as it's not needed) and the library have been updated.


lcn2 (approved):
Nice 👍 idea 💡 @xexyl

PR #958, opened 2024-09-05
Synced jparse subdirectory from the jparse repo

xexyl:
Synced jparse subdirectory from jparse repo
(https://github.com/xexyl/jparse/) with some useful updates and fixes.
See the git log of that repo for more detailed information. Changes that
these updates required here (and some that are in the jparse
subdirectory):

- The test_ioccc/ioccc_test.sh script no longer runs the jstr_test.sh,
jnum_chk tests and it does not run jparse_test.sh except
on the test_ioccc/test_JSON directory as the jparse/Makefile runs the
appropriate tests in test_jparse/ (in this repo under jparse/test_jparse) -
obviously the jparse repo knows nothing about test_ioccc/test_JSON so it has
to be done this way. It's likely that with the -Z topdir option (that was
recently added to jparse_test.sh) or some other workaround the script could
run from test_ioccc/ioccc_test.sh but it is redundant and not useful so it no
longer does.
- Due to jparse_test.sh path updates (one of the changes is to fix it to work
on its own but this required some changes in test error files) the error files
under jparse/test_jparse/test_JSON/bad_loc have been updated here (this is
another reason that we cannot as easily run jparse_test.sh from ioccc_test/
without the -Z topdir hack or some other workaround).

A useful update (besides the addition of the -Z topdir hack) to
jparse_test.sh that was synced here is the new -f option for the files that
hold JSON blobs, one per line: it inverts the check, saying that the JSON blobs
must be invalid. This required a new file here,
jparse/test_jparse/json_teststr_fail.txt. As jparse_test.sh always runs on
at least json_teststr.txt (if no files specified) it might be good to not have
the option and always run a fail test on the new file but this can be worried
about another time. As the jparse/test_jparse/Makefile runs it with this
option it doesn't matter much anyway.

The jparse/test_jparse/jparse_test.sh version is now 1.0.5 2024-09-04 (fixed
from the international way 1.0.5 04-09-2024 that was added by habit, to
match the format of the other versions); the old version was 1.0.3
2023-08-01.

make release should be fine now, after the updates to ioccc_test.sh. The
version of that script is now "1.0.2 2024-09-05".

Repo release version is now 1.5.10 2024-09-05 (a recent update did not update
the version string so it jumped more than one). Except for some last minute
fixes that might be required it might be the last release until after IOCCC28
(let's all hope it is!).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *~
 .DS_Store
+._.DS_Store
 /Makefile.local
 /NOTES/
 /answers.txt
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,51 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.5.10 2024-09-05
+
+Synced `jparse` subdirectory from the [jparse
+repo](https://github.com/xexyl/jparse/) with some useful updates and fixes. See
+the git log of that repo for more detailed information. Changes that these
+updates required here (and some that are in the jparse subdirectory):
+
+- The `test_ioccc/ioccc_test.sh` script no longer runs the `jstr_test.sh`,
+`jnum_chk` tests and it does not run `jparse_test.sh` except
+on the `test_ioccc/test_JSON` directory as the `jparse/Makefile` runs the
+appropriate tests in `test_jparse/` (in this repo under `jparse/test_jparse`) -
+obviously the jparse repo knows nothing about `test_ioccc/test_JSON` so it has
+to be done this way. It's likely that with the `-Z topdir` option (that was
+recently added to `jparse_test.sh`) or some other workaround the script could
+run from `test_ioccc/ioccc_test.sh` but it is redundant and not useful so it no
+longer does.
+- Due to `jparse_test.sh` path updates (one of the changes is to fix it to work
+on its own but this required some changes in test error files) the error files
+under `jparse/test_jparse/test_JSON/bad_loc` have been updated here (this is
+another reason that we cannot as easily run `jparse_test.sh` from `ioccc_test/`
+without the `-Z topdir` hack or some other workaround).
+
+A useful update (besides the addition of the `-Z topdir` hack) to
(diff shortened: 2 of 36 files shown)


xexyl:
Have to go .. back tomorrow .. if any errors happen here I'll address today hopefully but should be fine (local tests suggest it).

xexyl:
Will fix problems later ... sorry for delays!

xexyl:
Hmm ... wonder why:

```

These test(s) failed:

    jparse/test_jparse/jparse_test.sh for general.json non-zero exit code: 6

About to exit: 28
```

when I don't get that locally. But have to leave again .. will fix later today I hope.

xexyl:
This is so bizarre ... trying the same command at the same location and it works. Or trying it from the root directory. It works locally running `make test`.

I can't figure out what it is for this reason, obviously, and so I'm not sure how to fix it. I'm trying something and if it works locally I'll try committing it and seeing if it solves it. If not I'm going to have to think about it as I have other things i need to take care of.

xexyl:
Will see if this workaround works .... works locally also.

xexyl:
Ah. In that directory there are no bad_loc files. I wonder if that's why I had the option in the first place. Maybe if I add some files in that directory it'll work. I'll try that next. If it does then we can try rolling back the previous hack. Otherwise have to look at it further. Still strange it works locally but not here.

xexyl:
Ah! I think I know the problem. Since empty directories cannot be added to git it cannot find the bad_loc directory when `-d` is used here but can locally.

I also have to fix the exit codes (make them consistent after the other change). So the task is to add a file in bad_loc and then also fix the exit codes. That should hopefully solve the problem.

xexyl:
Great! Solved. Now for another minor fix.

lcn2 (approved):
Good stuff, @xexyl thank you!

xexyl:
> Good stuff, @xexyl thank you!

You're welcome. Hopefully today's commit is all that'll be needed here.

PR #961, opened 2024-09-07
Sync jparse/ from jparse repo (again)

xexyl:
It was decided that the jparse error reporting could be improved. This was done in the following ways:

- If verbosity level is > 0, then it will show the invalid token and hex value of that token (along with the line and column).
- If verbosity is not specified, then it will just show the syntax error (the bad token with the line and column) and then the warning that the JSON tree is NULL (just like if verbosity specified) and then the error message (from jparse(1) itself).

The error files in the bad_loc have been updated as now the error output has changed.

Updated jparse(1) version to 1.1.6 2024-09-07.

The man pages also now refer to the jparse repo, not the mkiocccentry repo.

### DIFF
--- a/jparse/.gitignore
+++ b/jparse/.gitignore
@@ -4,6 +4,7 @@
 .*.sw[a-zA-Z0-9]*
 .sw[a-zA-Z0-9]*
 .build.log.*
+build.log
 .DS_Store
 /bug-report*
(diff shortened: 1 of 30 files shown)


xexyl:
Will take care of timestamp issues in a bit ..

lcn2 (approved):

PR #962, opened 2024-09-07
Remove timestamps from Makefiles

xexyl:
Although it can be useful at times it can also be a bit much, especially
with 'make -C foo depend' and so on.

The CHANGES.md file was updated for the above but also for 
0192122984a42294154f1a2eacd6af0a17df428f as it was left out by accident.


lcn2 (approved):

PR #963, opened 2024-09-07
Fix clone rules to exclude .github subdirectory

xexyl:
Fix clone rules in Makefile to exclude the .github subdirectory of outside
repos.

Removed the .github subdirectory from the dbg/ and dyn_array/
subdirectories as they serve no purpose here. Unfortunately without 
thinking (prior to realising the repo included the subdirectory) I did 
'rm -r jparse/.github' so and then did a commit so it appears that 
jparse/.github might not have been deleted from the repo. If this is the
case it'll have to be removed by someone with a different .git status.

Updated .exclude to include (not that that makes sense :-) ) the glob
'._.DS_Store' which sometimes shows up in macOS (it is different from
'.DS_Store').


lcn2 (approved):

PR #966, opened 2024-09-08
Add enhancement issue template

xexyl:
There have been many issues opened that are enhancements rather than a bug report or something else so it seems fitting that there should be an enhancement issue template (that adds the enhancement tag). One might even say that I added this because I had an issue with the issues. :-)

This comes directly from the jparse repo
(https://github.com/xexyl/jparse/).


lcn2 (approved):
👍

PR #967, opened 2024-09-08
Sync dbg and dyn_array from repos

xexyl:
The Makefiles were out of date and now the legacy_clobber rule removes
the old library filenames.


lcn2 (approved):
Sync 👍

PR #968, opened 2024-09-08
Sync jparse/ from repo

xexyl:
Some fixes to the jparse Makefiles had to be made and these are now
synced to this repo. These fixes include:

- Fix make clobber to remove jparse_test.log and Makefile.orig.

- Fix make legacy_clobber to remove jparse.a.

- Fix ${RM} in Makefiles to use ${Q} variable (not in make depend as it is
used in an earlier command in the multiple line commands), in some cases changed
from the wrong variable, and ${RM_V} (where this was not done).

- Do not use -r in rm in Makefiles unless removing a directory, for safety.

The MKIOCCCENTRY_REPO_VERSION was updated to reflect the CHANGES.md.

### DIFF
--- a/.github/ISSUE_TEMPLATE/enhancement.yml
+++ b/.github/ISSUE_TEMPLATE/enhancement.yml
@@ -1,0 +1,47 @@
+name: Enhancement
+description: Suggest an enhancement
+title: "Enhancement: <enhancement title>"
+labels: ["enhancement"]
+body:
+- type: checkboxes
+  attributes:
+    label: Is there an existing issue for this?
+    description: Please search to see if an issue already exists for the feature you're suggesting.
+    options:
+    - label: I have searched for existing issues and did not find anything like this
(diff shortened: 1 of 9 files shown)


xexyl:
The `-v` thing should be done in the dbg and dyn_array repos too ... going to reply to comments now and then I hope to look at the `prep.sh` fixes required. But I have other things I have to do too and unfortunately I'm running out of time before I can finish those so I'm not sure what else I can get done here. Assuming there are no additional fixes needed (and maybe I missed some) then I can at least get to the script tomorrow if not today.

lcn2 (approved):
`sync(2)` 🤓 @xexyl

xexyl:
> `sync(2)` 🤓 @xexyl

I prefer `sink(1)`. What is this tool? Here's a man page I just wrote:

Screenshot 2024-09-08 at 12 06 24

PR #969, opened 2024-09-08
Fixes in Makefiles

xexyl:
The clone rules left Makefile.orig files that should be removed and they
now do (for dbg/, dyn_array/ and jparse/).

The ${RM} variable in the Makefiles that did not use ${Q} now do.

The ${RM} no longer uses -r if a directory is not being removed. This
was done in jparse/ at least and the local Makefiles but if it was done
in dbg and dyn_array repos it is not yet merged to be synced up (if it
is not committed it needs to be done).

The ${RM_V} variable is now empty by default except for dbg/ and
dyn_array as those have to be committed, merged and then synced again.
In jparse/ this was already done.


lcn2 (approved):
👍

PR #973, opened 2024-09-11
Sync jparse/ with jparse repo

xexyl:
This includes a number of fixes and improvements to various tools and the Makefiles (which will possibly be implemented here too). One change not mentioned in the changelog in the jparse repo was a legacy of jparse being at one time in this repo, where the tags related rules referred to ../dbg and (incorrectly) ../dyn_alloc (should have been ../dyn_array). These references have been removed outright.


lcn2 (approved):
👍

PR #974, opened 2024-09-12
Bug fixes and enhancements in jstr{de,en}code

xexyl:
Synced jparse/ from the jparse repo (https://github.com/xexyl/jparse/). This makes some important fixes to the tools jstrdecode(1) and jstrencode(1) where in jstrdecode(1) the -Q option did not work and for both the printing of everything should happen after everything is parsed (especially for the -Q option but not strictly for that reason). The pr_jparse_test tool had a bug fix as well where the -h option did not work.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Major changes to the IOCCC entry toolkit
 
-## Release 1.5.15 2024-09-12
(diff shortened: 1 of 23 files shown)


lcn2 (approved):
ok

PR #975, opened 2024-09-15
Add helper script soup/is_available.sh

xexyl:
This script takes a single arg and exits 0 if it can be found via 'type -P', otherwise 1. If -v 1 or greater it gives either the tool's path or says it cannot be found.

This script will be used in prep.sh for a rather important enhancement to help prevent superfluous bug reports.

PR #976, opened 2024-09-15
Fix make check_man, sync jparse

xexyl:
Fix check_man Makefile rule in all Makefiles. Previously it ignored the result
which meant that prep.sh never reported a problem in a problematic man page.

The dbg and dyn_array repos need to have this fix applied and I will do
this as well.

Synced jparse repo for recent changes, which includes the above fix, a 
man page fix and a new helper function besides.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.5.17 2024-09-15
+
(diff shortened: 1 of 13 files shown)


lcn2 (approved):
The limitations and issues with [GH-pullrequestreview-230561870](https://github.com/ioccc-src/mkiocccentry/pull/975#pullrequestreview-2305618703) not withstanding, we will approve this PR as is.

Issue #13, opened 2022-01-18
Use tar options to disable including of the username/group in the archive.


ilyakurdyukov:
Check the binary content of these archives: 
```
$ tar -cf test.tar README.md
$ hexdump -C test.tar

$ tar --numeric-owner --owner=0 --group=0 -cf test.tar README.md
$ hexdump -C test.tar

# aliases: --old-archive, --portability, --format=v7
$ tar --format=v7 -cf test.tar README.md 
$ hexdump -C test.tar
```

lcn2:
Sadly @ilyakurdyukov, the --owned and --group are not found on several distributions.  However --numeric-owner appears to be fairly common.  With just --numeric-owner by itself, the ASCII username and group no longer are stored.  And while we don't 'peak', knowing the UID and GID doesn't say anything useful without the authorship.

On the extraction side, we don't extract as root :-), and we can use --no-same-owner.

lcn2:
The use of --format=v7 appears to have a similar effect as just using -numeric-owner, @ilyakurdyukov so we wonder which is more common.  Any ideas?

lcn2:
We presume that --format=v7 is fairly common now. @ilyakurdyukov and use of it has several effects:

1. No symbolic ownership: the same effect as --numeric-owner
2. Maximum value of user or group ID is limited to 7777777 octal
3. No store special files: no block devices, no character devices, no fifo devices, etc.
4. Max length of a file name is 99 characters

Perhaps  --format=v7 should used instead of just -numeric-owner?

lcn2:
Given the benefits of the more restrictive --format=v7, we will use that in the next release (version 0.27 2022-01-17).

We hope this helps.

PR #978, opened 2024-09-24
Sync jparse/ from jparse repo

xexyl:
Sync jparse/ from jparse repo (https://github.com/xexyl/jparse/). The following was added to the jparse CHANGES.md file:

- Implement boolean unicode of struct json_string in the decoding functions. If json_conv_string() finds that calling json_decode() which calls decode_json_string() causes the bool unicode to be false it sets the converted boolean to false. This will then flag an error in parsing if there is an invalid Unicode symbol. This does not mean that decoding all symbols work: it simply means that we detect if there are invalid Unicode symbols. The check is done on the original string but if it turns out it has to be done on the decoded string that can be done easily.

- The json_decode() had a bug fix: it allocated memory for the return string when decode_json_string() does that and this caused a memory leak.

- If decode_json_string() (which is now a static function in json_parse.c) detects an error, the allocated memory is freed before returning NULL.

- Updated JSON_PARSER_VERSION to "1.1.6 2024-09-24".

- Updated JPARSE_REPO_VERSION to "1.0.13 2024-09-24".

- Rename jenc to byte2asciistr in json_parse.c to avoid confusion about its purpose.

- Expand the output of jstrencode -t and jstrdecode -t to express that the encode/decode tests have not yet been written. This depends on bug #13 being resolved first.

- Changed optimisation flags in the Makefiles to not specify -g3 as debug symbols are almost useless when optimising and we have -O3 in use. During debugging one can always use:

    make C_OPT="-g3" clobber all

to compile in debug symbols or have a file in the respective directories (those needed) called makefile.local with the line:

    C_OPT= -g3

which is used for development purposes but should not normally be done.

- Fix potential use without initialisation of inputlen in jstrencode.c.

- Add function decode_json_string() to help simplify the json_decode() function as it's quite long. This new function takes the length and calculated


xexyl:
For some reason the json utf8 files are not being added ....

xexyl:
Oh because you already did it ... hmm so why is it failing? But I have to go afk soon ..

xexyl:
:( ... I just missed some precious moments with one of my dogs due to resolving this problem ... but should be resolved now .. I hope. Workflow works in jparse now.

xexyl:
With the most recent commit here for the version improvements in jparse I believe that I can now focus on the version strings of the tools here and the issue of preparing for the code freeze. ... though I see this commit failed. Okay after the fix then.

xexyl:
Ugh ... I did not think of this!

The `soup/version.h` has the same if guard as the jparse/version.h .. so the tools' version strings cannot be found.

xexyl:
But an odd thing is now the version strings are not being picked up. It seems weird at first since it used to work but that's the odd thing as soup/version.h was never included ...

Unfortunately I have to go for a bit. I can hopefully fix this easily but I cannot do that now and I am not sure if I can do it today as I have several other things I have to get done.

xexyl:
Ah .. the reason it worked before is soup.h is included and that has version.h .. so why is it not picking it up now ?

xexyl:
Answer: I copied the version.h over to soup by accident. I can fix this ...

xexyl:
Now that it is resolved I will be afk .. if I don't finish the version stuff today I'll get to it tomorrow I hope. Then after that I can focus on the code freeze thing and then go back to the other repo.

I think it's quite likely I will not get to the version stuff today. Tomorrow might be a busier day too and Thursday also but I hope to do a bit each day at the least.

xexyl:
It seems I accidentally cut off the log here too .. oh well. I have too many things here that would not be worth trying to redo. If anyone wants to see the full log they can always look at jparse/CHANGES.md.

lcn2:
Thanks @xexyl

PR #983, opened 2024-09-26
Add note about libraries in FAQ

xexyl:
The three libraries found in three repos that are used by this toolkit are copied into this repo to be used so it is not required to have them. The note now clearly states that it is not required to have them installed and even if you do have them installed the Makefiles link in the mkiocccentry's copy.


xexyl:
Have to leave .. hopefully can do more tomorrow.

lcn2 (approved):
OK

PR #984, opened 2024-09-28
Fix entry.reference.json - add new members

xexyl:


xexyl:
And as for the fun file it's now frozen.

I have other things I have to do now but I guess tomorrow I can do more.

PR #986, opened 2024-09-29
Add use of is_available.sh in prep.sh

xexyl:
For certain rules the is_available.sh script is used and if it fails the test is skipped. The tool is not used in the Makefiles (yet?) but it could be. In the case of prep.sh it's mostly for maintainers but since they are not vital tools they can allow others to proceed without having a problem (we can fix it instead).


lcn2 (approved):
👍

PR #989, opened 2024-10-01
Improve is_available.sh - check for checknr(1)

xexyl:
Improve soup/is_available.sh to now verify that checknr(1) is reliable and if it is unreliable or cannot be found, the step is skipped in prep.sh. The original version of checknr(1) always returned 0 even with an error so this check in is_available.sh expects a non-zero value, not 0, in order to verify it works okay, as in order to make sure that it works right we construct a temporary erroneous man page.

The prep.sh here now skips check_man rule if checknr cannot be found or is unreliable.

The Makefiles (except for dbg and dyn_array as these repos do not yet have the script or the changes) have been updated to better explain the problem if is_available.sh reports non-zero.

Sync jparse from the jparse repo (https://github.com/xexyl/jparse/) with the above changes (jparse is where it originated).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,25 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 1.5.22 2024-10-01
+
+Improve `soup/is_available.sh` to now verify that `checknr(1)` is reliable and
+if it is unreliable or cannot be found, the step is skipped in `prep.sh`. The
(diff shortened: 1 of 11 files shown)


xexyl:
I do plan to do this for dbg and dyn_array btw. Adding the script (the new version which I just remembered I forgot to update the version .. will do that now) and then update the Makefiles.

lcn2 (approved):
OK

PR #993, opened 2024-10-03
Update .gitignore

xexyl:
This comes from 'make depend' and was not previously ignored.


SirWumpus:
You probably want to add more general `*.orig` and `*.rej` (?) patch reject files.

xexyl:
> You probably want to add more general `*.orig` and `*.rej` (?) patch reject files.

The first one is an interesting idea for sure ... the other maybe too. We do use patch in one place. I'll do that. Thanks!

xexyl:
Done .. thanks!

lcn2 (approved):
👍

PR #995, opened 2024-10-08
Sync jparse from jparse repo

xexyl:
This syncing should resolve JSON decoding bugs. Although a couple things have to be done still to close that issue
(https://github.com/xexyl/jparse/issues/13) out it appears everything is good now. Some 'good' JSON files were changed to be 'bad' as they have invalid UTF-8 bytes, it seems, but if this turns out to be false (or if the so-called JSON spec allows for accepting it), this can be changed back, especially as some test suites suggest that those files are legal, even though multiple sources say otherwise.

Rather than duplicate the rest of the changes here or in the CHANGES.md file, I refer anyone who must know to the jparse CHANGES.md file or its git log.


xexyl:
Oh. Nice catch! Where does this come from? I'll fix in the other repo and copy over ...

xexyl:
Let's hope this solves it. The warn() should be fine. Unsure about the other one. What bothers me is why the compiler did not flag the one. The other is I hope just due to the casting.

xexyl:
Okay all good. Well I was doing that part as I was waking up. I still would like to know what checked that though because I'd like to add it over there. Do you know?

lcn2 (approved):
thanks @xexyl

PR #997, opened 2024-10-08
Sync jparse/ again

xexyl:
Removed deprecated XXX comments in the jstrencode/jstrdecode -t option.


lcn2 (approved):
`fsync(2)` 🤓

xexyl:
> `fsync(2)` 🤓

If a clumsy person drops their phone into a pool of water or a bathtub full of water is it syncing? Or is it sinking?

lcn2:
> > `fsync(2)` 🤓
> 
> If a clumsy person drops their phone into a pool of water or a bathtub full of water is it syncing? Or is it sinking?

Is is the

```sh
sync
sync
sync
```

that people often typed before using a shutdown on the Vax in the hopes that the disk buffers were all written out.

Except that those filesystem would update the access time of a file, running `sync(1)` would cause the inode of the command to be written.  That and the fact that the shell had to do a search along `$PATH`, accessing those directories and also accessing the root mount point too.  This is why those who knew would turn off `aupdate` on the root filesystem.

xexyl:
> > > `fsync(2)` 🤓
> 
> > 
> 
> > If a clumsy person drops their phone into a pool of water or a bathtub full of water is it syncing? Or is it sinking?
> 
> 
> 
> Is is the
> 
> 
> 
> ```sh
> 
> sync
> 
> sync
> 
> sync
> 
> ```
> 
> 
> 
> that people often typed before using a shutdown on the Vax in the hopes that the disk buffers were all written out.
> 
> 
> 
> Except that those filesystem would update the access time of a file, running `sync(1)` would cause the inode of the command to be written.  That and the fact that the shell had to do a search along `$PATH`, accessing those directories and also accessing the root mount point too.  This is why those who knew would turn off `aupdate` on the root filesystem.  

I didn't know all the history but I of course know of the command and have used it to be sure things were written to disk. 

Thanks for the extra bits (and bytes!) of information! Always fascinating.

lcn2:
People who did a shortcut:

```sh
sync; sync;sync
```

might sing too fast.

Some people wrote a shutdown script that did a `sleep 1` between each `sync(1)` before doing a shutdown.

xexyl:
> People who did a shortcut:
> 
> ```shell
> sync; sync;sync
> ```
> 
> might sync too fast.
> 
> Some people wrote a shutdown script that did a `sleep 1` between each `sync(1)` before doing a shutdown.

We encountered a bug in the MUD based on that too (`cmd;cmd;cmd` etc.) where it would bypass a wait state. Ironically the point of that state was to make them wait after performing a skill or casting a spell but by stacking it like that it got past that! I don't remember the exact fix but I have MANY times written shell scripts that sleep between commands for a variety of reasons.

There are a lot of Easter eggs we put in there too .. some of which I'm quite proud of, others I just discovered (some before the code and some after). Well it was pretty inevitable that I would be good at it and it's unsurprising that I discovered so many things I did that I did before I had the code. 

But these are for another time, maybe. I have to leave. Hope you're no longer contagious very soon!

xexyl:
> ## UPDATE 0
> ... all avoidable by mounting the root filesystem with `noatime`. That also saved **LOTS** of writes to the critical root filesystem. Instead of the root filesystem only writing when software was updated, the drive was "worm out" doing silly things like updating the access time for / .. which happens frequently .. which was a common cause for the critical root filesystem failing.

I can imagine. I guess there's also the wasting of inodes. I disable atime as well.

Anyway .. let me know how this goes in the other repo! Best wishes!

PR #1000, opened 2024-10-09
Fixes in FAQ and {soup,test_ioccc}/README.md

xexyl:


xexyl:
After these are merged I can check the files. I also will try and have a final look over jparse. But I'm not sure if I can do that today. It might not be until tomorrow morning but very possibly before you will do a code freeze. For now I must leave but if I have a bit of time later today I'll take a look. I would not count on that though.

lcn2 (approved):
🍜

xexyl:
> 🍜

You know of our kitchen, right?

If not I suggest you take a look! And make some soup while you're at it. Don't let it get cold!

PR #1002, opened 2024-10-10
Add new option -N to jstrencode and jstrdecode

xexyl:
As requested this new option ignores a final newline in input for encoding/decoding. It does NOT change the JSON encoding/decoding rule that an unescaped newline is not allowed; it simply changes the mechanics of the tools. Further rules for this option are being discussed.


lcn2 (approved):
See the next PR to jparse to complete the -N processing.

PR #1005, opened 2024-10-22
Update MKIOCCCENTRY_ANSWERS__VERSION for IOCCC28

xexyl:


lcn2 (approved):
We should have said "MKIOCCCENTRY_ANSWERS_IOCCC28-1.0" as per the comment.

We will fix this.

xexyl:
> We should have said "MKIOCCCENTRY_ANSWERS_IOCCC28-1.0" as per the comment.
> 

Happens. Though I don't know that I did add the `-1.0`. Anyway might be better to have a number indeed.

> We will fix this.

Thanks.

xexyl:
I also see you synced the other repos.

... but something went wrong. Was it before the other fix?

xexyl:
Oh dear. The failure happens to do with the `-ldbg` and `-ldyn_array` rather than the direct linking in of the libraries in this repo.

Maybe you have an idea on how to resolve that (so I can go take care of some other things that need to be done)?

xexyl:
Oh I know ... I can fix it easily I think.

xexyl:
Fixing ... then will do other things almost certainly for the rest of the day.

xexyl:
Took longer because I misread (bad timing) the failure. Right after I did the commit to jparse the other one failed from the pull. Had to look at it and then I noticed that it was from the pull, not the push.

So hopefully it's resolved now. Was missing `${LD_DIR2}`.

PR #1008, opened 2024-10-24
Sync jparse fixes

xexyl:
The make clobber rule did not remove a temporary file that is part of the test suite (a recent update to it).

The .gitignore had to be updated for the same file that make clobber did not remove.

A name change of a function that happened many aeons ago was not updated in comments and now it has been updated.


lcn2 (approved):
Thank you 🙏 @xexyl

PR #1010, opened 2024-10-30
Bad typo fix in prep.sh, improve test workflow

xexyl:
The typo was that the variable LOGFILE had no G.

The workflow now runs make bug_report. This does run make test again but it seems like a good thing to run as it can help uncover more problems that might not be noticed otherwise. The bug_report.sh script is slow but this seems like a worthy thing to do.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,5 +43,7 @@ jobs:
       run: make slow_release
     - name: make clobber all test
       run: make clobber all test
+    - name: make bug_report
+      run: make bug_report
     - name: make clobber
       run: make clobber
--- a/test_ioccc/prep.sh
+++ b/test_ioccc/prep.sh
@@ -504,11 +504,11 @@ make_action 31 test
 
 # If we have a logfile, count the number of Notice: messages in the logfile
 #
-LOFILE_NOTICE_COUNT=0
+LOGFILE_NOTICE_COUNT=0
 export NOTICE_COUNT
 if [[ -n "$LOGFILE" ]]; then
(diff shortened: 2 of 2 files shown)


lcn2 (approved):
Good catch ‼️

xexyl:
> Good catch ‼️

I don't know which of us wrote that but it's odd that I never saw it before. Very possibly it was a substitute towards the last time that script was updated (I recall making some of the names more consistent). Oh well, at least it's fixed now.

Best wishes with your time away!

xexyl:
An interesting thing with the make bug_report in the workflow though.

```
## RUNNING: git status on the code
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   jparse/jparse.lex.ref.h
	modified:   jparse/jparse.ref.c

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	checknr/
	independ/
	jparse/lex.jparse_.c
	picky/
	seqcexit/

no changes added to commit (use "git add" and/or "git commit -a")

## git status ABOVE
```

That's because of the dependencies. Probably it's not a big deal as everything should be cleaned up afterwards. But still something noteworthy. The workflow script could remove those subdirectories before running `make bug_report` and as for `lex.jparse_.c` I'm not sure about. Might be from `make parser-o` but the file is not seen after running the script (I think) so it seems okay.

Anyway just figured I'd note it even if it doesn't make any real difference.

PR #1019, opened 2024-11-07
Sync jparse/ with some important updates

xexyl:
Unused code (that we thought might be needed but seem to not) was removed.

Extra sanity checks were added to a function.

A file was renamed and this was done here as well to match that.

The CHANGES.md file was corrected to have a missing release so that there's no confusion on the jump in release numbers.


xexyl:
Sorry about this! I might add that the bug report script a while back was updated in jparse with some useful changes. I can look into what those were (it happened in more than one edit) and it might be that they could be o use here too. But only if you want.

Trying to fix the script with the website and if all is good I'll be done for the day.

xexyl:
The updates involved, now that I think on it, some changes that make it easier to parse the bug report log. That might be worth doing?

lcn2 (approved):
👍

PR #1025, opened 2024-11-14
Swapped decode/encode terms again

xexyl:
Synced jparse/ from the jparse repo (https://github.com/xexyl/jparse/). The term encode and decode were swapped back to the original meaning. This problem occurred due to the fact that when focusing for some time on the encoding/decoding bug of \uxxxx code points in the jparse repo the fact that it is a **JSON** encoder was lost in focus. This change also means that terms were swapped here.

In the process of an issue in 'the other repo', some bugs were uncovered in jstrencode(1) but this should not, I believe, affect the timeline of the next IOCCC, fortunately. The website tools have to be updated which I will work on next a bit later on today.


xexyl:
I will do the website in a while ... afk now.

xexyl:
Oh I forgot to run make release before I shut down. Okay I will fix this later. Sorry!

xexyl:
QUESTION 

Should we have a new rule for the workflow that is like the slow one but in the case where it fails it cats the build log?

I think that's a good idea. I will consider doing that later after I fix this issue whilst waiting for make www to do its thing in the other repo.

xexyl:
That stupid problem was fixed. Due to trying to get a commit in soon I forgot I had other changes in the file ... which is unfortunate. But they did help uncover some bugs in the encode tool.

Doing `make www` at the other repo while I do other things now. If all is good I'll commit there. I expect to not do anything else, however, until tomorrow.

lcn2 (approved):
That's quite a set of diffs .. thanks you !

xexyl:
> That's quite a set of diffs .. thanks you !

Sadly yes. That's what can happen when you're so focused on one thing that you lose sight of something else.

Have to check on make www again ... I would not be surprised if there are other problems to work out.

xexyl:
.. so far so good though.

lcn2:
However ...

```sh
$ jstrencode hi
hi
```

To be properly JSON encoded, it should produce:

```
"hi"
```

lcn2:
And:

```sh
$ jstrdecode hi
hi
```

should have produced an error, as hi is NOT a valid JSON encoded string.

And:

```sh
$ jstrdecode '"hi"'
Warning: json_decode: found \-escaped char: "
Warning: main: error while decoding processing arg: 0

```

Is not correct because:

```
"hi"
```

is a proper JSON encoded string.

xexyl:
> However ...
> 
> ```shell
> $ jstrencode hi
> hi
> ```
> 
> To be properly JSON encoded, it should produce:
> 
> ```
> "hi"
> ```

That's true. I wonder when that cropped in.

xexyl:
Actually I seem to recall the second one always did work. Not sure.

xexyl:
Yes the `jstrdecode hi` has worked for a long time, long before the swap. I just did a check from August. Perhaps you'd like to address this? Assuming that it's even right to do it. Maybe a JavaScript example ought to be done.

xexyl:
> And:
> 
> ```shell
> $ jstrdecode hi
> hi
> ```
> 
> should have produced an error, as hi is NOT a valid JSON encoded string.
> 
> And:
> 
> ```shell
> $ jstrdecode '"hi"'
> Warning: json_decode: found \-escaped char: "
> Warning: main: error while decoding processing arg: 0
> ```
> 
> Is not correct because:
> 
> ```
> "hi"
> ```
> 
> is a proper JSON encoded string.

As far as this goes. I recall this being in here for months, perhaps always. Maybe I can look into it but not now.

xexyl:
Okay the problem is it's supposed to be escaped. It's an easy fix but is it right or not to remove that check? I don't know. Please advise.

xexyl:
This diff fixes it btw:

```diff
diff --git i/json_parse.c w/json_parse.c
index 0aa28e5878bf9361a94dcbecafb01df0f9b8a988..a6e183c7567c45f5316849cd25cfe4afece5e539 100644
--- i/json_parse.c
+++ w/json_parse.c
@@ -1700,11 +1700,10 @@ json_decode(char const *ptr, size_t len, size_t *retlen)
                    *retlen = 0;
                }
                warn(__func__, "found non-\\-escaped char: 0x%02x", (uint8_t)c);
                return NULL;
                break;
-       case '"':   /*fallthrough*/
            case '\\':
                /* error - clear allocated length */
                if (retlen != NULL) {
                    *retlen = 0;
                }

```

xexyl:
I think it could be a bigger problem. In some cases it should not be escaped like the example you gave. But in other times it should be like inside a string.

Maybe the decode tool needs an option to pass to the function to decide whether to ignore it or not? Though that could get kind of ugly.

Let me know how you wish to proceed .. or when back open a pull request yourself.

lcn2:
Lets revert

xexyl:
> Lets revert.
> 
> See [GH-issuecomment-2477242730](https://github.com/ioccc-src/temp-test-ioccc/issues/2752#issuecomment-2477242730).

Well I don't see why since everything is going as planned with the updates I have made. Reverting would also lose other changes too, quite a few important ones.

lcn2:
We will be happy to look at the state of things in about 2 days.   Best wishes.

xexyl:
> However ...
> 
> ```shell
> $ jstrencode hi
> hi
> ```
> 
> To be properly JSON encoded, it should produce:
> 
> ```
> "hi"
> ```

With the commit I made a short bit ago you can do:

```sh
jstrencode -s hi
```

and get:

```sh
$ jstrencode -s hi
"hi"
```

As noted elsewhere, the reason for not assuming a string by default is there are many other options and it would be annoying to have to specify an option to disable it in the case that it's not a string; there are, after all, many types, not just one.

xexyl:
> And:
> 
> ```shell
> $ jstrdecode hi
> hi
> ```
> 
> should have produced an error, as hi is NOT a valid JSON encoded string.

As for this ... I think that I have to update the `-s` option to also verify that it starts / ends with a quote. I missed this part or rather I forgot about it.

> 
> And:
> 
> ```shell
> $ jstrdecode '"hi"'
> Warning: json_decode: found \-escaped char: "
> Warning: main: error while decoding processing arg: 0
> ```
> 
> Is not correct because:
> 
> ```
> "hi"
> ```
> 
> is a proper JSON encoded string.

Try:

```sh
$ jstrdecode -s '"hi"'
hi
```

The reason for `-s` is as above.

xexyl:
> > And:
> > ```shell
> > $ jstrdecode hi
> > hi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > should have produced an error, as hi is NOT a valid JSON encoded string.
> 
> As for this ... I think that I have to update the `-s` option to also verify that it starts / ends with a quote. I missed this part or rather I forgot about it.
> 

Oops .. this is also fixed actually. Same option as both are to do with whether or not it's a string. All should be good now unless some other issue is found, plus of course the issues I discovered yesterday with jstrdecode .. or encode, don't remember: was a long day and early morning.

PR #1027, opened 2024-11-16
Somewhat important fixes in bug_report and prep

xexyl:
Rather than assume that date(1) is at /bin/date, just use 'date'.

Minor format fixes/clarifications in FAQ.


lcn2 (approved):
Thanks @xexyl

PR #1031, opened 2024-11-18
Fix critical bug in jparse/test_jparse/Makefile

xexyl:
For linux the test_jparse/Makefile needs -lm for util_test. Without this it will not compile and this will cause a big problem with using the library here.

Other improvements and bug fixes in jparse were made as well that might as well be synced up.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.0.1 2024-11-18
+
+Fix critical compilation error in `jparse/test_jparse/Makefile` for linux.
+
+Other bug fixes and improvements were added to `jparse/` from the jparse repo.
+
+
 ## Release 2.0 2024-11-17
 
 This is a formal release for public use.
--- a/jparse/CHANGES.md
+++ b/jparse/CHANGES.md
@@ -1,6 +1,53 @@
 # Significant changes in the JSON parser repo
 
 
+## Release 2.1.3 2024-11-18
+
+Improve function `parse_json_str()` and `parse_json()` by allowing for empty or
+NULL strings in the following way: `parse_json_str()` will pass a NULL filename
+to `parse_json()` which will keep the filename as NULL, which will, due to
+updates in the error function, not show a file (the difference is that it now
+checks that it's not NULL **AND** not a NUL string).
+Thus to parse a string rather than a file, one can use either
+`parse_json_str()` (a simplified version) or else `parse_json()` with a NULL
+filename. Using an empty filename in the latter function will set it to `"-"` for
+stdin, though it's important to realise that the function `parse_json()` acts on
+a `char *`. If one needs to read in a file, they should instead use
+`parse_json_file()` (if they have only a filename) or else `parse_json_stream()`
+if they have a `FILE *`.
+
+Renamed some internal functions that were too ambiguous with the `parse_json()`
+family of functions. In particular the functions that parse specific JSON types
+like `json_parse_string()` which used to be `parse_json_string()` (too similar
+to `parse_json_str()` which has a very different purpose). All of the
+`parse_json_()` functions were renamed to `json_parse_()` which also match the
+conversion and allocation functions in name format.
+
+Fixed `jparse.3` man page and added a new file (also a symlink like the others)
+for `parse_json_str()`.
+
+Fixed at least one issue with `jstrdecode(1)` by first using the jparse parser
+on the string prior to decoding. This will solve the problem of input like
+`'"foo\"'` not reporting an error (previously it would throw a warning but not
+an error but it is NOT valid JSON). Furthermore something like `'"\"\"\""\'`
+will now report an error:
+
+```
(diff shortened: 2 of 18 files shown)


xexyl:
Sorry about this. I know you wanted to not have any more commits but this seems like a critical problem. I know multiple linux distros do not include the maths library by default. It seems when `util_test` was added (a good thing to have) it broke compilation in linux.

But at least some bugs were fixed and other improvements were made, I guess.

lcn2:
> Sorry about this. I know you wanted to not have any more commits but this seems like a critical problem. I know multiple linux distros do not include the maths library by default. It seems when `util_test` was added (a good thing to have) it broke compilation in linux.
> 
> 
> 
> But at least some bugs were fixed and other improvements were made, I guess.

We will need to do some testing to evaluate and determine the impact.

lcn2 (changes requested):
> Other improvements and bug fixes in jparse were made as well that might as well be synced 

This isn't just adding a link to some maths library for some t st code.  This PR changes 18 files!

As such this PR is changing too much.

Also under what Linux distribution(s) does the lack of -lm break compiling?

xexyl:
The interesting thing: I actually thought of the first part. Will fix soon. Probably all I can do today. 

As far as which systems: at least Rocky but it very likely is also fedora. It also doesn't link in the maths library.

xexyl:
I don't think you need to  issue a new release. 

I almost didn't sync the whole thing but against better judgement I did. 


Nonetheless I will close this soon and in a bit open a new one.

xexyl:
I wonder if the workflows can use Red Hat systems. If so it might help find these problems sooner.  

Will look into that. Closing this and will see if I can figure out why it even needs it. In any case I will open a new pull request with a fix. The `-lm` is probably the smallest change though. 

Closing.

xexyl:
Okay booting laptop up and will look at this then. 

Thanks! 

I agree with you btw: I did almost only sync the Makefile mod. 

Hope you're sleeping well!

xexyl:
Unfortunately I had accidentally committed to master. This means that some other changes might be added unless I can figure out how to do this. I think I have to find out when the last commit occurred in jparse before the code freeze here and then pull that in the clone here. Then I can just sync the test_jparse/Makefile here. I'll look into that in a little while.

xexyl:
Oh ... I forgot the `.git` subdirectory is not synced. I have to come up with something else instead. Hmm ...

xexyl:
I have an idea. I'll clone my fork on the server and then do a commit that way. Will have to configure so I can push but that's not a problem. Will fix soon.

xexyl:
That might not work either ... since it wants the master branch :(

xexyl:
I am sorry but the only thing I can think of is for you to add the change. Then when I pull from the mkiocccentry repo directly it will overwrite my master branch (well you know what I mean).

So the change necessary is:

```diff
b387e32b0a66d jparse/test_jparse/Makefile 
diff --git a/jparse/test_jparse/Makefile b/jparse/test_jparse/Makefile
index 49a092b5c680617c05683049c59389f1e32a7fe6..9ac7ab6bef3055afdfed67f0a369a559149e20eb 100644
--- a/jparse/test_jparse/Makefile
+++ b/jparse/test_jparse/Makefile
@@ -449,11 +449,11 @@ util_test.c: ../util.c
 
 util_test.o: util_test.c
        ${CC} ${CFLAGS} -DUTIL_TEST util_test.c -c
 
 util_test: util_test.o ../libjparse.a
-   ${CC} ${CFLAGS} $^ -o $@ ${LD_DIR2} -ldbg -ldyn_array
+ ${CC} ${CFLAGS} $^ -o $@ ${LD_DIR2} -ldbg -ldyn_array -lm
```

Otherwise multiple Red Hat systems will not be able to compile the mkiocccentry/jparse directory.

Sorry. I can't think of another way except for deleting my fork but I'd really rather not do that. Maybe I can come up with something else sometime and maybe even tomorrow (today seems unlikely) but not today. I mean I could checkout the code freeze release but it doesn't seem to want to work :(

xexyl:
I will test something but if not I'm afraid you're going to have to do it. Stay tuned!

xexyl:
It looks like the idea will still require a merge (due to GitHub mis-feature) of the other commits from yesterday. I have one more idea and if it fails you'll have to make the change, else Red Hat systems probably can't compile the code due to the util_test not working (even though util.o does .. ah maybe that's why?).

xexyl:
Tested one more idea that should (I think) undo the changes on my master branch and then apply to this repo which would mean the changes in the code would not be applied (in the end).

Otherwise you'll have to apply the diff. Sorry.

xexyl:
You're never going to believe this.

It was because of a makefile.local file!

All of this because of that ... well at least the problem is resolved there. My clone is messed up and I might have to wait on another commit before that can be resolved but at least this is resolved.

xexyl:
I also know why the changes did not go through here. It was because it went to the jparse repo itself.

Will try undoing that one.

xexyl:
Everything in both repos should be fixed now, thankfully. Hope you have a great day!

PR #1032, opened 2024-11-19
Fix critical bug in jparse/test_jparse/Makefile

xexyl:
A note is that due to an accidental commit to master branch (in my fork) it is proving a problem to update the jparse/ directory without the changes after the code freeze *here* from the jparse repo. It is hoped that since the state of jparse/, again *here*, is from right before the code freeze *here*, that there won't actually be changes in the mkiocccentry repo so that the only file that changes *there* is the jparse/test_jparse/Makefile. If this does not happen then Landon can add just the change in the test_jparse/Makefile.

This critical fix is that in Red Hat systems the test_jparse/util_test cannot compile and without this Red Hat systems will not compile the jparse library due to the compilation failure.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.0.1 2024-11-18
+
+Fix critical compilation error in `jparse/test_jparse/Makefile` for linux.
+
+Other bug fixes and improvements were added to `jparse/` from the jparse repo.
+
+
 ## Release 2.0 2024-11-17
 
 This is a formal release for public use.
--- a/jparse/CHANGES.md
+++ b/jparse/CHANGES.md
@@ -1,6 +1,53 @@
 # Significant changes in the JSON parser repo
 
 
+## Release 2.1.3 2024-11-18
+
+Improve function `parse_json_str()` and `parse_json()` by allowing for empty or
+NULL strings in the following way: `parse_json_str()` will pass a NULL filename
+to `parse_json()` which will keep the filename as NULL, which will, due to
+updates in the error function, not show a file (the difference is that it now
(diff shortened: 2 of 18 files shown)


xexyl:
Problem is that the other commit seems to be lingering even though I deleted the branch.

If it has to be merged then I guess it will be likely that all the files will be changed. If this seems the case you'll have to do the change. Sorry :( .. I usually check out a new branch first but sometimes I forget.

xexyl:
Trying more thing due to GitHub thinking it's a good idea to have to merge commits even those that are closed (makes no sense to me).

xexyl:
Will leave this open in case you have any ideas but I think it's not going to work unless you or someone else applies the diff :(

xexyl:
See if this solves the problem .. if not you'll have to do the applying of the diff.

xexyl:
Oh no .. I committed in jparse/ itself .. so it's on the master branch there. I'll try again.

xexyl:
An unfortunate thing is that although HEAD is at the code freeze I can't commit the jparse updates.

Also I stupidly deleted the branches I had so I can't use them as no backup had been made yet.

Seems i'm definitely out of luck.

xexyl:
This is weird: I just checked out in Rocky Linux before the fix yesterday and it now compiles (??) ...

Even so my clone is totally messed up unless there is a way in GitHub to roll back the master branch somehow. Maybe git allows that somehow but I'll have to look at that another time - and not today I think.

xexyl:
Closing this as it seems that this might not be necessary after all. If someone reports this as a problem then we can fix it then.

I'll have to figure out my branch another time.

xexyl:
Everything in both repos should be fixed now, thankfully. Hope you have a great day!

PR #1048, opened 2024-12-30
Add missing jparse man pages

xexyl:
The Makefile was also updated to install them. These files were added to jparse a while back but only discovered missing today.


xexyl:
Since there is no code changes and since it's only missing files I figured I'd do this now before the other issue is resolved (though still need clarification .. if you gave it already I did not see it but if you let me know I can check it - tomorrow).

lcn2 (approved):
👍

PR #1049, opened 2024-12-31
Improve jparse error messages in yyerror().

xexyl:
In particular, instead of surrounding the bad token with <>s, which could be one of the invalid chars and which might make someone think that that's the problematic character, do not surround it any more. Since it is by itself after the ':' (and the next char is the newline) it should be clearer this way.

Updated JPARSE_VERSION and JPARSE_LIBRARY_VERSION to "1.2.8 2024-12-31" and "2.2.3 2024-12-31" respectively.

NOTE: there is ABSOLUTELY NO functional change in this, just a display improvement, and so it's a useful change to include here too.


xexyl:
I aim to look at the issue about submit server soon ... then when I can get that in maybe (unless you discover something else?) we can have another code freeze before the contest opens? Well we'll see I guess.

xexyl:
BTW: I don't think it will affect us (though I didn't look in great detail .. though it does seem to affect a lot of people anyway) but I got this warning in workflow in jparse when I forgot to update the error location files (to make sure error location testing works, if you recall):

> build
> ubuntu-latest pipelines will use ubuntu-24.04 soon. For more details, see https://github.com/actions/runner-images/issues/10636

We'll see but thought you might want to know!

xexyl:
And I see a failed check .. hmm .. fixing this first.

xexyl:
Apparently this repo needs a rule to rebuild error location files too .. I forgot we have additional error files here! Will fix.

xexyl:
Display problem .. trying to fix with my bleary eyes. The rule already existed but needed to be updated. Unfortunately it looks funny when running from this repo ... there might be another way to handle it though. 

After this is solved I hope to look at the other issue, the top priority one .. this one seemed pretty useful though and since I discovered it I wanted to fix it before I forgot.

xexyl:
.. discovered another potential problem .. will require a commit to jparse and to this repo too. The display issue I have an idea of what it might be but it only happens here and only from the top level Makefile rule. I guess it's not worth trying to figure out for now as only we would run it and more than likely only I would run it as it only has to be done if and only if jparse error messages changed. And that is extremely rare and it's very unlikely it'll change again: if it does it'll not be for a long time I guess.

xexyl:
Oh I found the problem ... it was a typo in the Makefile in jparse/test_jparse! I'll fix that along with the other change and then merge it here and do the fix as well .. so this should be fine soon.

xexyl:
The commit I just made should theoretically solve the problems. I need to focus my eyes and then I can hopefully look at your comment at the other issue and resolve the two top priority issues, hopefully allowing for a code freeze! Then after that I can look at the issues in the winner repo!

xexyl:
Oops .. let's try again!

xexyl:
Oh bugger me .... the problem is the path to jparse in the error files and to do with the way jparse(1) will show the program name. Since I can put party.json in jparse/test_jparse/test_JSON/bad_loc (via jparse repo) I can get rid of the directory here as that's the only file. Since `make test` here runs `make test` in `jparse` it is fine.  I'll do that ... making some final tests and preparations.

xexyl:
Seems I got it .. have to do a commit over there and then I have to do the commit here and it should be okay!

xexyl:
Almost there .. had to go afk a bit. Made commit at jparse repo ... now recreating clone here and then will do make all test and if all is good I'll commit.

xexyl:
This should do it. Only one minor functional difference here and the benefit is better error reporting in jparse.

I'll next look at the other issue ... though not sure if that's right away or not.

lcn2 (approved):
Happy New Year 🥳‼️

And thanks 🙏 @xexyl

xexyl:
> Happy New Year 🥳‼️
> 
> And thanks 🙏 @xexyl

Happy New Year to you too!

Whatever else might or might not go: at least we're at the point where the IOCCC can run again!

I'm going to try looking at the other repo but I'm not sure what I'll get done today. I'm rather tired. I do know one thing though that should be done and I believe I did it but I have to verify it. Anyway that's for over there.

Does this solve the two top priority issues?

lcn2:
> > Happy New Year 🥳‼️
> 
> > 
> 
> > And thanks 🙏 @xexyl
> 
> 
> 
> Happy New Year to you too!
> 
> 
> 
> Whatever else might or might not go: at least we're at the point where the IOCCC can run again!
> 
> 
> 
> I'm going to try looking at the other repo but I'm not sure what I'll get done today. I'm rather tired. I do know one thing though that should be done and I believe I did it but I have to verify it. Anyway that's for over there.
> 
> 
> 
> Does this solve the two top priority issues?

Very close, @xexyl.  See our suggestion in [GH-issuecomment-2566672002](https://github.com/ioccc-src/mkiocccentry/pull/1050#issuecomment-2566672002).  With that we think we can release a new mkiocccentry repo.

BTW: We suggest that you call that repo version 2.2.

PR #1059, opened 2025-01-01
jparse related fixes

xexyl:
Fix exit code errors in jparse_test.sh.

Remove rule rebuild_jparse_err_files from top level Makefile. This rule should only be in jparse/. The reason it would cause a failure is because of argv[0] being in error messages and since in this repo jparse binary is in a subdirectory it changes the name of the program so the error files are incorrect. It might be possible to do a 'cd jparse && ...' but this is not necessary as there are no error location files in this repo itself due to precisely this reason.

Happy New Year!

### DIFF
--- a/jparse/Makefile
+++ b/jparse/Makefile
@@ -830,8 +830,7 @@ rebuild_jparse_err_files: jparse
 	${S} echo "${OUR_NAME}: make $@ starting"
 	${S} echo
 	${Q} ${RM} ${RM_V} -f test_jparse/test_JSON/bad_loc/*.err
-	-${Q} for i in test_jparse/test_JSON/./bad_loc/*.json; do \
-	    echo './jparse -v 0 -- "'$$i'" 2> "'$$i'.err"' ;  \
+	-@for i in test_jparse/test_JSON/./bad_loc/*.json; do \
 	    ./jparse -v 0 -- "$$i" 2> "$$i.err" ;  \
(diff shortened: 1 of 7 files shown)


jparse/Makefile
```diff
@@ -830,8 +830,7 @@ rebuild_jparse_err_files: jparse
 	${S} echo "${OUR_NAME}: make $@ starting"
 	${S} echo
 	${Q} ${RM} ${RM_V} -f test_jparse/test_JSON/bad_loc/*.err
-	-${Q} for i in test_jparse/test_JSON/./bad_loc/*.json; do \
-	    echo './jparse -v 0 -- "'$$i'" 2> "'$$i'.err"' ;  \
+	-@for i in test_jparse/test_JSON/./bad_loc/*.json; do \
```
lcn2, line 833: Out of curiosity, why do you want to hide the actions being performed?

Another curiosity is the use of explicit '@' instead of '${Q}' (that allows one to turn default quiet mode off).

We are sure there is a reason, we are just curious why.

If this was a mistake, please issue another PR to restore this as we will accept this PR as is for now.

lcn2 (approved):
Thanks @xexyl 

But do review [GH-discussion_r1900443973](https://github.com/ioccc-src/mkiocccentry/pull/1059#discussion_r1900443973) when you have the opportunity to do so.

PR #1063, opened 2025-01-03
Add back missing debug message in jparse/

xexyl:


xexyl:
Thanks for pointing this out.

lcn2 (approved):
👍

lcn2 (changes requested):
The for loop above still uses –@ instead of -${Q}

Please fix that for loop as well so that it quiet mode could be turned off by the Q make value, @xexyl 

This was the second part of the debug code that we put into this make action.

xexyl:
> The for loop above still uses –@ instead of -${Q}
> 
> Please fix that for loop as well so that it quiet mode could be turned off by the Q make value, @xexyl
> 
> This was the second part of the debug code that we put into this make action.

Oops. Thanks. That was a copy paste error.

xexyl:
Does that commit handle it ?

lcn2 (approved):
Thank you @xexyl

PR #1067, opened 2025-01-08
Various updates for subdirectories in mkiocccentry

xexyl:
Update mkiocccentry_test.sh to test the new subdirectory option of mkiocccentry, with the depth being valid and the depth being too high.

Fix regexp in error message for basename of files according to how sane_relative_path() functions.

Sync jparse/ from jparse repo (https://github.com/xexyl/jparse/) with a comment to sane_relative_path() (about regexp it enforces) and additional test cases to the util test code.

Updated FAQ about how to use mkiocccentry as well as the man page (with some details on how the mkiocccentry will copy directories now). The FAQ, guidelines and rules have to be updated in the winner repo and that'll come next unless something comes up here.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,23 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.5 2025-01-08
+
+Update `mkiocccentry_test.sh` to test the new subdirectory option of
+`mkiocccentry`, with the depth being valid and the depth being too high.
+
+Fix regexp in error message for basename of files according to how
+`sane_relative_path()` functions.
+
+Sync `jparse/` from [jparse repo](https://github.com/xexyl/jparse/) with a
+comment to `sane_relative_path()` (about regexp it enforces) and additional test
+cases to the util test code.
+
+Updated FAQ about how to use `mkiocccentry as well as the man page (with some
+details on how the mkiocccentry will copy directories now). The FAQ, guidelines
+and rules have to be updated in the winner repo and that'll come next unless
+something comes up here.
+
 ## Release 2.3.4 2025-01-07
 
 Submission tarballs can now have subdirectories! This is a major release that
--- a/README.md
+++ b/README.md
@@ -42,3 +42,12 @@ answers to a number of possible problems as well as to see how to report issues.
 If your problem is a problem compiling the tools or you have encountered what
 you think is a bug, please see the
 FAQ on "[reporting bugs](FAQ.md#bugs)".
+
+Additionally, if you have more questions on the use of the toolkit, see the FAQ
+section [Entering the IOCCC: the bare minimum you need to
+know](https:/www.ioccc.org/faq.html#enter_questions), the FAQ section [Entering
+the IOCCC: more help and
+details](https://www.ioccc.org/faq.html#submitting_help), the [IOCCC
+Guidelines](https://www.ioccc.org/next/guidelines.html) and the [IOCCC
+Rules](https://www.ioccc.org/next/rules.html) at the [Official IOCCC
+website](https://www.ioccc.org).
--- a/jparse/CHANGES.md
+++ b/jparse/CHANGES.md
@@ -1,5 +1,16 @@
 # Significant changes in the JSON parser repo
 
+## Release 2.2.4 2025-01-08
+
+Add comment about the regexp that `sane_relative_path()` should enforce, namely:
+
+```regexp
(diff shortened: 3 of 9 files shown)


xexyl:
I hope to look at the rules and guidelines today but if not then hopefully tomorrow. Once that is done I can look at the 'polite' words list. I also have a minor update to 2020/ferguson1 in mind though - one you asked me to delay until after the great fork merge.

lcn2 (changes requested):
The use of :alnum: is not correct. 

Their interpretation depends on the LC_CTYPE locale; for example, ‘[[:alnum:]]’ means the character class of numbers and letters in the current locale.

We recommend reverting this back to the explicit ASCII character range.

xexyl:
> The use of :alnum: is not correct.
> 
> Their interpretation depends on the LC_CTYPE locale; for example, ‘[[:alnum:]]’ means the character class of numbers and letters in the current locale.
> 

Good point.

> We recommend reverting this back to the explicit ASCII character range.

I can try but the additional requirements was getting me. Perhaps you can provide one?

SirWumpus:
Would it be simpler to just `setlocale(LC_ALL, "C")` in `main()` to assert the desired locale?  Then `[:alnum:]` would make sense.

xexyl:
> Would it be simpler to just `setlocale(LC_ALL, "C")` in `main()` to assert the desired locale? Then `[:alnum:]` would make sense.

Well no because it's an error message. And strangely changing `:alnum:` to the other form does not work ... so I can't do it right now.

xexyl:
.. and on that note I'll likely have to look at the guidelines etc. tomorrow, not today. About to go do other things and not feeling the best. It was a hard night and my sinuses are really bothering me. Hopefully no infection (even though it's much worse than usual) but see the Ents tomorrow! (as the joke goes.)

I'll in any case be clearer headed later on.

Even so @lcn2: given that this is just an error message does it really matter that much? We could put in comments that it refers to a specific locale.

lcn2:
> Even so @lcn2: given that this is just an error message does it really matter that much? We could put in comments that it refers to a specific locale.

As this is an error message, and it would then differ with the guidance with "the other repo", we request that another PR be submitted to change the error messages.

lcn2 (approved):
For the sake of making progress, we accept this PR with the hope that a future PR will be issued to correct and restore the regular expression in the error message as well as to correct the `jparse/CHANEGS.md` file.

xexyl:
> For the sake of making progress, we accept this PR with the hope that a future PR will be issued to correct and restore the regular expression in the error message as well as to correct the `jparse/CHANEGS.md` file. 

So see my comment. I will do that as well as in the website. 

I will mention that it's in the C locale too. 

But I have to not type now. Using my neb. Lungs tight. 

Will try doing this in a while.

lcn2:
> I have to update it over there anyway and I already planned on it. But I will change the message too to say that it depends on the C locale.
> 
> Does that work?

It is probably better and simpler to just use `[0-9A-Za-z]` and not bring up the complexity of the locale.

lcn2:
> Would it be simpler to just `setlocale(LC_ALL, "C")` in `main()` to assert the desired locale?  Then `[:alnum:]` would make sense.

Now, @SirWumpus, we can't help but picture the "Knights of say Nee" requiring us to cut down the tallest tree 🌲 in for forest with a Herring Locale 🐟.  😆

SirWumpus:
> > Would it be simpler to just `setlocale(LC_ALL, "C")` in `main()` to assert the desired locale?  Then `[:alnum:]` would make sense.
> 
> Now, @SirWumpus, we can't help but picture the "Knights of say Nee" requiring us to cut down the tallest tree 🌲 in for forest with a Herring Locale 🐟. 😆

[Fishie Chowder](https://www.youtube.com/watch?v=svxiWwF-fyk)

SirWumpus:
What editor/tool you using?  You probably need to escape the search pattern, since `[` and `]` are special.
```
\[:alnum:\]
```
BUT if your are using `vi(1)` or `sed(1)` which historically uses BRE then `[:0-9A-Za-z:]` should be fine to replace.  Not sure what `vim` defaults to, but I think there is an option for ERE.  GNU sed I think supports `-E`.

xexyl:
> What editor/tool you using? You probably need to escape the search pattern, since `[` and `]` are special.
> 
> ```
> \[:alnum:\]
> ```
> 
> BUT if your are using `vi(1)` or `sed(1)` which historically uses BRE then `[:0-9A-Za-z:]` should be fine to replace. Not sure what `vim` defaults to, but I think there is an option for ERE. GNU sed I think supports `-E`.

I did an echo | grep test.

PR #1074, opened 2025-01-13
New utility functions, allow inventory.html

xexyl:
Sync jparse repo (https://github.com/xexyl/jparse/) to jparse/ with new utility functions that act on directory paths. These functions are:

- dir_name(): takes a path and strips off level components (i.e. /). In the case of successive /s it removes those as well. The comments at the top of the function explains in more detail the way the level works. This is modelled after the base_name() function which functions as the basename(3) function; dir_name() functions as dirname(3).
- count_comps(): counts in a string the number of components delimited by the component character (a char). Successive component characters are counted as one. The comments at the top of the function details specifics.
- count_dirs(): using count_comps() (with component /), count the number of directory components in a path.

Updated the jparse utility test code to test the new functions.

Remove blacklisting of inventory.html file. This involved a new version of mkiocccentry, txzchk and chkentry as all of these tools prevented such files from existing.


lcn2 (approved):
👍

PR #1076, opened 2025-01-14
Improve and rename test_extra_file()

xexyl:
Renamed test_extra_file() to better account for what it is: test_extra_filename(). Improved the function to use two char *[]s so that we don't have to have checks on each file manually in each tool that uses it, though at present only chkentry(1) uses it (it is not clear if any other tool can use it in the future but this update allows for more easily maintaining the list of filenames that must be present and that must not be present).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,17 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.9 2025-01-14
+
+Renamed `test_extra_file()` to better account for what it is:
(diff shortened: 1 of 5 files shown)


xexyl:
As I noted in a comment it seems that `make test` is failing due to the max depth and txzchk. I asked (marked as `QUESTION`) in a comment what you prefer to be done with this. But if the make test fails here this is why and it should not be because of the code here .. though given that another commit was made without any issues maybe it's something on my end.

Anyway must leave now.

lcn2 (approved):
👍

PR #1078, opened 2025-01-16
Fix warnings for mkiocccentry -d and -s seed

xexyl:


xexyl:
Okay afk a while now .. when back will try and look at documentations comment.

lcn2 (approved):
Thank you image

xexyl:
> Thank you image

You are most welcome!

# QUESTION

Did you update the mkiocccentry_test.sh script to do this? If not does it need to be ? If yes what should it do ?

lcn2:
> QUESTION
> Did you update the mkiocccentry_test.sh script to do this? If not does it need to be ? If yes what should it do ?

No we didn't update and yes, if needed please do, @xexyl

xexyl:
> > QUESTION
> > Did you update the mkiocccentry_test.sh script to do this? If not does it need to be ? If yes what should it do ?
> 
> No we didn't update and yes, if needed please do, @xexyl

Question is how should it test it ? Well it might not need the `answers` file that we use? Actually I think it could just do an additional test using the feature: keep the others in as they have specific things but also add a new one? Does that sound good?

Okay afk now ... back later.

lcn2:
> Actually I think it could just do an additional test using the feature: keep the others in as they have specific things but also add a new one? Does that sound good?

Yes

PR #1080, opened 2025-01-16
Blacklist file GNUmakefile in submissions

xexyl:
Added to list of forbidden files GNUmakefile. This involved an update to the forbidden_filenames array in soup/entry_util.c and making both txzchk and mkiocccentry use that array. This means that the MKIOCCCENTRY_VERSION was bumped again, making it "1.2.1 2025-01-16". The TXZCHK_VERSION was updated to "1.1.3 2025-01-16". The test JSON files were updated for both of these version updates. A new bad test file for txzchk was added along with its appropriate err file.


lcn2 (approved):
👍

PR #1082, opened 2025-01-16
Fixes with mkiocccentry -s seed/-d

xexyl:
Make mkiocccentry_test.sh test -d option (which uses -s seed). Updated
version of script to "1.0.4 2025-01-16".

Removed blank lines from mkiocccentry.1 (they're not needed and some say it's
incorrect).

Make mkiocccentry random answers file prepend http://example.com/ (and
https://example.com/ to the URLs. The emails probably should be @example.com
too but that can be worried about another time.


lcn2 (approved):
thank you  @xexyl

PR #1083, opened 2025-01-17
Bug fix txzchk for required files

xexyl:
Bug fix txzchk to check for required files in top level directory only (with the exception that dot files are not allowed in any subdirectory and the only dot files allowed in the top level directory are .info.json and .auth.json).

Added new test files to txzchk test suite, both good and bad (with associated .err files).

This fix involved a bug fix and enhancement to the jparse library (not the JSON parser but the util.c file) which involved syncing the jparse repo again.

Updated TXZCHK_VERSION to "1.1.4 2025-01-17".


xexyl:
Fixed the problem here ... a funny one! I still don't understand why having the same workflows over there as here that the code was not flagged.

Of course we already knew at this point the pointer was not NULL and we just check for NULL before freeing but GitHub's point is fair enough and better to be checked.

lcn2 (approved):
👍

PR #1084, opened 2025-01-17
Add JPARSE_UTILS_VERSION string for util.c version

xexyl:
Added new macro JPARSE_UTILS_VERSION set at "1.0.0 2025-01-17". This is
strictly so that the JSON parser library version is not changed when something
not strictly related the jparse JSON parse(r) related routines are changed. In
other words, json_util.c is a jparse library version (like jparse.l and
jparse.y) change but a change in util.c is a change in the
JPARSE_UTILS_VERSION. The tools (in jparse and here) now refer to this 
in both -h and -V.

As JPARSE_LIBRARY_VERSION was already changed today I have left that
change in but from now on a change in util.c should update the
JPARSE_UTILS_VERSION (in version.h).


lcn2 (approved):
👍

PR #1088, opened 2025-01-18
Increment OPTIONAL_SUBMISSION_FILES by 1

xexyl:
As try.alt.sh is now included as an optional file there are three optional files, not 2.

This does increase the total number of files included but as optional files are not extra files it is not a problem. Besides the new total should be 39, a prime.


lcn2 (approved):
👍

PR #1092, opened 2025-01-18
Bug fix bug_report.sh to check Makefile.local

xexyl:
Bug fix bug_report.sh to check for Makefile.local, not makefile.local.
This same change was done in other files that still referred to it with the
exception of this file and .gitignore. It was done in dbg/ and dyn_array/ but
the change needs to be done in the repos too.


xexyl:
I am glad I caught this. I'll see about doing this quickly in the other two repos and then I really will be off until tomorrow.

lcn2 (approved):
Good catch @xexyl

xexyl:
> Good catch @xexyl 

Thank you. It's curious that the jparse one had another bug as well but this issue was never a problem with jparse. 

This fix though is quite important. I can easily imagine someone having an overriding Makefile and we don't even know it. Could be scratching our heads wondering why they are having a certain issue with something that we can't duplicate all because they have overridden something but we don't know it due to a different bug!

I guess that makes sense to you. I hope so. Exhausted. Sleep well when you do!

PR #1093, opened 2025-01-19
Fix inspect_Makefile() in mkiocccentry for rule 17

xexyl:
If someone uses a built-in Makefile rule such as .NOTPARALLEL it could make the function detect 'all' rule as not the first rule which puts the submitter in a position of having to override the warning which might put them at a great risk of violating rule 17. As there are other built-in Makefile rules it simply checks for the first character being a '.'.

Also added missing not_reached() calls to noprompt_yes_or_no().


xexyl:
How I caught this is the recent update to the mkiocccentry_test.sh script where it flagged the Makefile as not having 'all' rule first.

I saw it back then but it never dawned on me until today that this could be a big problem.

It also skips variable lines though I didn't make that clear in the log by accident. This was done because some variables also have a `:` in it!

xexyl:
Hmm .. that check I could have sworn I protected against. Sigh. Will fix but I really should be going ...

xexyl:
Okay this is one of the most idiotic things I've seen:

Screenshot 2025-01-19 at 07 18 09

Not even sure how to fix this to satisfy GitHub ... I have a thought but not sure. it seems like it might not understand short-circuiting? I'll try changing it.

xexyl:
Let's see if this solves it .. if so it means GitHub does not understand short-circuiting. It's the only thing I can think of but if it's the case it is crazy.

xexyl:
Hopefully this fixes it .. afk now. If it does not I"ll just remove the original check for `line` as we're already passing it to `strchr(3)` anyway.

I hope to look at the other issue when back.

xexyl:
Okay so that's a bug or mis-feature of GitHub's advanced security bot. 

I find that amusing, bemusing, annoying and rather crazy (or at least very silly). 

Oh well. At least this important bug here has been fixed. 

I will be back later on today and I hope that I can start the interface for the simplified command line. Still have to decide how to structure the ifdefs. 

Well anyway we'll see. Hopefully it will help you with it when you're at it (I do not see how I can finish it in time but I will try).

lcn2 (approved):
thanks, @xexyl for working thru the issue where GitHub's bot is a bit "botty"

xexyl:
> thanks, @xexyl for working thru the issue where GitHub's bot is a bit "botty"

I would have used the words 'batty' and maybe 'ridiculous' as well. But point taken well enough!

xexyl:
Of course far more important is the fix here. Perhaps I should report this issue to GitHub but I really can't be bothered. 

I hope that I can do something here today but it is currently not looking very likely I am afraid.

PR #1095, opened 2025-01-20
More work on mkiocccentry command line enhancement

xexyl:
Add to the macro MKIOCCCENTRY_DEV so that for now it at least will run the tests on prog.c, Makefile and remarks.md and form the tarball even though it's not complete (it won't ignore the right files, it doesn't do anything with the other files etc. but this way at least the program won't fail when working on the issue).

Changed MKIOCCCENTRY_VERSION to "1.2.4 2025-01-20".

Change mkiocccentry_test.sh to not test -d as it now sets -E which causes a failure.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@ the `test_ioccc/test_JSON` tree.
 
 Fixed a number of broken tests in the `test_ioccc/test_JSON` tree.
 
-Updated and sorted `.gitignore'.
+Updated and sorted `.gitignore`.
 
 Fixed debug message that incorrectly stated that was unknown.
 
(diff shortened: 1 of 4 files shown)


xexyl:
This is all I'll do today but it is helpful at least ... I also have a question but I'll put that elsewhere.

lcn2 (approved):
Thanks for making progress, @xexyl 

It is also nice that such a change to the version of `mkiocccentry(1)` does not have to also edit hundreds of test files too.

xexyl:
> Thanks for making progress, @xexyl

I hope it helps! It's about all I can do today I guess but hopefully tomorrow will be better.

> 
> It is also nice that such a change to the version of `mkiocccentry(1)` does not have to also edit hundreds of test files too.

Yes that's so much better! Thanks!

Best wishes on progress with the two tools!

PR #1096, opened 2025-01-21
Typo fix and update comments in fnamchk

xexyl:


xexyl:
The rationale behind this btw is a great portion of it I did for txzchk etc. But of course if you'd rather not I can just do the typo fix instead. Up to you. I'm going to try and reply to comment soon. I couldn't yesterday. I fell asleep and I also went to bed early.

Hope you're sleeping well.

lcn2 (approved):
👍

lcn2 (changes requested):
Removing a copyright and copyright date is not the best practice.

Including yourself into the credits are welcome and important, however.

lcn2:
With commit 1df31267f31986062f87e7ee67c3a2a1d730fb03 we propose a better way to achieve joint credit.

Instead of removing the copyright and license, we
modify the copyright to include both authors.
Moreover as a fair mix, the order of the copyright authors
is the reverse order to the ".. tool was co-developed ..".
The quote at the bottom was moved to the top.

We propose this as a better model for jointly developed code.

xexyl:
> With commit 1df31267f31986062f87e7ee67c3a2a1d730fb03 we propose a better way to achieve joint credit.
> 
> 
> 
> Instead of removing the copyright and license, we
> 
> modify the copyright to include both authors.
> 
> Moreover as a fair mix, the order of the copyright authors
> 
> is the reverse order to the ".. tool was co-developed ..".
> 
> The quote at the bottom was moved to the top.
> 
> 
> 
> We propose this as a better model for jointly developed code.
> 
> 
> 
> ## UPDATE 0
> 
> 
> 
> If the above is acceptable for you, @xexyl, then we plan to roll out such changes in other files.  But this is a very low priority.

I actually like it a lot!

Do you think it should be done with chkentry and the jparse files? If so I have to also update the private tool but that is okay. 

The good news is my F key is working now. Bad news is for a time the Q key stopped working right. Now my left finger is feeling very weird. That haptic feeling is awful. 

But I hope it settles soon (it does seem like it might be okay - hopefully it stays that way). 

Anyway:

# QUESTION 

Do you want me to update the other files (quite a few but that's okay) to match this style, @lcn2  ?

Let me know and I will address it!

xexyl:
Ah I see your update. 

Sure. But to do that I have to also update the private tool and regenerate the files. 

In that case I might as well just do it. 

The one that will take longer is all the jparse files but it can be done anyway.

xexyl:
Actually right now I can't deal with it .. but I can try later today hopefully. Not feeling well unfortunately. Of course if you wish to do it you could also do it as a pull request in jparse and I could then only do the private tool. But that would go with chkentry as that's where it belongs.

I hope to in a bit address the guidelines and rules a bit bit I'm not sure how that's going to go. I'll check a discussion over in the winner repo if I can too.

lcn2:
> QUESTION
> Do you want me to update the other files (quite a few but that's okay) to match this style, @lcn2 ?
> 
> Let me know and I will address it!

Yes please

xexyl:
> > QUESTION
> > Do you want me to update the other files (quite a few but that's okay) to match this style, @lcn2 ?
> > Let me know and I will address it!
> 
> Yes please

Sure .. don't think I can do that now but I actually have a question on this. For chkentry it's easy enough - except the jparse part. So what should it say? For example: for the jparse files should it be something like?

```
diff --git i/jparse_main.c w/jparse_main.c
index e26ec61ead2e3b9d7c7ff002540f9485a0978c7d..5eaf1765c9d423d1ea763262ee4dd1624c15ec62 100644
--- i/jparse_main.c
+++ w/jparse_main.c
@@ -1,13 +1,34 @@
 /*
  * jparse_main - main function for the JSON parser stand-alone tool
  *
  * "Because specs w/o version numbers are forced to commit to their original design flaws." :-)
  *
- * This JSON parser was co-developed in 2022 by:
+ * Copyright (c) 2022-2025 by Landon Curt Noll and Cody Boone Ferguson.  All Rights Reserved.
  *
- * @xexyl
+ * Permission to use, copy, modify, and distribute this software and
+ * its documentation for any purpose and without fee is hereby granted,
+ * provided that the above copyright, this permission notice and text
+ * this comment, and the disclaimer below appear in all of the following:
+ *
+ *       supporting documentation
+ *       source copies
+ *       source works derived from this source
+ *       binaries derived from this source or from derived source
+ *
+ * THE AUTHORS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+ * INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN NO
+ * EVENT SHALL LANDON CURT NOLL BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+ * CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF
+ * USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+ * OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+ * PERFORMANCE OF THIS SOFTWARE.
+ *
+ * This JSON parser and tool were co-developed in 2022 by Cody Boone Ferguson
+ * and Landon Curt Noll:
+ *
+ *  @xexyl
  *     https://xexyl.net               Cody Boone Ferguson
  *     https://ioccc.xexyl.net
  * and:
  *     chongo (Landon Curt Noll, http://www.isthe.com/chongo/index.html) /\oo/\
  *
```

Also how do you wish me to address txzchk and bug_report.sh and others I wrote mostly (or entirely) on my own? Of course as you say also low priority but answering this (when you can) would be helpful.

Perhaps it doesn't happen until after the next IOCCC which would be okay I think. This might even be better as then we can focus on other more important things. But how does the above look?

If that looks good it can be done similar in chkentry I guess?

Best wishes on the submit server today! Off to do other things.

lcn2:
> Perhaps it doesn't happen until after the next IOCCC which would be okay I think. This might even be better as then we can focus on other more important things. But how does the above look?

Yes, it's not a high priority and can be addressed after the IOCC28 opens or closes as needed. 

> If that looks good it can be done similar in chkentry I guess?

Yes.

PR #1102, opened 2025-01-26
More work on #1070

xexyl:
The check on files being found now is done from the directory that the user specifies. This involved an enhancement to the jparse function sane_relative_path(). That enhancement allows filenames to start with "./" as when walking from "." the function prepends "./" to each filename. There very possibly will be a new enhancement/utility function required before the filenames can be written to the .info.json file but that will have to be determined later.

A memory error was fixed in the jparse function count_comps().

Updated SOUP_VERSION to "1.1.13 2025-01-26".
Updated MKIOCCCENTRY_VERSION to "1.2.6 2025-01-26". Updated TXZCHK_VERSION to "1.1.6 2025-01-26".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,23 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.16 2025-01-26
+
+More work on #1070: the check on files being found now is done from the
+directory that the user specifies. This involved an enhancement to the jparse
+function `sane_relative_path()`. That enhancement allows filenames to start with
+`./` as when walking from `.` the function prepends `./` to each filename. There
+very possibly will be a new enhancement/utility function required before the
+filenames can be written to the .info.json file but that will have to be
+determined later.
+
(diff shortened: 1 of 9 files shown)


xexyl:
This is probably all I can do today ... extremely exhausted unfortunately. But at least one issue is resolved even if other related issues have to be resolved later on.

lcn2 (changes requested):
Due to check failures, this PR as it is cannot be processed.

xexyl:
The iOS 18 bug of not notifying is so annoying.

And this GitHub check that is dubious is equally as annoying. I'll fix.

xexyl:
If it works in jparse now I'll do it here too ... then leaving again.

xexyl:
Let's see if this solves it ...

xexyl:
> The iOS 18 bug of not notifying is so annoying.
> 
> And this GitHub check that is dubious is equally as annoying. I'll fix.

Okay so in this case it's slightly less dubious but even so it seems it has in some cases a context issue. Anyway all good now.

PR #1105, opened 2025-01-28
Fix new utility is_* functions

xexyl:
They should not use strncasecmp() but rather strcasecmp() as otherwise a filename like try.shh would be flagged as an optional filename when it's not.


lcn2 (approved):
👍

PR #1106, opened 2025-01-29
More work on #1070

xexyl:
The collect_topdir_files() now verifies that all three required files (prog.c, Makefile and remarks.md) are found.

The collect_topdir_files() also only records extra files if it's not an optional file or a required file.

Updated MKIOCCCENTRY_VERSION to "1.2.8 2025-01-29".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,19 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.18 2025-01-29
+
+More work on #1070.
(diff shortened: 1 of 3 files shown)


xexyl:
I have another good update here ... doing `make prep` first. If all good I'll commit. Then probably have to do other things for the rest of the day. I still have to update the man page here for this commit but that can come later. Too tired to do more at this time.

lcn2 (approved):
👍

PR #1107, opened 2025-01-29
Removed MKIOCCCENTRY_DEV macro

xexyl:
Instead it now defaults to the new way but with the ability to (for now) 
specify extra files as additional args. This is not how it will remain 
but this change allows for easier development. The mkiocccentry_test.sh
script was updated to account for this. The warn_*() functions no longer
take a char const * of the file they are warning against as it's
required to be the specific names so there is no need to repeat it (plus
it's called in multiple locations and we do not always have the path).

This means that make test can work in the new way though later the script will
have to be changed again.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,19 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.18 2025-01-29
+
+More work on #1070.
+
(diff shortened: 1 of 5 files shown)


xexyl:
The man page does need to be updated of course but that can happen later on. As long as the tests here go okay I'll likely be doing other things.

lcn2 (approved):
ok

PR #1108, opened 2025-01-29
Fix debug output in txzchk

xexyl:
This comes from the fact that subdirectories can now be in tarballs. It used to be that only the basename was added to the list but now it's the full path.

Updated TXZCHK_VERSION to "1.1.7 2025-01-29".


lcn2 (approved):
👍

PR #1109, opened 2025-01-30
More work on #1070

xexyl:
The files (unless try.sh or try.alt.sh but see below) copied are now read-only (as in -r--r--r--) and directories are drwxr-xr-x) and txzchk(1) now verifies this. In order to get tests to work the copying of extra files is not done so technically the try.sh and try.alt.sh part is for future reference. To be specific, the following permissions will be set by mkiocccentry and verified by txzchk:

- executable files MUST be either try.sh or try.alt.sh and MUST be in the top directory and MUST be -r-xr-xr-x. If any other executable file is found it is an error.
- directories MUST be drwxr-xr-x. If any other permissions for directories are found it is an error.
- other files MUST be -r--r--r--. This includes the three mandatory files AND .info.json AND .auth.json. If any other permissions for files are set it is an error.

The test suite still checks for depths that are too deep because the collect_topdir_files() scans the entire tree. The only difference now is that no extra files are copied. As the check_extra_data_files() function would become obsolete when this issue is resolved it is not a big deal that the function is now gone, especially as the way it was called was not how it will be done (the workdir is the source of all files not just the required files). Because of this change the extra_count is set to 0. However the test files still can be verified and the test txzchk files are not affected by this fact. Thus the only change here is that extra data files are not currently copied to a submission directory and won't be until the rest of the topdir is processed.

Updated MKIOCCCENTRY_VERSION to "1.2.9 2025-01-30". Updated TXZCHK_VERSION to "1.1.8 2025-01-30".


xexyl:
Oh great.

```
The  being operated upon was previously , but the underlying file may have been changed since then.
```

Of course it was ... since it was written to and then closed. I'm not even sure how to fix this complaint. Maybe a fsync(2)? I'll try that ... though not sure if it'll work for GitHub.

xexyl:
.... actually not sure of that. I'll try another way.

xexyl:
If this does not work I'll have to think about it and maybe come back later. Perhaps it's okay to run the function on an open file but I'm not sure if it is (I guessed not but I'm not sure what GitHub is concerned about - at least not entirely).

Unless it means maybe the file could have been moved? If this fails I'll look at that.

xexyl:
Now an issue with picky. Let's see if everything is resolved now.

xexyl:
I am not sure right now how to fix this. Any idea @lcn2 ? I have other things I have to do now ..

It's amusing that external tools being run on it is not a problem but a syscall is. I'll have to look into when this can be used and when not, maybe doing a test.

xexyl:
One thing I just thought of is that if I set the read-only mode before writing to the file that could cause another issue. One solution would be to not require the .auth.json and .info.json to be read-only but is that what you want? Maybe you have another idea?

xexyl:
I have an idea. Will test it out ..

xexyl:
Trying `fchmod(2)`. If this does not work I'll have to come back later.

xexyl:
There we go. Hopefully the bits are okay in `open(2)`. If you want a different set please let me know and I'll be happy to update it but it seems at first glance okay.

xexyl:
Oh I forgot to close the file descriptor ... will do that now.

lcn2 (approved):
Great progress, @xexyl

xexyl:
> Great progress, @xexyl

Thank you. Just saw in time that you merged so opened another pull request to use `close(2)` on the descriptors.

That's all I'll do today though. Probably very little if anything tomorrow but Saturday should be good.

PR #1111, opened 2025-01-31
Updated mkiocccentry.1 for newer syntax

xexyl:
Other fixes (typos and formatting) were also applied but it is quite likely that this update is incomplete or has additional problems. These will be looked at (and if fixed if necessary) later.


lcn2 (approved):
👍

PR #1112, opened 2025-01-31
Fix txzchk to use macros for try scripts

xexyl:
Fix txzchk to use macro names for try.sh and try.alt.sh. Also add new
count variables for files with invalid permissions and number of executable
files. Added a good test file and a bad test file. Rebuilt error files.

Updated TXZCHK_VERSION to "1.1.9 2025-01-31".


xexyl:
Very possibly all I can do today but hopefully I can reply to some comments from yesterday. Best wishes either way.

lcn2 (approved):
👍

PR #1123, opened 2025-02-03
Improve copyfile() from jparse/

xexyl:
Improve function copyfile() so that it can either copy the mode from the 
source file to the destination file (mode as in stat(2)'s st_mode) OR
set to a specific mode. This necessitated updating mkiocccentry(1).

Updated MKIOCCCENTRY_VERSION to "1.2.13 2025-02-03.


xexyl:
Hopefully I did not miss anything and checks are okay but if not I'll fix. I'll have to go for a while but when I am back I hope to (not guaranteed) to work on mkiocccentry more. But this had to be done in case one wanted to set mode `0000` by some chance (though allowing for setting mode might make it not a true copy in some cases but it's useful in a lot of cases).

PR #1131, opened 2025-02-08
Make some fixes for issue #1070

xexyl:
Add missing lists to present to user (before copying files) such as: directories
to be made; unsafe files and directories to be ignored (instead of making it an
error - it is an error only if in the submission directory). When traversing the
submission directory create lists for directories so these can be presented to
the user as well (if any that are in the topdir are not found in the submission
directory).

Use macros instead of raw octal modes (for mkdir(2)/mkdirs()/copyfile()).

Various other fixes might also have been made.

Updated MKIOCCCENTRY_VERSION to "1.2.18 2025-02-08".
Updated SOUP_VERSION to "1.1.17 2025-02-08".


xexyl:
I am thinking about modularising the two functions some by adding more functions .. they're quite long and it's hard to follow some of it (or harder).

But I know documentation also has to be updated (some of which I have started) so there's that too and since this works ...

Still this commit shows that I had missed some things (some of which might not have been thought of before) so it might be good to modularise it for that reason too.

xexyl:
I am modularising it now though I will have to go afk for a bit soon.

xexyl:
Code is good I think but have to fix comments .. back in a while. At that point I'll finish the comments and commit. After that I think I'll be able to update the documentation (that you mentioned). Hopefully I can get the man page updated today too but that's very possibly not going to happen today.

PR #1133, opened 2025-02-09
Add maximum directories for submissions

xexyl:
Now one cannot abuse the fact that there was no limit on the number of
directories besides the depth of subdirectories.

For txzchk(1) (due to how it works) it counts the submission directory
itself (which is why it is MAX_DIR_COUNT and not MAX_EXTRA_DIR_COUNT)
but for mkiocccentry(1) it does not count the topdir/submission
directory as that would be confusing to users (it would seem like there
is a bug as well whereas now it only would be confusing if they look at
the macro - but since that's unlikely it is less of a problem and it's
less confusing than in txzchk(1) as in the tarball you do see the
submission directory itself). This update does require a chance in
documentation.

The number chosen was arbitrarily selected and certainly could be modified by
the judges if desired but I chose 13: not only did it seem like a reasonable
choice but it's a prime and also is a swap of the digits in the max extra file
count (which also happens to be prime).

Updated MKIOCCCENTRY_VERSION to "1.2.19 2025-02-09".
Updated TXZCHK_VERSION to "1.1.12 2025-02-09".
Updated SOUP_VERSION to "1.1.18 2025-02-09".


xexyl:
The man page and the website files need to be updated now but I am not sure if I can do that today. I hope to work on it tomorrow and hopefully also get the other needed changes in (or start working on them).

xexyl:
I have just done some clean up ... found some dead code (I wondered about it and suspected it) which was causing confusion for documentation. I'll be committing to another pull request soon. Once that is done it should make it a bit easier to document things (though not sure if I'll have time to do that today).

PR #1135, opened 2025-02-10
Change MAX_DIR_COUNT to MAX_EXTRA_DIR_COUNT

xexyl:
To make it much less confusing the submission directory is no longer considered
with the maximum number of directories allowed. This required an algorithm
change in txzchk(1) as well as some updates to how mkiocccentry(1) checks
if there are too many extra directories.

As for txzchk(1) the struct txz_file now has a char *top_dirname (which is
calculated like dir_name(path, -1);). In the case fnamchk(1) did not fail
(i.e. dirname - renamed from dir_name is not NULL) and we found a
directory (first char of the line is d) we get the directory name with the
trailing / removed (dir_name(path, 0);) and compare it with the top
directory name; if they're not the same and the dirname (the submission
directory name) is the same as the top directory name then it's an extra
directory.

Added function free_txz_file() which takes a pointer to a pointer to a struct
txz_file and is called by free_txz_files_list().

Updated MKIOCCCENTRY_VERSION to "1.2.20 2025-02-10".
Updated TXZCHK_VERSION to "1.1.12 2025-02-09".
Updated SOUP_VERSION to "1.1.19 2025-02-10".


xexyl:
Oh dear it looks like I need to update jparse. I did test it but not well enough it seems. Not sure how though as I actually ran mkiocccentry directly. 

Unable to strdup() some strings which is really strange. I might have to put in more debug output to see why first. 

I have to be afk a bit and then I will fix it so that this enhancement will work.

xexyl:
Hmm maybe not strdup() .. I see that only in the app. I see here on the laptop:

```
test_ioccc/txzchk_test.sh: Warning: in run_test: expected errors: ./test_ioccc/test_txzchk/bad/submit.test-6.1922598667.txt.err do not match result of test: .txzchk_test.stderr.CxozLFcH0Z
```

but then:

```sh
$ ./test_ioccc/txzchk_test.sh -v 3
./test_ioccc/txzchk_test.sh: debug[3]: assume TOPDIR is .: /Users/cody/code/xexyl-mkiocccentry
./test_ioccc/txzchk_test.sh: debug[3]: TOPDIR is the current directory: /Users/cody/code/xexyl-mkiocccentry
./test_ioccc/txzchk_test.sh: debug[3]: about to run txzchk tests that must pass: text files
./test_ioccc/txzchk_test.sh: debug[3]: about to run txzchk tests that must fail: text files
./test_ioccc/txzchk_test.sh: debug[1]: all tests PASSED
```

??

xexyl:
Oh I wonder if it's a linux versus macOS output thing .. will check in linux.

xexyl:
Yup .. linux issue. Great.

xexyl:
Next step is to update the FAQ and the man page .. hopefully guidelines and rules too and of course the submit server screenshots have to be updated.

But afk a while.

xexyl:
I've found some other issues to be resolved .. working on those now. Some `fts_info` types I missed.

PR #1144, opened 2025-02-11
NetBSD fixes

xexyl:
Fix test_ioccc/ioccc_test.sh to have a -m make option for those who need to
specify a path to a GNU make(1) tool. One can use this with make test by way
of:

    make MAKE=/path/to/gnumake test

for example.

Fix test_ioccc/gen_test_JSON.sh to work for NetBSD (change in grep regexp).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,18 @@ requires more work).
 
 Fix `bug_report.sh` to check `test_ioccc/gen_test_JSON.sh`.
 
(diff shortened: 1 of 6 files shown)


xexyl:
Oh great. If `MAKE` is just `make` it fails. I have to check how it was done previously (I thought I had).

xexyl:
Should be fixed now. Checking ...

PR #1145, opened 2025-02-11
Add hints to copy_topdir()

xexyl:
Add hints to mkiocccentry when reporting files/directories/symlinks that will
not be added/made. If need_hints is false this is not done. This should help
users out so that they do not have to go to the FAQ just to find out what is
wrong.

Updated MKIOCCCENTRY_VERSION to "1.2.21 2025-02-11".


lcn2 (commented):
👍

PR #1146, opened 2025-02-11
Fix ioccc_test.sh for -m make

xexyl:
We cannot test that the file is a regular executable file because there
is no guarantee there is a path supplied and doing test -x make will
fail in most cases. Since make was used previously without a check the
removal of the check should be fine.


lcn2 (approved):
👍

PR #1147, opened 2025-02-11
Fix mk_submission_dir() to use mkdir() and chmod()

xexyl:
Change mk_submission_dir() to use mkdir(2) with mode 0 and then use
chmod(2) with the correct modes. This should fix the issue where the directory
made has the wrong mode due to umasks like what happened with NetBSD. (This is
also how the mkdirs() function from jparse works.)


lcn2 (approved):
👍

PR #1153, opened 2025-02-12
Add to list of ignored directories for other RCSs

xexyl:
When printing lists of ignored directories/files (not the ones found in the
topdir but those that we do not allow like ignored directories, forbidden
filenames etc.) it now prints them in tabular form. The same could be done for
what is in topdir (and later on submission directory) but not doing this makes
it easier to verify everything is okay.

Updated SOUP_VERSION to "1.1.20 2025-02-12".


lcn2 (approved):
👍

PR #1155, opened 2025-02-12
Add even more sanity to check_submission()

xexyl:
These checks make sure that files (depending on name) and directories 
are the correct permissions. This is done by new jparse/ util functions
(synced from the [jparse repo](https://github.com/xexyl/jparse/)). Now
not only will txzchk(1) detect invalid permissions but so will
mkiocccentry(1) (and a bug fix in the mkdirs() function was also made
but not documented by mistake).

mkiocccentry_test.sh now tests the optional files as well (that is
prog.alt.c, try.sh and try.alt.sh).

Updated MKIOCCCENTRY_TEST_VERSION to "1.0.11 2025-02-12".


lcn2 (approved):
👍

PR #1157, opened 2025-02-13
New util functions for #1070 and #1152

xexyl:
Sync jparse repo (https://github.com/xexyl/jparse/) to jparse/ for new utility functions (FTS related as well as at least one other). These FTS related ones are useful for both #1070 (which it now uses) and also the new issue #1152.  Most importantly (at this time) is that we no longer use the fts_open() and fts_read() functions directly: instead it is read_fts() which is a wrapper to it but one can call it multiple times to get the next entry in the tree.  Using this has multiple advantages which will be more fully realised once issue #1152 is being worked on (that includes possibly the new function find_file(), though theoretically #1070 could also use it - it just does not need to). The function find_file() returns the path from the directory searched so one can fopen(3) it if they wish.

As a special feature of read_fts(), one can switch back to the original directory, depending on the args (all NULL / -1 except for the int *cwd which must be not NULL and *cwd must be >= 0); this is useful because the function, if dir != NULL (or dirfd is a valid directory FD) will change the directory. This can be used after find_file() as well as long as you make sure that cwd is not NULL and *cwd is a valid FD (the feature will call close(*cwd) so it is no longer valid until it is obtained again perhaps by another call to the function).

Added /test_ioccc/topdir to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,7 @@
 /test_ioccc/test_JSON/
 /test_ioccc/test_ioccc.log
 /test_ioccc/test_iocccsize/
-/test_ioccc/test_src/
+/test_ioccc/topdir/
 /test_ioccc/workdir/
 /test_ioccc/txzchk_test.log
 /test_ioccc/utf8_test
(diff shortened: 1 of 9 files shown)


lcn2 (approved):
👍

PR #1158, opened 2025-02-13
Start work on issue #1152

xexyl:
Added ignored filenames list option (-i foo).

Only creates the array, adds strings (if not added yet) and at the end 
frees the array. This does not interfere with the current way 
chkentry(1) works.

Man page not updated yet as more work has to be done.


lcn2 (approved):
👍

Issue #1159, opened 2025-02-14
Bug: The answer file produced by mkiocccentry -A does not work for mkiocccentry -i


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

We used performed:

```sh
/bin/rm -rf workdir/*
mkiocccentry -A answer/answer.0 workdir good-dir/12345678-1234-4321-abcd-1234567890ab-0
```

to produce the following `answer/answer.0` content:

```
MKIOCCCENTRY_ANSWERS_IOCCC28-1.0
12345678-1234-4321-abcd-1234567890ab
0
title.0
abstract for slot 0
1
first0 middle0 last0
AU
user0@example.au
http://host0.example.au/index.html
https://alt0.example.au/index.html
@user0@example.au
@user0
affiliation for user 0
n
first0_middle0_last0
ANSWERS_EOF
```

However when we did:

```sh
/bin/rm -rf workdir/*
mkiocccentry -i answer/answer.0 workdir good-dir/12345678-1234-4321-abcd-1234567890ab-0
```

This produced the following error:

```
Welcome to mkiocccentry version: 1.2.22 2025-02-12

Performing sanity checks on your environment ...
... environment looks OK


The format of the non-test IOCCC contest ID appears to be valid.

The following is a list of unsafe filenames that will be ignored:

.auth.json
.info.json
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
Please enter either y (yes) or n (no)
ERROR[107]: copy_topdir: aborting because user said unsafe filenames list is not OK
zsh: exit 107   mkiocccentry -i answer/answer.0 workdir
```

### What you expect

An answer file produced by `-A foo` will, under the same content, work for `-i foo`.

### Environment

- OS: macOS Sonoma 14.5
- Device: MacBook Pro M1
- Compiler:  Apple clang version 15.0.0 (clang-1500.3.9.4)

mkiocccentry -V produced:

```
mkiocccentry version: 1.2.22 2025-02-12
jparse utils version: 1.0.7 2025-02-12
[...]

PR #1160, opened 2025-02-14
Fix issue #1159 - -i answers always answers yes

xexyl:
This is absolutely ESSENTIAL so that we don't have to worry about changes in file sets that would cause problems; if any file set (that we have to confirm is OK) changed the answers file would be ENTIRELY useless because the order of the answers would be different (i.e. we would need an additional line!). To be helpful if answer_yes == true (i.e. -y is used or implied) and !quiet) (i.e. -q not use) we will warn the user that we will answer yes to every question (at the beginning right after the Welcome to mkiocccentry message). We use the print macro from jparse/ rather than msg() from dbg or printf with a single check because it performs many more checks which is very important when always answering yes (perhaps the function pr() that print uses should allow a string without having to have a format but that's another matter entirely).

Improved the handling of user saying 'no' to a question: namely it suggests the user fix their topdir (giving the absolute path we determined earlier) and remove the submission directory (which is under the workdir) also giving the absolute path determined earlier on.

Note that mkiocccentry_test.sh already uses -q so we don't need to update it (since it's not necessary to show this stuff for tests). The reason this issue was not detected earlier is because the script uses -y. Even so the code copy_topdir() and also check_submission() now do if (!answer_yes && !read_answers_flag_used) instead of just if (!answer_yes) (not strictly necessary but this adds another layer of defence in case something goes wrong).

A message was incorrect too (referred to the wrong type of file).

Additionally an unnecessary arg to get_contest_id() was removed (it was a pointer to read_answers_flag_used but that is a global variable so it's unnecessary to have it in the function as we can just set read/set its value as necessary.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,42 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.33 2025-02-14
+
+Fix issue #1159 - -i answers always answers yes.
+
+This is absolutely **ESSENTIAL** so that we don't have to worry about changes in
+file sets that would cause problems; if any file set (that we have to confirm is
+OK) changed the answers file would be **ENTIRELY** useless because the order of
+the answers would be different (i.e. we would need an **additional** line!). To
+be helpful if `answer_yes == true` (i.e. `-y` is used or implied) and `!quiet`)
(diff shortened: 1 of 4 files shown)


lcn2 (approved):
👍

PR #1164, opened 2025-02-16
Multiple fixes and enhancements in jparse

xexyl:
Multiple fixes and enhancements in some of the jparse util functions. Updated mkiocccentry for these fixes/enhancements. Note that fts_open(3) has to have either FTS_PHYSICAL or FTS_LOGICAL so the read_fts() function (prior to calling fts_open(3)) now removes both bits and then sets the right one based on the boolean. This is why it's no longer in the options to the function call - it would be removed anyway since we can't assume what is needed (for here we want FTS_PHYSICAL but in some cases it might not be desired and there was no way to be correct otherwise).

There are other improvements and fixes as well.

Updated MKIOCCCENTRY_VERSION to "1.2.26 2025-02-16".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,22 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.35 2025-02-16
+
+Multiple fixes and enhancements in some of the jparse util functions. Updated
+mkiocccentry for these fixes/enhancements. Note that `fts_open(3)` has to
+have either `FTS_PHYSICAL` or `FTS_LOGICAL` so the `read_fts()` function
+(prior to calling `fts_open(3)`) now removes both bits and then sets the
+right one based on the boolean. This is why it's no longer in the
+options to the function call - it would be removed anyway since we can't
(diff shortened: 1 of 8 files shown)


xexyl:
Very unlikely I can do more today. Have other things I have to do and it'll almost certainly be an early night. Best wishes with the submit server tool.

lcn2 (approved):
There are some nice improvements in those PR, @xexyl, thanks!

xexyl:
> There are some nice improvements in those PR, @xexyl, thanks!

Yes. It's interesting how some of it was uncovered. It took longer than it should have but I'm glad I got it done at all after the last two nights (all good at least). I suspect that it'll be a few days and I'll be back to normal[0] but I also guess that each day I'll get a bit more done (depending on various things of course).

Anyway you're welcome. Just finishing up some other things and then will be shutting down to get some food. Probably ed in a few hours (and no more than four I'd guess).

[0] Well as normal as I might be. The word is troublesome in my case but I'm fine with it and often have fun with it too.

PR #1167, opened 2025-02-20
General clean-up

xexyl:
Removed the no longer used `vermod.sh` script (it was used to modify the versions of tools in the JSON files under `test_ioccc/test_JSON` but these files are now created at need).

Since we now recommend that one install the tools (since trying to run the tools to create a submission tarball from the mkiocccentry directory itself is more burdensome and can be problematic) the man pages (in cases where it is necessary - and that were not missed) no longer have `./` in front of commands (in the examples).

Also, for the same reason (we now recommend in most cases to install the tools), the default locations of the tools mkiocccentry (and fnamchk) uses (use) are in `/usr/local/bin`. Since it also checks the `./` it is okay. One must be careful that the right versions of the tools are used but this always went and it's in the documentation on the website too.

Clean up some usage messages to be shorter (in some ways) especially the length of lines (some formatting fixes were also made).


xexyl:
Next (hopefully today - but for now I have to break) I can look at the chkentry enhancement. I hope that will be today but we'll see.

xexyl:
Fun fact: this commit was +238 -685 . Does that mean that there were 447 lines removed (let's make it simple)? Or since it was -685 does that mean it was 447 + 685 == 1132 lines removed? We can pose the question also as: does it mean that there were actually no additions at all (the same as before but how can a removal be no additions if a commit is an addition to a file or files? Obviously that's a stretch but we could still do it!)? Or is it actually what git is claiming: there were 238 additions and 685 deletions? But still even if that's the case how can you say that that's what it is when it's + and -?

Just something fun to think about :-)

PR #1169, opened 2025-02-20
Sync jparse with new util function for #1152

xexyl:
This function finds a path in a paths array.

BTW: we do realise that a lot of the util functions in jparse have
nothing to do with JSON parsing but there is a historical reason behind
this.


lcn2 (approved):
👍

PR #1170, opened 2025-02-20
Make mkiocccentry use chkentry one time

xexyl:
Change mkiocccentry to form .auth.json and then .info.json and then run chkentry(1) on both of the files. This will changed to running it with a single arg - the topdir - but this change allows it to be easier to do once chkentry(1) has been modified to do this (part of it was done - namely that the topdir arg is there and 'set' but it's not used yet).

There was a sequenced exit code change in jparse/util.c. This will be committed 'over there' but this is not strictly a sync as it was from make seqcexit here.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,12 @@ paths array.
 BTW: we do realise that a lot of the util functions in jparse have nothing to do
 with JSON parsing but there is a historical reason behind this.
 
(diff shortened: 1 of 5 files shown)


xexyl:
Hopefully nothing important was missed here but if so I'll fix (meaning a GitHub workflow issue).

PR #1172, opened 2025-02-20
Improve display of mkiocccentry_test.sh

xexyl:
It can be a bit confusing with all the different tests that it runs (running it numerous times where it must fail and times where it must succeed) so I added a '--' between invocations.

Previously it looked weird in that it would show that all warnings are ignored and then immediately give an error (because the depth is too deep). But before that warning it would show that one (because of 'test' mode) could not use the tarball as a submission. So this should help separate it a bit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -46,6 +46,13 @@ single arg - the topdir - but this change allows it to be easier to do once
 `chkentry(1)` has been modified to do this (part of it was done - namely that
 the `topdir` arg is there and 'set' but it's not used yet).
 
+More fixes in `mkiocccentry_test.sh`. It no longer shows the commands of `make
+clobber` from the `mkiocccentry` function `check_submission()`. This is done by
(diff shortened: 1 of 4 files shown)


xexyl:
What would be even better is to show the command that is about to be run. I'll do that later if I have the time, energy and remember (I guess it's not exactly a high priority issue either).

xexyl:
Actually I am aware of a much more confusing issue in the script ... and not sure if it's because of this change or not (seems unlikely). But it should be fixed if at all possible.

I can say more later. I have to go for now and maybe the day.

xexyl:
Oh I know why. It's because `make clobber` is showing the commands to be printed in the Makefile. That's unfortunate. In normal execution it is not shown but in the script it is.

So the test file could have a blank clobber rule I think. I'll see about that ...

xexyl:
Fixed ... now going afk and maybe for the day.

lcn2 (approved):
👍

PR #1176, opened 2025-02-21
Various improvements with mkiocccentry

xexyl:
Due to what will be needed with chkentry (and also in mkiocccentry - which has been updated) some of the jparse util functions were updated and new ones were added. It cannot always be assumed that finding files should be done in a case-insensitive manner so there is a new boolean in those functions (and in one or two of the new functions). The mkiocccentry had a function that is based on one of them (array_has_path()).

The arrays in the info struct in mkiocccentry now are freed and set to NULL IFF (if and only if) they are empty. This not only allows for not checking them for NULL (except before showing the contents) but it also allows for more checks that were not previously done (some were thought of in the process of this and the below change).

Update exit codes in mkiocccentry, allocating 4 for any issue (and there are a lot, even more now due to more checks (as above) and 5 for when the user says something is not okay. Updated the man page for this though the guidelines/rules/FAQ and quick start guide do need to be updated as well (for this change possibly and other more recent changes as well). This also lets the mkiocccentry_test.sh script verify that the exact error code is encountered (when testing errors) and not just not 0. Besides additional testing (after code freeze and of course before) this should, with the exception of documentation, complete the workdir topdir model, unless something else useful or important is thought of (like the allocation of exit codes and the new approach to handling the arrays).

The function free_info() now uses the recent function free_paths_array() for the dynamic arrays of paths.

Updated MKIOCCCENTRY_VERSION "1.2.30 2025-02-21".
Updated SOUP_VERSION to "1.1.22 2025-02-21"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,37 @@ Fix many debug calls in `soup/entry_util.c` at level medium to not reveal
 information about the submission. This includes GitHub handle and country code,
 amongst others.
 
+Due to what will be needed with chkentry (and also in mkiocccentry - which has
+been updated) some of the jparse util functions were updated and new ones were
+added. It cannot always be assumed that finding files should be done in a
+case-insensitive manner so there is a new boolean in those functions (and in one
+or two of the new functions). The `mkiocccentry` had a function that is based on
+one of them (`array_has_path()`).
+
+The arrays in the info struct in mkiocccentry now are freed and set to NULL IFF
+(if and only if) they are empty. This not only allows for not checking them for
+NULL (except before showing the contents) but it also allows for more checks
+that were not previously done (some were thought of in the process of this and
+the below change).
+
+Update exit codes in mkiocccentry, allocating `4` for any issue (and there are a
+lot, even more now due to more checks (as above) and `5` for when the user says
+something is not okay. Updated the man page for this though the
+guidelines/rules/FAQ and quick start guide do need to be updated as well (for
+this change possibly and other more recent changes as well). This also lets the
+`mkiocccentry_test.sh` script verify that the exact error code is encountered
+(when testing errors) and not just not 0. Besides additional testing (after code
+freeze and of course before) this should, with the exception of documentation,
+complete the `workdir topdir` model, unless something else useful or important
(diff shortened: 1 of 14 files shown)


xexyl:
And yes those two functions in particular could really be refactored. But if that happens it'll be after IOCCC28. I think if it's at all possible to do that it would be good though. They're really cumbersome. All the features are needed but it'd be nice if the arg count could be reduced somehow. But that's for another day.

xexyl:
Will be taking a break .. when back I'm going to see about working on chkentry. I do not know how much I can get done but I hope some. If not at least I got these important things done. I think the next thing to do for chkentry is to just act like no new options are specified - simply verify the files exist and that they are the right permissions (and to do these I can use the jparse util functions).

That is what I aim to do when I'm back.

lcn2 (approved):
👍

lcn2:
> Will be taking a break .. when back I'm going to see about working on chkentry. I do not know how much I can get done but I hope some. If not at least I got these important things done. I think the next thing to do for chkentry is to just act like no new options are specified - simply verify the files exist and that they are the right permissions (and to do these I can use the jparse util functions).
> 
> That is what I aim to do when I'm back.

Best wishes for a restful and well deserved break, @xexyl

xexyl:
> > Will be taking a break .. when back I'm going to see about working on chkentry. I do not know how much I can get done but I hope some. If not at least I got these important things done. I think the next thing to do for chkentry is to just act like no new options are specified - simply verify the files exist and that they are the right permissions (and to do these I can use the jparse util functions).
> 
> > 
> 
> > That is what I aim to do when I'm back.
> 
> 
> 
> Best wishes for a restful and well deserved break, @xexyl 

Thank you! Right now it's not exactly restful but that's what I want. Just taking a break from thinking much. 

I hope that I won't be too long but I am also at the mercy of having trouble focusing a lot of the time. 

Anyway what matters is that I am pretty sure I know how to proceed with chkentry. 

Adding the code to check files before worrying about what to ignore will make it easier to test and develop. 

The getting one arg only working (verifying both of the JSON files) yesterday was exactly what I needed. 

Best wishes for you too!

PR #1177, opened 2025-02-21
Improve topdir checks in chkentry

xexyl:
Better checks on topdir in chkentry (as in even though the pointers were checked for NULL they are now only done if topdir is a directory that can be searched).

Updated CHKENTRY_VERSION to "1.1.2 2025-02-21".


lcn2 (approved):
👍

PR #1178, opened 2025-02-22
Significantly simplify the FTS util functions

xexyl:
By refactoring the FTS functions in jparse/util.c the functions are far easier to use now. A great deal of thought and effort was put into this and when the topdir is passed to chk_manifest() (which is being worked on) it will be far easier. The benefit of these changes, besides being much cleaner, simpler and it being modularised, is that if some issue is discovered or some new feature is needed (as has already happened numerous times) function calls do not have to be updated.

mkiocccentry has been updated to use the simplified interface as well.

A note: jparse/ has not been fully synced because of a change in json_sem for chkentry that is not yet complete (this was done prior to deciding to rework the FTS functions). So the CHANGES.md file and the json_sem.[ch] files committed to the jparse repo are not yet here.

Updated MKIOCCCENTRY_VERSION to "1.2.31 2025-02-22"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,25 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.40 2025-02-22
+
+Significantly simplify and refactor the FTS functions in jparse/util.c. A great
+deal of thought and effort was put into this and when the `topdir` is passed to
+`chk_manifest()` (which is being worked on) it will be far easier. The benefit
+of these changes, besides being much cleaner, simpler and it being modularised,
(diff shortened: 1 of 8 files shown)


lcn2 (approved):
This looks like really nice work
🙏 @xexyl

xexyl:
> This looks like really nice work 🙏 @xexyl

Thank you! It took a lot of effort and time - more so with waking up so early. But it is definitely worth it. Good thing I found that subtle bug about absolute paths too. Well 'find' is not the right word - I realised it could happen.

Best wishes! I'll do more tomorrow. I am afraid I am spent for the day.

PR #1185, opened 2025-02-25
Improve mkiocccentry_test.sh

xexyl:
Add only required files test.

Updated MKIOCCCENTRY_TEST_VERSION to "1.0.14 2025-02-25".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 # Major changes to the IOCCC entry toolkit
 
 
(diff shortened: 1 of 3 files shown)


xexyl:
Okay now that this is fixed I can work on chkentry ... but not now. I'm going to take a break for a bit. When I'm back and more alert (I hope - well I'm alert but I mean more alert than I am) I will look at chkentry and fix the loophole I thought of yesterday.

PR #1186, opened 2025-02-25
More work on #1152

xexyl:
Plug a loophole in chkentry(1) and make it do even more tests that were missing/not previously thought of. This might or might not resolve the issue entirely but it is getting close to being resolved even if it is not yet resolved.

Updated CHKENTRY_VERSION to "1.1.5 2025-02-25".


lcn2 (approved):
good loop hole 🕳️ plug 🔌🤓

xexyl:
> good loop hole 🕳️ plug 🔌🤓

I was inspired to make this:

rare-photo-of-a-loophole

PR #1188, opened 2025-02-26
Rename topdir -> submission_dir in chkentry

xexyl:
The tool is run on the submission directory that has the files that came  
from the topdir, not the topdir itself.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.44 2025-02-26
+
+Remove unused arg in `write_json_files()`.
+
+Updated `MKIOCCCENTRY_VERSION` to `"1.2.34 2025-02-26"`.
+
+
 ## Release 2.3.43 2025-02-25
 
 Improve `mkiocccentry_test.sh` - add only required files test.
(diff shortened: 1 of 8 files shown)


lcn2 (approved):
Seems reasonable 👍

xexyl:
> Seems reasonable 👍

I thought so too. It is amusing it only dawned on me today but at least it's fixed now. It could be confusing to some otherwise.

lcn2 (approved):
Good idea 👍 

Does anything in "the other repo", such that the FAQ, need to change in a similar way, @xexyl ?

xexyl:
> Good idea 👍
> 
> Does anything in "the other repo", such that the FAQ, need to change in a similar way, @xexyl ?

I was actually working on that when I thought I noticed something missing (a check) which made me see this. So in answer to the question - yes. Working on it and in general going through stuff about chkentry for the recent changes.

I might also wish to do so for mkiocccentry though I did do some of that a while back. Still I might have made further updates (I know I did in the lower level stuff but not sure on the higher level).

xexyl:
Just synced the updated html files to the server .. will check them and if all good I'll commit. I'm unsure if I can do a thorough check today but if not I will tomorrow. I'm having a hard time seeing now and I'll probably be leaving soon as well.

PR #1189, opened 2025-02-27
Some [of the] final changes before code freeze

xexyl:
Prepare for code freeze (28 February 2025) with some final changes, some important (and useful) and some not that important (but done to be complete).

Sync jparse repo (https://github.com/xexyl/jparse/) to jparse/ before code freeze (28 February 2025).

Updated copyright messages, some missing, some updated (based on brief discussions a while back and more recently, recent updates to code and perhaps other things) to many files.

The script bug_report.sh actually can have the script in the TOOLS variable as it will not actually cause an infinite loop as it only gets the version of the tool - which is quite useful to know that the user is using the most recent version of the script. This was also done in jparse/ but that doesn't matter here; it's only mentioned due to the sync.


xexyl:
Fun fact: currently (before this) there are 1039 pull requests closed: a prime number!

lcn2 (approved):
👍

xexyl:
> 👍

Oh sure now you ruined the prime number!!

Okay actually - thanks. I am going to try and work on the website now but I was also pondering how to update the chkentry_test.sh script.

I guess you might also have some work on that in the background ?

PR #1190, opened 2025-02-27
Improve -i answers to prompt for y/n questions

xexyl:
With -i answers used, unless -Y is used, mkiocccentry now will ask you to confirm things are okay (in other words if something goes wrong that would normally be warned about, you'll get to see it).

If you need to not have this you can use the new -Y option which forces -y. This is not usually a good idea but it's required for the test script (mkiocccentry_test.sh) as it is not interactive and shouldn't be.

Updated MKIOCCCENTRY_VERSION to "1.2.35 2025-02-27". Updated MKIOCCCENTRY_TEST_VERSION to "1.0.15 2025-02-26".


lcn2 (approved):
👍

PR #1193, opened 2025-02-28
Resequence exit codes in jparse/util

xexyl:
It appears that this was not done or something went wrong when doing so
(as running make seqcexit updated the exit codes and this comes from
after running it in jparse/ and committing and then syncing from
jparse to jparse/).

Also as an update to the previous commit with the minor fix and
improvement to mkiocccentry -y and -Y warnings, I meant to update the
MKIOCCCENTRY_VERSION in the CHANGES.md; it is now "1.2.36 2025-02-28".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,23 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.3.45 2025-02-28
+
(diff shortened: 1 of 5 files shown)


lcn2 (approved):
👍

Issue #1207, opened 2025-03-07
Enhancement: mkiocccentry does not ignore dirrectories such as SCCS


SirWumpus:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

In the list of ignored directories related to version control, `SCCS` has been overlooked.

```
We ignore certain directory names, in particular:

              .bzr       .circleci       .fslckout
              .git         .github         .gitlab
               .hg           .idea            .svn
         BitKeeper             CVS            RCS
          _FOSSIL_            _MTN          _darcs

This is so that one may use an RCS without having to worry about
copying everything to another directory simply to use this tool.
```

### What you expect

Ignore `SCCS` directories.  Also as I recall RCS `ci(1)` and `co(1)` used an `RCS` directory. 

### Environment

- OS:
- Device:
- Compiler:

xexyl:
Thanks for the suggestion! Good call. 

Would you please assign this to me @lcn2  ? I can take care of it tomorrow. I actually also implemented the uuid file option to make it much quicker to use. 

But I can talk about that later. 

Have to go again. 

Thanks @SirWumpus for the suggestion!

xexyl:
To help out @SirWumpus: please give a precise list of directory names. 

If you did then please just confirm (I am rushed and I also am reading this with my phone and distance glasses). Thanks!

lcn2:
If this is addressed before IOCCC28 "closes" (and that is a big *IF), then the [Important NOTE from issue 1209](https://github.com/ioccc-src/mkiocccentry/issues/1209) needs to be considered.

lcn2:
We have relabeled this a feature.  The stuff such as `.idea` are ignored by `mkiocccentry(1)` by default.

Things such as `SCCS` (that is probably what `RCCS` means in the top level comment) or `CVS` less likely to be used and so les likely to be a problem.  Even if, say `CVS` or `BitKeeper` were found **ANY** the a rule violation were not triggered (say because there wasn't enough "extra" files to go over the limit), then the IOCCC judges can simply ignore such content.

xexyl:
I'll quickly add this to the soup file ... thanks @SirWumpus!

xexyl:
I added both directory names to this btw. So when I do the commit this will be resolved.

SirWumpus:
> I added both directory names to this btw. So when I do the commit this will be resolved.
These two right?
```
SCCS
RCS
```

[...]

Issue #1210, opened 2025-03-07
Enhancement: Directory ignore list confusing


SirWumpus:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

When you reach the point where you are prompted with a list of directories to ignore you are asked if this list is ok or not (Y/N).  I had assumed that by saying `N` that I would be able to amend the list of ignored directories, but instead the `mkiocccentry` simply terminates.  

```
The following is a list of directories that will be ignored:

.git

Is this OK? [Yn]: n
```

Post IOCCC28, consider revisiting the UI and/or explanatory text to explain what happens when No is answered and if there is a prompt to amend the ignore list further down.

lcn2:
If this is addressed before IOCCC28 "closes" (and that is a big *IF), then the [Important NOTE from issue 1209](https://github.com/ioccc-src/mkiocccentry/issues/1209) needs to be considered.

xexyl:
The only thing that is done: it aborts. I can add to the messages that it will just abort if that's helpful. Let me see if I can take care of that simply ...

xexyl:
As for amending the ignore list: how do you mean? What we ignore we ignore and there's no way to ignore wha we ignore. :-)

SirWumpus:
> The only thing that is done: it aborts. I can add to the messages that it will just abort if that's helpful. Let me see if I can take care of that simply ...

I got trip up by this at every prompt that gives a Y/n request that says "Is this list OK? " instead of a more accurate `Should we continue?` or similar phrasing to continue.  What is the point of asking the question if the only choice is to abort.

> As for amending the ignore list: how do you mean? What we ignore we ignore and there's no way to ignore wha we ignore. :-)

There are directories NOT named `.something`, eg `reference/`, or other development files like `*.tws` (TextPad workspace) which appear in `.gitignore` that I don't want to include.  So it would be nice to be able to amend the list of things it should not include, rather than assume everything that isn't hidden is fair game.

This places a burden on the developer to shuffle / rename stuff temporarily just to avoid including it.  Maybe having the developer create their own file-list-manifest file (similar to the `tar(1)` -I/-T option), that is then validated.

Anyway, that would have to be a post-IOCCC28 thing to do cause.  So far I've not been able to submit an entry cause every prompt that I answer `N` to, simple aborts, without telling me that is what will happens.  Again a better prompt like `Do you wish to continue?` would imply that a negative would abort.

[...]

Issue #1209, opened 2025-03-07
Bug: potential mkiocccentry nit


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

Some people place the `workdir` under `topdir`.

We found an un-tarred tree with `topdir/workdir/UUID/` where UUID was the username.

The `topdir/workdir/UUID/` directory is empty.

The `topdir/workdir/' directory only has the `UUID` directory under it.

### What you expect

The `mkiocccentry(1)` tool would disallow a `workdir` anwyere under `topdir`.

### Environment

- OS: unknown
- Device: unknown
- Compiler: unknown

FYI, the `.info.json` showed:

```json
    ...
    "IOCCC_info_version" : "1.16 2024-05-18",
    ...
    "mkiocccentry_version" : "2.0.0 2025-02-28",
    "iocccsize_version" : "28.15 2024-06-27",
    "chkentry_version" : "2.0.0 2025-02-28",
    "fnamchk_version" : "2.0.0 2025-02-28",
    "txzchk_version" : "2.0.0 2025-02-28",
```

### bug_report.sh output

n/a

### Anything else?

When we run `chkentry(1)` on the un-tarred tree we see these errors:

```
...
.auth.json validation error ERROR[149]: chk_mkiocccentry_version: invalid mkiocccentry_version
.auth.json validation error ERROR[120]: chk_chkentry_version: invalid chkentry_version
...
.info.json validation error ERROR[149]: chk_mkiocccentry_version: invalid mkiocccentry_version
.info.json validation error ERROR[120]: chk_chkentry_version: invalid chkentry_version
.info.json validation error ERROR[172]: chk_txzchk_version: invalid txzchk_version
...
ERROR[1]: main: one or more tests failed
```

So it is also possible that someone has an outdated mkiocccentry toolkit.

It is also worth noting that the `workdir` did **NOT** go into the `.info.json` manifest because where were no files under the `workdir` tree that was put under `topdior`.

# Important NOTE

**IF** this results in a change to `mkiocccentry(1)` (and that is a big **IF**), resulting in version changes to **ANY** if the version values found in `.info.json`, then this **WILL COMPLICATE** the version checking.

Consider, just as an example, that for some reason the `mkiocccentry_version` changes from "2.0.1 2025-03-02" to "2.0.2 2025-03-07".  Remember, this is just as a hypothetical example!

[...]

PR #1211, opened 2025-03-07
Fix critical issues in mkiocccentry

xexyl:
It is an error if the topdir and the workdir are the same (same device and inode).

If the workdir is found in the topdir it is skipped.

Before doing anything make sure the args are actually directories that can be searched and (in the case of workdir) written to.

To help users not have to repeatedly copy and paste their UUID there is a new option, -u uuid which will read from a file the UUID. If the file does not exist or cannot be read or it does not have a valid UUID the program will prompt for it as if you didn't use the option.

Extra sanity checks in the FTS code.

Make it clearer what happens when one does not agree to a prompt in the scanning/copying the topdir and the checking of the submission directory. The prompt also is now 'Do you wish to continue?' instead of 'Is this OK?'.

When checking the submission directory if a directory that is the workdir OR topdir is encountered it is an error.

Ignore directories for additional RCSs: namely RCS and SCCS.

I believe this should resolve issues #1207, #1209 and #1210.

IMPORTANT REMINDER: the version of the tools have NOT been updated because doing so would INVALIDATE submissions already uploaded!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,40 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.3 2025-03-07
+
+Fix critical issues in mkiocccentry.
+
+It is an error if the topdir and the workdir are the same (same device and
+inode).
+
+If the workdir is found in the topdir it is skipped.
+
+Before doing anything make sure the args are actually directories that can be
+searched and (in the case of workdir) written to.
+
+To help users not have to repeatedly copy and paste their UUID there is a new
+option, `-u uuid` which will read from a file the UUID. If the file does not
+exist or cannot be read or it does not have a valid UUID the program will prompt
+for it as if you didn't use the option.
+
+Extra sanity checks in the FTS code.
+
+Make it clearer what happens when one does not agree to a prompt in the
+scanning/copying the topdir and the checking of the submission directory. The
+prompt also is now 'Do you wish to continue?' instead of 'Is this OK?'.
+
+When checking the submission directory if a directory that is the workdir OR
+topdir is encountered it is an error.
+
+Ignore directories for additional RCSs: namely `RCS` and `SCCS`.
+
+I believe this should resolve issues #1207, #1209 and #1210.
+
+**IMPORTANT REMINDER**: the version of the tools have NOT been updated because
+doing so would **INVALIDATE** submissions already uploaded!
+
+
 ## Release 2.4.2 2025-03-02
 
 Resolve issue #1201.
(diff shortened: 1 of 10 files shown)


lcn2 (changes requested):
Out of an abundance of caution, and in light of issue #1215, we request that this PR be withdrawn until issue #1215 is resolved and well tested as described in that issue.

We realize that this PR **seems** to be safe, however, we do not want to find ourselves in a case where a change impacts already existing situations.

We also need to consider that some of these changes are **NOT** critical to IOCCC28.

For example issue #1207 is not a critical fix.

And while `workdir` under `topdir` might be allowing, issue #1209  is not an "show stopper" (as the expression goes) for IOCCC28.

And while it is confusing, it could be argued that issue #1210 is also not an "show stopper" (as the expression goes) for IOCCC28.

Given that the text for this PR implies that the above mentioned non-critical issues part of this PR, we request that this PR be withdrawn until  issue #1215 is resolved and well tested as described in that issue.

xexyl:
I'm actually working on that issue right now.

And yeah the log is kind of not entirely true. I'll have that issue sorted out soonish.

xexyl:
> # IMPORTANT REQUEST
> Please close this PR as any subsequent PR you issue, such as resoling issue #1215 will be impacted and complicated by this open PR.

Well that's problematic as I'm not sure how to save the changes. I'll work that out first and then close it.

xexyl:
Oh great. This is causing me to be in a detached head state.

xexyl:
Will close and add the other stuff first - once I figure out how to fix the script that generates JSON files ....

lcn2:
# QUESTION

Is it possible to split up the this PR, post #1215 into separate PR.  If that is too difficult that is OK, but for the sake of those who might wish to address a post #1215 resolution PR issue, it might be better to separate ... **IF PRACTICAL** addressing issues into separate PRs if that is reasonable, @xexyl 

Thank you for withdrawing this PR @xexyl and sorry (tm Canada 🇨🇦) for the complcaiotn.

But we need to be **EXTREMELY CAREFUL** of modifications while the contest is open.

For this reason, we recommend that you discuss in the given issue, if resolving a given issue while IOCCC28 is useful, important, and needed during IOCCC28.  Then if we think it the case, then issue a PR accordingly .. wait for the PR to be processed before considering another PR as GitHub PRs tend to stack on top of each other.

There may be a case to put in a useful mod to the tool set while IOCCC28 is open, particularly if issue #1215 allows us to preserve existing submissions as valid.  We just have to be careful.

Issue #1221, opened 2025-03-08
Bug: `mkiocccentry` fails if the `Makefile` contains `:=` assignments.


SirWumpus:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

When the `Makefile` contains `:=` assignments for example:

```
CC	:= cc
CSILENCE:=
CFLAGS	:= -std=gnu17 -Os -Wall -Wextra -pedantic ${CSILENCE} ${DBG}
CPPFLAGS:= -D_XOPEN_SOURCE=700
```

Then `mkiocccentry` fails with the following...

```
Checking Makefile ...

At least one problem was detected with the Makefile provided:

The all rule appears to not be the first (default) rule.

Makefiles must have the following Makefile rules:

    all - compile the submission, must be the first submission
    clean - remove intermediate compilation files
    clobber - clean, remove compiled submission, restore to the original submission state
    try - invoke the submission at least once

While this program's parser may have missed finding those Makefile rules,
chances are this file is not a proper Makefile under the IOCCC rules.

We LIKE submissions that use and we recommend and encourage you to use the
example Makefile found here:

        https://www.ioccc.org/next/Makefile.example
```

### What you expect

`gmake` and BSD `make` both support `:=` macro assignments and should not cause the Makefile check to fail.

### Environment

- OS: Windows 10 Home using NetBSD VM
- Device:
- Compiler:

### Anything else?

The Makefile check allow SUS `make(1)` pseudo targets:

```
.POSIX :
.SUFFIXES :
.SUFFIXES : .c .h .i $O $E
```

Also inference rules also happily pass through...
```
.c.i:
	${CC} ${CFLAGS} ${CPPFLAGS} -E $*.c >$*.i

.c$E:
	${CC} ${CFLAGS} ${CPPFLAGS} -o $*$E $< ${LIBS}
```

Also the BSD Make `?=` and `!=` assignments is also supported by `gmake` and passes through without issue.

xexyl:
> ### Is there an existing issue for this?
> 
> * [x]  I have searched for existing issues and did not find anything like this
> 
> ### Describe the bug
> 
> When the `Makefile` contains `:=` assignments for example:
> 
> ```
> CC	:= cc
> CSILENCE:=
> CFLAGS	:= -std=gnu17 -Os -Wall -Wextra -pedantic ${CSILENCE} ${DBG}
[...]

Issue #1206, opened 2025-03-04
Enhancement: add a new option to mkiocccentry to read from a file the UUID for easier updating and creating new tarballs


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

Similar to an rc file except that it would simply read in the UUID itself. So if a person's UUID (and obviously this is bogus) was:

```
foo.bar.baz
```

they could put in a text file that id.

It probably should have a tag (header and footer) to indicate it's correct but I don't know about that - it might be annoying: I can see it both ways.

This would be easy to implement and I could adapt other code I wrote for this.

As for the option letter I don't know yet - perhaps `-u` or `-U`. I would think it be a big mistake to have a default path though maybe some would disagree with that.

### Anything else?

It is not clear to me how many would find this useful but I know I would. This is post iOCCC28.

Please assign to me if you wish and after IOCCC28 I'll look at it - along with the other enhancements needed. This should be a low priority issue I would guess.

We'll also see how IOCCC28 goes - maybe this isn't really useful though it still seems convenient and like the answers file I initially added.

xexyl:
I actually had the idea of quickly writing this. But if I do it probably would not be committed until after IOCCC28. Still might be useful to me but I guess I will see. 

Wouldn't commit unless someone finds a bug that needs to be fixed. But hopefully that doesn't happen: it's been tested a lot. 

But if I were to do this it wouldn't take much time. 

Well anyway after the next contest I will add this feature. Perhaps not everyone would use it but I think it's a worthwhile addition. 

Hope you're sleeping well!

I have been awake since half past 1 so I am quite tired although I am quite awake. 

Best wishes and let's hope everything goes smoothly!

lcn2:
We believe we have addressed all of the current questions that still need answering at this time. If we've missed something or something else needs to be clarified, please ask again.

lcn2:
We believe we have addressed all of the current questions that still need answering at this time. If we've missed something or something else needs to be clarified, please ask again.

xexyl:
> BTW: This commit is pending the successful resolution and extended testing of the resolution of issue [#1215](https://github.com/ioccc-src/mkiocccentry/issues/1215).

Working on it.

lcn2:
Perhaps for post-IOCCC28,  you could add a `-U UUID` command line option which tell `mkiocccentry(1)` what the UUID username should be?

[...]

PR #1223, opened 2025-03-09
Resolve some (mostly top priority) issues

xexyl:
Resolve issue #1206. Some people wanted this option to submit multiple submissions without having to repeatedly copy/paste the UUID. Now they can just put the UUID in a text file and use -u uuid. If the file is not a regular readable file or it does not have a valid UUID it'll prompt like before. If the -i answers flag is used it is not relevant.

Resolve issue #1210. The directory/file lists to be ignored prompting was confusing to some people. The question of 'Is this OK?' was reworded to 'Do you wish to continue?' and the explanation is hopefully a bit clearer too.

Resolve issue #1221. Removed the check for first rule is all (in Makefiles).  This allows one to also have earlier on the format of 'CC:= cc' rather than just 'CC= cc', should they wish. The .info.json file still has this bool but it's always set to true and chkentry will ignore it. AFTER IOCCC28 it will be removed from .info.json and chkentry code will no longer have the functions involved.

Resolve issue #1209. Although not labelled 'top priority' it was already done in the code. If topdir is the same as workdir it is an error. If workdir is under topdir it is an error. If topdir is somehow slipped into the submission directory it is an error. If workdir is encountered in topdir we skip it with fts_set() so as to not descend into it. Also, we now check that the directories are directories an are the right permissions, prior to even trying to scan/copy files/directories (to show a better error message).

IMPORTANT NOTE: none of these will cause a previously uploaded submission to be invalidated. You do NOT need to install or use the updated tools. These are for those who want or need (or feel they need) the features (and fixes) only.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,41 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.4 2025-03-09
+
+Resolve some (mostly top priority) issues.
+
+Resolve issue #1206. Some people wanted this option to submit multiple
+submissions without having to repeatedly copy/paste the UUID. Now they can just
+put the UUID in a text file and use `-u uuid`. If the file is not a regular
+readable file or it does not have a valid UUID it'll prompt like before. If the
+`-i answers` flag is used it is not relevant.
+
+Resolve issue #1210. The directory/file lists to be ignored prompting was
+confusing to some people. The question of 'Is this OK?' was reworded to 'Do you
+wish to continue?' and the explanation is hopefully a bit clearer too.
+
+Resolve issue #1221. Removed the check for first rule is all (in Makefiles).
+This allows one to also have earlier on the format of `CC:= cc` rather than just
+`CC= cc`, should they wish. The .info.json file still has this bool but it's
+always set to true and chkentry will ignore it. **AFTER** IOCCC28 it will be
+removed from .info.json and chkentry code will no longer have the functions
+involved.
+
+Resolve issue #1209. Although not labelled 'top priority' it was already done in
+the code. If topdir is the same as workdir it is an error. If workdir is under
+topdir it is an error. If topdir is somehow slipped into the submission
+directory it is an error. If workdir is encountered in topdir we skip it with
+`fts_set()` so as to not descend into it. Also, we now check that the
+directories are directories an are the right permissions, prior to even trying
+to scan/copy files/directories (to show a better error message).
+
+**IMPORTANT NOTE**: none of these will cause a previously uploaded submission to
+be invalidated. You do **NOT** need to install or use the updated tools. These
+are for those who want or need (or feel they need) the features (and fixes)
+only.
+
+
 ## Release 2.4.3 2025-03-07
 
 Resolve issue #1215.
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -142,6 +142,8 @@ static const char * const usage_msg3 =
     "\t-A answers\twrite answers file even if it already exists\n"
     "\t-i answers\tread answers from file previously written by -a|-A answers\n"
     "\t\t\t    NOTE: One cannot use both -a/-A answers and -i answers.\n"
+    "\t-u uuid\t\tread UUID from a file (def: prompt for UUID)\n"
+    "\t\t\t    NOTE: if an invalid UUID is in the file, it will try the usual way\n"
     "\t-s seed\t\tGenerate and use pseudo-random answers, seeding with\n"
     "\t\t\t    seed & 0x%08u (def: do not)\n"
     "\t-d\t\tAlias for -s %u\n"
@@ -190,6 +192,8 @@ static bool copying_topdir = false;    /* true ==> copying topdir and checking s
 static bool saved_answer_yes = false;   /* set to answer_yes before modifying it for scanning/copying topdir */
 static bool saved_silence_prompt = false;   /* set to silence_prompt before modifying it for scanning/copying topdir */
 static bool force_yes = false;          /* force -y even when scanning/copying/verifying in -i answers mode */
+static struct stat topdir_st;           /* stat(2) information of topdir */
+static struct stat workdir_st;          /* stat(2) information of workdir */
 
 /*
  * forward declarations
@@ -218,6 +222,8 @@ main(int argc, char *argv[])
     char *make = MAKE_PATH_0;                   /* path to make(1) executable */
     char *answers = NULL;			/* path to the answers file (recording input given on stdin) */
     FILE *answersp = NULL;			/* file pointer to the answers file */
+    char *uuid = NULL;			/* path to the UUID file */
+    FILE *uuidp = NULL;			/* file pointer to the UUID file */
     char *submission_dir = NULL;		/* submission directory from which to form a compressed tarball */
     char *tarball_path = NULL;			/* path of the compressed tarball to form */
     struct info info;				/* data to form .info.json */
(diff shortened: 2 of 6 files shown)


lcn2:
After the IOCCC Judges individually work on their individual tax forms, speaking of obfuscation :-(, we plan to look at these poll requests one at a time test and process them.

That may take some time, including, but not withstanding the time spent in deobfuscation of tax forms.  

Nevertheless, @xexyl, we **DO** appreciate the effort you have made in your recent PR requests, and will give them the testing and care that they deserve in due time.

xexyl:
> After the IOCCC Judges individually work on their individual tax forms, speaking of obfuscation :-(, we plan to look at these poll requests one at a time test and process them.
> 
> That may take some time, including, but not withstanding the time spent in deobfuscation of tax forms.
> 
> Nevertheless, @xexyl, we **DO** appreciate the effort you have made in your recent PR requests, and will give them the testing and care that they deserve in due time.

Good luck an take your time! Best wishes.

lcn2:
We suggest, @xexyl, the following mod, if it is not to difficult, to this PR:

Add `-U UUID` (we suggest using this that case as well), which will conflict with `-u uuid`, that has the same effect in setting the UUID expect that instead of reading a file, the `optarg` is the value.

This would allow someone to set an environment variable and use it:

```sh
export UUID="12345...ef"
mkiocccentry -U $UUID ...
```

The `-U UUID` would be useful for our planned work on the slot directory too.

If you have time, consider amending this PR .. or it if is easier, add another PR to add `-U UUID`.

lcn2:
> The problem with that is it requires always setting it. Where as currently one can just put their uuid in a file and then use -u file.
> 
> Much easier that way. That was the idea originally. Sure your way is better than it currently is but it's more cumbersome.

Some people can set UUID or even IOCCC_UUID in their environment, say under `~/.foorc` (where foo is their current login shell).

So instead of having to keep some file around and remember where it is, the `$IOCCC_UUID` (or `$UUID` or whatever) is used.

Your idea of a file works, others may prefer to keep such values in their environment.  It is matter of personal preference.

> How would having it your idea be better for you? I'm not sure how it would change anything except that a file is easier to use for users - which was the point.

We plan to use multiple UUIDs under slots, and in some cases "mix and match" to create invalid stuff.

> As for whether or not it'd be easier here - another pull request would probably be better as I'll be leaving very soon.

Another PR is fine.

lcn2:
> Resolve issue #1210. The directory/file lists to be ignored prompting was confusing to some people. The question of 'Is this OK?' was reworded to 'Do you wish to continue?' and the explanation is hopefully a bit clearer too.

Good call, @xexyl 👍

xexyl:
> > The problem with that is it requires always setting it. Where as currently one can just put their uuid in a file and then use -u file.
> > Much easier that way. That was the idea originally. Sure your way is better than it currently is but it's more cumbersome.
> 
> Some people can set UUID or even IOCCC_UUID in their environment, say under `~/.foorc` (where foo is their current login shell).
> 
> So instead of having to keep some file around and remember where it is, the `$IOCCC_UUID` (or `$UUID` or whatever) is used.
> 
> Your idea of a file works, others may prefer to keep such values in their environment. It is matter of personal preference.

See below for a compromise.

> 
> > How would having it your idea be better for you? I'm not sure how it would change anything except that a file is easier to use for users - which was the point.
> 
> We plan to use multiple UUIDs under slots, and in some cases "mix and match" to create invalid stuff.
> 
> > As for whether or not it'd be easier here - another pull request would probably be better as I'll be leaving very soon.
> 
> Another PR is fine.

Okay in that case why not have it take as an option an environmental variable called UUID? Then they can decide. I think that would work well as it would allow both modes. Or it could be -U in addition (UUID variable seems more useful to me personally but that's me).

Having to type the variable name kind of defeats my idea and for me personally is not much better than the way it is now - again for me personally. I mean it might be but it's easier to type `-u uuid` than `-u "$UUID"` if you understand me.

lcn2 (approved):
👍

We plan to review and test this PR in more detail later on.  So while it passes the "visual sniff" test (as the expression goes), we will wait until we  have time on a test server to experiment with these mods.

Thank you ￼ @xexyl, for your excellent work on this PR.

lcn2:
> Okay in that case why not have it take as an option an environmental variable called UUID? Then they can decide. I think that would work well as it would allow both modes. Or it could be -U in addition (UUID variable seems more useful to me personally but that's me).
> 
> Having to type the variable name kind of defeats my idea and for me personally is not much better than the way it is now - again for me personally. I mean it might be but it's easier to type -u uuid than -u "$UUID" if you understand me.

Why?  Because the "`optarg`" value of `-U UUID` might be a string set in a shell script, it might be some other environment variable such as `$IOCCC_UUID` or `$IOCCC28` or `$IOCCC_USERNAME` or whatever the uses chooses.   So:

```sh
mkiocccentry -U thing ...
```

Works because "thing" can be a literal string, an environment variable, or anything that evaluates/generated the "thing".

Your `-u uuid` is fine, and apparently works for you.  The `-U UUID` is another model to so something similar, that all.

xexyl:
> > Okay in that case why not have it take as an option an environmental variable called UUID? Then they can decide. I think that would work well as it would allow both modes. Or it could be -U in addition (UUID variable seems more useful to me personally but that's me).
> > Having to type the variable name kind of defeats my idea and for me personally is not much better than the way it is now - again for me personally. I mean it might be but it's easier to type -u uuid than -u "$UUID" if you understand me.
> 
> Why? Because the "`optarg`" value of `-U UUID` might be a string set in a shell script, it might be some other environment variable such as `$IOCCC_UUID` or `$IOCCC28` or `$IOCCC_USERNAME` or whatever the uses chooses. So:
> 
> ```shell
> mkiocccentry -U thing ...
> ```
> 
> Works because "thing" can be a literal string, an environment variable, or anything that evaluates/generated the "thing".
> 
> Your `-u uuid` is fine, and apparently works for you. The `-U UUID` is another model to so something similar, that all.

Oh! Do you mean having both? That's fine. I thought you meant instead of. I was reading quickly and I have a headache. Sure I can do this. I'd be happy to do that. Will do tomorrow. 

Thanks for clarifying and good day! Remind me of I don't get to this though.

lcn2:
> Oh! Do you mean having both? That's fine. I thought you meant instead of. I was reading quickly and I have a headache. Sure I can do this. I'd be happy to do that. Will do tomorrow.

Yes, both.

However it might be a good idea to make the use of both `-u uuid` and `-U UUID` a command line error.

So actually "either or", not both on the same command line.  😉

xexyl:
> > Oh! Do you mean having both? That's fine. I thought you meant instead of. I was reading quickly and I have a headache. Sure I can do this. I'd be happy to do that. Will do tomorrow.
> 
> Yes, both.
> 
> However it might be a good idea to make the use of both `-u uuid` and `-U UUID` on the same command line, to be a command line error.
> 
> So actually "either or", not both on the same command line. 😉

Makes sense to me. So before I leave - to be clear I got this right.

The `-U` option would have the optag be the UUID itself? I'm sure that can be done quite quickly but I better get going.

Best wishes for a good rest of your day!

lcn2:
> The -U option would have the optag be the UUID itself?

Yes.

lcn2:
> I'm sure that can be done quite quickly 

That is just fine.  We are going to take our time testing these PRs, due in part to the fact that **IOCCC28** is still open and changing the "took kit" (or is it the "tool set"?) is **NOT** something to be done in haste.

-- Cue Ent remarks 😉

xexyl:
> > I'm sure that can be done quite quickly
> 
> That is just fine. We are going to take our time testing these PRs, due in part to the fact that **IOCCC28** is still open and changing the "took kit" (or is it the "tool set"?) is **NOT** something to be done in haste.
> 

To be strictly correct it's 'toolkit'.

> -- Cue Ent remarks 😉
> 
> ## UPDATE 0
> .. and after the IOCCC Ents 🌲 have met and come to agreement 🤓

Oh dear.  We're going to be here for a very long time because they do not say anything in Old Entish, unless it is worth taking a long time to say, and to listen to.

xexyl:
> > Resolve issue #1210. The directory/file lists to be ignored prompting was confusing to some people. The question of 'Is this OK?' was reworded to 'Do you wish to continue?' and the explanation is hopefully a bit clearer too.
> 
> Good call, @xexyl 👍

Well for that exact wording we can thank @SirWumpus. It does make much more sense though. I actually did not like the question as is but I was rushing to get things in and never thought to change it.

Fortunately he opened the issue!

xexyl:
> > As for whether or not it'd be easier here - another pull request would probably be better as I'll be leaving very soon.
> 
> Another PR is fine.

As you probably already saw: I opened one a while ago. Going to work on guidelines / FAQ soon I hope.

PR #1227, opened 2025-03-09
Plus loophole in min version checks

xexyl:
Added a max version too. This will prevent someone from having their own 
version that might not be released or official and it can be flagged
(okay so they might still have a version in between but that's
inconsequential).

Updated gen_test_JSON.sh version to "1.0.3 2025-03-09".


xexyl:
Closing due to your excellent reminder @lcn2. I got it fixed on my end but by closing this one it's easy to do. Then I can also fix the FAQ over there - with a new commit.

xexyl:
I believe it's all good over there too - and updated the log on GitHub.

Thanks!

Issue #1208, opened 2025-03-07
Feature: Option to force overwrite of `workdir`


SirWumpus:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the feature

I've had to repeated restart `mkiocccentry` to prepare a submission, but once you get part way through and `mkiocccentry` terminates on an error or user interrupt, it leaves behind an incomplete `workdir`.

An option (`-f` force or `-o` or `-O` overwrite) a workdir would be nice and save repeated need to `rm -rf workdir; mkdir -p worddir`.

xexyl:
Yeah as far as this goes I think the problem is it's dangerous to delete the directory. 

But we will see what Landon thinks. 

What do you think @lcn2  ?

I have usually not been bothered too much but it might also be nice.

xexyl:
.. although later on we do get the absolute path. 

But that's later when scanning topdir.

SirWumpus:
> Yeah as far as this goes I think the problem is it's dangerous to delete the directory.

A working or staging or cache directory are typically far game to nuke.  So I don't understand why it would be dangerous.

lcn2:
If this is addressed before IOCCC28 "closes" (and that is a big **IF*), then the [Important NOTE from GH-](issue 1209) needs to be considered.

SirWumpus:
> If this is addressed before IOCCC28 "closes" (and that is a big *_IF_), then the [Important NOTE from GH-](issue 1209) needs to be considered.

I think it can wait to after IOCCC28.  Its a UX thing that can be fixed later.

lcn2:
> Yeah as far as this goes I think the problem is it's dangerous to delete the directory.
> 
> But we will see what Landon thinks.
> 
> What do you think [@lcn2](https://github.com/lcn2) ?
> 
> I have usually not been bothered too much but it might also be nice.

If someone is going to use some option to "**FORCE((" overwrite of `workdir`, and this change the default behavior, then one must presume that they understand the impact of forcing it.

This is no different than, say `mv -f` or `cp -f` or `rm -f`.

So as long as it is not the default, and someone "**FORCES**" then well .. "that's what they get" (as the expression goes).

xexyl:
> If this is addressed before IOCCC28 "closes" (and that is a big **IF*), then the [Important NOTE from GH-](issue 1209) needs to be considered.

Agreed.

xexyl:
[...]

PR #1231, opened 2025-03-10
Partial fix to issue #1208

xexyl:
The -x option to force delete the submission directory and -r
rm option to set path to rm(1) were added to mkiocccentry as part of
issue #1208. The change in order of args is NOT done and will not be
done until AFTER IOCCC28.

Fixed bug where overwrite_answers was always true by default
(mkiocccentry).


lcn2:
thank you @xexyl

PR #1232, opened 2025-03-10
Fixes for -u uuidfile/-U uuidstr options

xexyl:
Since the two options cannot be used together do not check for one if 
the other failed.

Also tell the user they can use the -x option if the submission 
directory already exists (this comes after the part of telling them to 
move it, remove it or use a different directory).


lcn2 (approved):
thank you @xexyl

Issue #1233, opened 2025-03-10
Bug: `mkiocccentry` invokes `make` instead of `gmake` reporting an error.


SirWumpus:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

It appears that an invocation of `make` by `mkiocccentry` fails on NetBSD (again) instead of invoking `gmake`.

```
If this list is incorrect, you will have to fix your topdir
and try again.

Do you wish to continue? [Yn]:
make[1]: "/home/achowe/git/ver/.stage/UUID-0/Makefile" line 119: Invalid line type
make[1]: "/home/achowe/git/ver/.stage/UUID-0/Makefile" line 125: Invalid line type
make[1]: "/home/achowe/git/ver/.stage/UUID-0/Makefile" line 129: Invalid line type
make[1]: "/home/achowe/git/ver/.stage/UUID-0/Makefile" line 135: Invalid line type
make[1]: Fatal errors encountered -- cannot continue
make[1]: stopped in /home/achowe/git/ver/.stage/UUID-0
Warning: check_submission_dir: make -f Makefile clobber failed
Checking Makefile ...
... completed Makefile check.
```

### What you expect

Detect `gmake` vs `make` on those platforms that have it and use `gmake`.

### Environment

- OS: NetBSD 10.1 VM
- Device: Windows 10 Home using VirtualBox
- Compiler: gcc (GCC) 14.2.0

xexyl:
You can use `-m /path/to/make`.

xexyl:
I have an idea though. I can do it in a bit.

If the use of the default fails it can try `gmake`. If that fails it can then let the user know about the option.

Would that work for you @SirWumpus ?

SirWumpus:
* Is `mkiocccentry` assuming absolute paths for `make` (and other tools) instead of using `$PATH`?
* `-m gmake` didn't work, since it didn't search `$PATH`.
* `-m /usr/pkg/bin/gmake` worked.

>If the use of the default fails it can try gmake. If that fails it can then let the user know about the option.
>
> Would that work for you @SirWumpus ?

I would have expected something similar to how `gmake` is detected and used to build the tools.  In that case I think you check/try for `gmake` before `make`, whereas you just suggested above the opposite order.

xexyl:
But in answer to your questions...

> * Is `mkiocccentry` assuming absolute paths for `make` (and other tools) instead of using `$PATH`?

[...]

Issue #1218, opened 2025-03-08
Enhancement: `mkiocccentry` should search `PATH` for utilities like `txzchk`


tcort:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

The tools from the `mkiocccentry` repository are in my `PATH` but `mkiocccentry` cannot find them. It would be nice if it searched `PATH` for them.

mkiocccentry version: 2.0.1 2025-03-02
uname -a output: Darwin megatron.local 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000 arm64

What I did: 
 
- I cloned the `mkiocccentry` repository to `~/repos/mkiocccentry`.
- I built the utilities `cd ~/repos/mkiocccentry && make`
- I added the directory to my `PATH` with `export PATH=~/repos/mkiocccentry:$PATH`
- I ran the `mkioccentry` tool from `~/repos/ioccc/2025`: `mkiocccentry foo bar`
- I get the output: `We cannot find txzchk.`
- That tool is definitely in my `PATH`:
  - `which txzchk` prints `/Users/tcort/repos/mkiocccentry/txzchk`
  - running `txzchk -h` prints the help information

xexyl:
Yes it explicitly looks at certain places. 

There are options to specify the path. 

It is however recommended to install the tools. 

As far ad this idea - it's interesting and can be discussed for after this one. 

Thanks!

lcn2:
There is a lot of merit to this idea.  And implementing this would be reasonably safe. 

The idea is that if the command line option to provide explicit path to a tool is NOT given, then calling `execvp(3)` with just the base name of the tool could be attempted.  This would cause a search along `$PATH` to be attempted. 

If that failed, then an `execv(3)` with a path that assumes one is running from the repo directory could be used. 

And the that fails, an error could be thrown.

xexyl:
> There is a lot of merit to this idea. And implementing this would be reasonably safe.
> 
> The idea is that if the command line option to provide explicit path to a tool is NOT given, then calling `execvp(3)` with just the base name of the tool could be attempted. This would cause a search along `$PATH` to be attempted.
> 
> If that failed, then an `execv(3)` with a path that assumes one is running from the repo directory could be used.
> 
> And the that fails, an error could be thrown.

[...]

PR #1234, opened 2025-03-10
Resolved issues #1233 and #1218

xexyl:
Both mkiocccentry and txzchk (the only ones that use other tools) now 
search $PATH for the tools first by way of the find_utils() function
(modified a fair bit) and a new util function in jparse called 
resolve_path(). The jparse util functions shell_cmd() and pipe_open()
were also improved to resolve paths if no / is in the command name. As
for make: we now search for gmake first as the Makefiles we need are GNU
Makefiles.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,13 @@ to true if the UUID was `"true"`.
 The `-u uuidfile` and `-U UUID` options may not be used with `-i answers`, `-d`
 or `-s seed`.
 
+Partial fix to issue #1208. The `-x` option to force delete the
+submission directory and `-r rm` option to set path to `rm(1)` were added to
+`mkiocccentry` as part of #1208. The change in order of args is **NOT** done and
+will not be done until **AFTER** IOCCC28.
+
+Fixed bug where `overwrite_answers` was always true by default (`mkiocccentry`).
+
 Updated man page for the above changes.
 
 **IMPORTANT NOTE**: you do NOT need to use this update in order to participate
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,8 @@ called `resolve_path()`. The jparse util functions `shell_cmd()` and
 name. As for `make`: we now search for `gmake` first as the Makefiles we need
 are GNU Makefiles.
 
+Add missing `-r` to `rm` in `mkiocccentry_test.sh`.
+
 **IMPORTANT NOTE**: you do NOT need to use this update in order to participate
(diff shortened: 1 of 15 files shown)


xexyl:
Okay that's weird ..  have an idea what might have gone wrong.

xexyl:
If nothing else this should fix an issue that I overlooked. We'll see if I missed something else.

xexyl:
Ah. This should be simple to fix. Looking at why this happened.

xexyl:
A `-r` was missing from `rm` in mkiocccentry_test.sh. I'm not sure how that was never detected.

xexyl:
Spellcheck failure now. Not sure how that happened either. Looking at it ....

xexyl:
Oh I know how. It's a good catch too.

xexyl:
Oddly: there was a `/` after a variable (a directory!) in the script. The `rm` needed `-r` of course and the `/` was not removed. Spellcheck locally does not complain now but I'm glad it did - even though that variable **should** always be set.

xexyl:
Well this is so very useful:


```
Error:
```

xexyl:
Each time something new is going wrong but `make prep` worked fine here. Hmm ....

xexyl:
I have this feeling it's to do with the $PATH change. I have a way to test it.

xexyl:
There's something screwy going on. I know the problem I just have to figure out why it's happening.

xexyl:
It's actually quite amusing, the problem. Have to figure out how to fix it.

xexyl:
.. actually it's to me hilarious. I have a fix.

xexyl:
Fixed that problem:

    The problem with having 'txzchk' before './txzchk' is that if it's not
    installed but one is in the mkiocccentry directory the functions
    is_file() and is_exec() will happily report that it's an executable file
    but when it comes time to executing it it won't work because '.' is not
    in $PATH.

xexyl:
There we go!

Lesson is probably something I won't end up doing but: if testing a problem one should uninstall the tools to really see what is going on in an environment that does not have the tools installed.

lcn2 (approved):
Thank you 🙏 @xexyl 

After processing this PR, we should be in a position to form a new formal release of this repo.  However, we will review the status of this repo **TOMORROW** and perform more test, before forming the release.

Then after the new formal release of this repo has been made (tomorrow), we will process the related PRs in "the other repo" and then announce a summary of these changes in IOCCC news, Mastodon and Discord.

PR #1236, opened 2025-03-11
Fix search of tools under $PATH

xexyl:
Due to the fact we have to have (for our tools) ./ ahead of other paths (including the tool names themselves) and also allow for a user to specify a path (even a bogus one, which if they do we will look for a proper path) and also search for installed paths, a more careful check has to be done in the find_utils() functions as otherwise it does not work when out of the repo directory.

Sync copyright fixes from jparse.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,20 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.4 2025-03-11
+
+Fix search of tools under `$PATH`.
+
+Due to the fact we have to have (for our tools) `./` ahead of other paths
+(including the tool names themselves) and also allow for a user to specify a
+path (even a bogus one, which if they do we will look for a proper path) and
+also search for installed paths, a more careful check has to be done in the
+`find_utils()` functions as otherwise it does not work when out of the repo
+directory.
+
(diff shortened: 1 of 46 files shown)


xexyl:
Hopefully GitHub doesn't have any issues with this change either. If it does not then it's all good now I think. I actually saw the oddity in the debug output yesterday but as I was in the repo directory at the time it did not dawn on me as odd. The ordering issue has turned out to be a lot of fun, for funny definitions of fun.

lcn2:
# QUESTION 

Do we need to perform the idea about looking for the source files (see [GH-issuecomment-2715178275](https://github.com/ioccc-src/mkiocccentry/issues/1233#issuecomment-2715178275)) as an additional safety check before simply executing tools out of "."?

xexyl:
> # QUESTION
> Do we need to perform the idea about looking for the source files (see [GH-issuecomment-2715178275](https://github.com/ioccc-src/mkiocccentry/issues/1233#issuecomment-2715178275)) as an additional safety check before simply executing tools out of "."?

Nope. I replied in that comment how I did it. I just tested it again. However I can test it when it's not installed and out of the directory: though it should be fine - I'd have to give the options to the tools.

PR #1237, opened 2025-03-11
Disable certain iocccsize warnings

xexyl:
Force set nul_warning to false in mkiocccentry(1) and disable checking of it in chkentry(1). Post IOCCC28 it will be removed from struct info, the .info.json files and chkentry(1) that checks it.

Rather than directly put in true in the writing of the .info.json file, force set first_rule_is_all to true in mkiocccentry(1) and reference that boolean in the writing of the .info.json file.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,20 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.4 2025-03-11
+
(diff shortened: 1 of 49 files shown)


xexyl:
Did I get everything now (i.e. are all the warnings to be disabled now actually disabled) @lcn2 ?

lcn2 (approved):
👍

PR #1260, opened 2025-05-02
Add note about chkentry -i aliases to chkentry.1

xexyl:


xexyl:
Closing in favour of removing that alias code you had me add. I agree it can be problematic.

PR #1261, opened 2025-05-02
Remove chkentry -i alias feature

xexyl:


xexyl:
> Thank you @xexyl
> 
> We apologize for suggesting this "misfeature" be added.

It's okay! Fortunately it took not even five seconds. The only downside is a joke had to be removed. Could have kept the array in but that seemed wasteful ... it was:

```c
    { "try"         , TRY_SH                }, /* try not to keep sorted */
    { "try.alt"     , TRY_ALT_SH            }, /* alternatively, try and keep sorted */
```

Oh well. Not the end of the world (but if it does end I'll blame you ... well actually I won't since I wouldn't be here and neither would you). But it is amusing enough to mention again.

PR #1267, opened 2025-06-06
Fixes in -i answers with -Y

xexyl:
When -Y is used, don't show warnings even if -i answers is used as that
was the point of it (one should use this with caution as it answers yes
to every question automatically). It is unclear if this should show the
list of files/directories to be copied (showing the contents of the
tarball is part of txzchk's algorithm so this is unrelated) as it
currently does.  It is also unclear if the use of -Y should show
warnings but just not confirm. Given that one is just an informational
notice and that one does say something (at least if the function is
called) it might be better to always show but that really depends on
what other people want (the point of -Y in any case is to always answer
yes even when the answers file is used, as some people do in fact want
this).

Updated MKIOCCCENTRY_VERSION to "2.0.9 2025-06-06".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.9 2025-06-04
+
+Fix bug where `-i answers` did not warn about issues found in prog.c and
+Makefile.
+
+
 ## Release 2.4.8 2025-05-05
 
 Allow a submitter's system clock to be as much as 48 hours in the future before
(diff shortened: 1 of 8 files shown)


lcn2 (approved):

xexyl:
> Before we head out to town, we are processing the mkiocccentry PRs.
> 
> 

Safe travels (if you mean out of town as I first read it as I know you're going on holiday soon)!

Either way hope you have a great rest of your day and if you did mean your trip then safe travels and have fun and stay safe!
> 
> Thank you 🙏 @xexyl and thanks for your patience with this PR.

It's quite all right and you're very welcome!

I wanted to get these things done so I have more time this next contest.

PR #1270, opened 2025-06-11
Resolve issue #1207 - ignore SCCS and RCS

xexyl:
It might be worth noting that if one has an ignored directory name as a
file that it appears to not warn about it even though it will (it
appears as well) ignore it. It is unclear to me if this should be
changed or not.

It would seem that both chkentry and txzchk have this problem as well. A
new issue for each will be opened to be resolved before IOCCC29.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.9 2025-06-04
+
+Fix bug where `-i answers` did not warn about issues found in prog.c and
+Makefile.
+
+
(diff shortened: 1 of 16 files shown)


xexyl:
> Before we head out to town, we are processing the mkiocccentry PRs.
> 
> 

Safe travels (if you mean out of town as I first read it as I know you're going on holiday soon)!

Either way hope you have a great rest of your day and if you did mean your trip then safe travels and have fun and stay safe!
> 
> Thank you 🙏 @xexyl and thanks for your patience with this PR.

It's quite all right and you're very welcome!

I wanted to get these things done so I have more time this next contest.

PR #1273, opened 2025-06-19
Support globs in mkiocccentry ignore option

xexyl:
Added support of globs with the mkiocccentry ignore option (-I). This 
involves syncing jparse/ from the jparse repo. It would appear that
GitHub changed something with workflows so this might fail the workflow 
check. If so that might require a new issue or looking at later. In any
case make prep works fine even if it does fail the workflow (it is not
clear to me if it's due to the dependencies but the last commit to
jparse worked fine and the build steps did not change at all).

Updated MKIOCCCENTRY_VERSION to "2.0.10 2025-06-19".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.9 2025-06-04
+
+Fix bug where `-i answers` did not warn about issues found in prog.c and
+Makefile.
+
+
 ## Release 2.4.8 2025-05-05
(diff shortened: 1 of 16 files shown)


xexyl:
Oh in this case it's a lack of a header file ... which is weird since I didi `make` and `jparse/util.h` should pull it in. Hopefully an upcoming commit will fix that part if nothing else.

xexyl:
Workaround seems to work fine.

lcn2 (approved):

Issue #1228, opened 2025-03-10
Enhancement: consider an --manifest option to mkiocccentry(1) for IOCCC29


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

As discussed in issue #1216, and somewhat in issue #1212 we propose a discussion of a new option for `mkiocccentry(1)` for **IOCCC29**:

```
mkiocccentry --manifest file ...
```

The use of `-manifest file` would cause `mkiocccentry(1)` to bypass the normal collection of files under `topdir`.

The `file`, which **MUST** exist as a readable non-empty file, will contain a list of files, one per line.

Leading and trailing whitespace (including the trailing newline) in `file` will be ignored.

Any text in `file`, starting with a "#" (i.e., hash or pound sign) until the end of the line will be ignored.

Empty lines in `file` will be ignored.

All file paths in `file` will assumed to be relative to `topdir`.

All of the components of file paths in `file` (with the possible exception of a leading "./") must undergo all of the normal checks that scanning under `topdir` would perform.

If any the components of file paths in `file` (with the possible exception of a leading "./")  are found to be invalid, then an exception is thrown and an error messages(s) giving the offending paths in `file` will be displayed.  This will allow someone to "pre-test" their manifest to determine of something is wrong.

If any the components of file paths in `file` are normally ignored, then an exception is thrown and an error messages(s) giving the offending paths in `file` will be displayed.  This will allow someone to "pre-test" their manifest to determine of something in their manifest is normally something that is ignored.

All of the normal checks on things such as the required files, the number of extra files, directory counts, directory depth limits, etc. will apply to the paths in `file` (i.e., the `file` maifest **MUST**  conform to the rules relating to the manifest of a valid submission).

The `file` **MUST** contain a complete manifest (i.e., nothing may be missing).

The `file` **MUST** contain **ONLY** the manifest (i.e., **NOT** contain anything that is normally ignored). 

If `file` is a valid manifest, then there will be nothing missing, nothing to ignore, no rule violations, etc.

While there will be **NO** mercy (as the expression goes).  If there is anything wrong with the manifest list, then `mkiocccentry(1)` **MUST** exit non-zero.  However, to be slightly nice, before exiting non-zero, the complete list of problems with the manifest **MUST** be displayed so that the user has a chance to correct the problems.

### Relevant images, screenshots or other files

`file` :-)

### Relevant links

- issue #1216 
[...]

PR #1274, opened 2025-06-20
Resolve issue #1228

xexyl:
Address issue #1228. By request one may use the -M manifest_file option 
to specify what files to include in the topdir. This required updates to
jparse/ (from the jparse repo) and also soup/. Due to the nature of the 
lists of files (there being separate lists depending on the file) the
check_submission_dir() function does not check that the directory does
not have files not in the list.  But given that the FTS functions would
not find those files EXACTLY BECAUSE OF this option it can't be checked
so this should not matter anyway. What it can and does check is that the
required files are there and that everything else is sane, even if by
some chance this option causes some insanity (i.e. a fileset that is not
what one might usually expect) :-) (diagnosis not yet made :-) ).

Updated MKIOCCCENTRY_VERSION to "2.0.11 2025-06-20".
Updated SOUP_VERSION to "2.0.3 2025-06-20".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.4.9 2025-06-04
+
+Fix bug where `-i answers` did not warn about issues found in prog.c and
+Makefile.
+
+
 ## Release 2.4.8 2025-05-05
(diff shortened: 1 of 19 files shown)


lcn2 (approved):

Issue #1216, opened 2025-03-07
Enhancement: Add an command line option to mkiocccentry to provide an ignore list file


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

Add `--ignore_list file` (or some other option name) that has an effect of `-I foo -I bar -I curds ... -I whey` where those files are found in a file called `file`.

Suggestion:

The file should be one file per line

Lines starting with # are comments and are ignored

Empty lines are ignored.

Leading and trailing whitespace are ignored.

### Relevant images, screenshots or other files

n/a

### Relevant links

See the closed issue #1212 and ask @SirWumpus 

### Anything else?

We recommend using a `--option` style for this more complex command line option.

This is an enhancement to consider for **IOCCC29**.

xexyl:
> ### Relevant links
> 
> See the closed issue [#1212](https://github.com/ioccc-src/mkiocccentry/issues/1212) and ask [@SirWumpus](https://github.com/SirWumpus)
> 

Brilliant. Thanks for the laugh!

Feel free to assign - for consideration down the road.

SirWumpus:
> See the closed issue #1212 and ask @SirWumpus

hehehehe. first smile i've had all day.

lcn2:
P.S. Consider also the idea of `--manifest file` where `file` is the **ONLY** list of files that are allowed (unless they are files that are to be ignored for other reasons) and that everything else is ignored.

xexyl:
> P.S. Consider also the idea of `--manifest file` where `file` is the **ONLY** list of files that are allowed (unless they are files that are to be ignored for other reasons) and that everything else is ignored.

That will be extremely problematic and it'd mean we would have some code not use fts(3) and some code use it. And it would mean we couldn't make as many checks. And it's really not any different from the old model is it? Have to specify a list of files rather than the simpler approach.

I do find it both amusing and bemusing that I was the one agains the change original and @SirWumpus was the one for it and now it's reversed. But we'll see.

lcn2:
> That will be extremely problematic and it'd mean we would have some code not use fts(3) and some code use it. And it would mean we couldn't make as many checks. And it's really not any different from the old model is it? Have to specify a list of files rather than the simpler approach.

You make a good point.

SirWumpus:
Can will table this topic for much much later (after IOCCC28), preferably when my head is not throbbing.

xexyl:
[...]

PR #1275, opened 2025-06-21
Address issue #1216 and fix fnmatch(3) bug

xexyl:
Address issue #1216. By using -X file (X for eXclude) one can provide a 
file with a list of path to ignore (one path per line). Like with -I
path and -M manifest_file one may have (in the file itself) globs (based
on the rules of fnmatch(3)). The reason a long option is not used is
because GNU and BSD have different behaviours in getopt_long(3) in ways
that might cause problems. This can be looked at later if necessary.

Fix bug with globs in manifest and ignore lists. Now when checking lists 
that can have a glob in an item it uses the updated jparse routines that
take a new bool fn.

Updated MKIOCCCENTRY_VERSION to "2.0.12 2025-06-21".
Updated CHKENTRY_VERSION to "2.0.6 2025-06-21".
Updated SOUP_VERSION to "2.0.4 2025-06-21".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,12 @@ Fix bug with globs in manifest and ignore lists. Now when checking lists that
 can have a glob in an item it uses the updated jparse routines that take a new
 `bool fn`.
 
+Change `append_unique_filename()` to not abort when the filename already exists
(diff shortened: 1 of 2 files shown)


lcn2 (approved):

Issue #1272, opened 2025-06-11
Bug:  chkentry needs to check forbidden dirnames and filenames as BOTH directories AND files


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

Because of an option in chkentry we have split the ignored and forbidden filenames into two arrays. It seems chkentry only checks one of those. It does check directory names.

It only checks ignored filenames in winning entry mode. It might be that it should check in both modes.

What's more is that the check might (I did not test this one out) need to be applied to the other type - as in a filename with the name of a forbidden directory name might slip through. That might not be the case - it appears in mkiocccentry it ignores both but does not warn about the other file type (i.e. if a file exists with the name of an ignored directory it won't warn about it but it won't add it either). I did not test that thoroughly however.

The latter part might not be an issue but it might be. In any case the other concerns should be considered.

### What you expect

That nothing should slip through.

### Environment

- OS: n/a
- Device: ''
- Compiler: ''

### bug_report.sh output

n/a

### Anything else?

Obviously this should be done before IOCCC29. See also issue #1271 which is similar.

xexyl:
I was pondering working on this but I'm not sure what part(s) (if any) are actually an issue so I'll wait for your feedback @lcn2 - which I guess won't happen until after judging concludes.

xexyl:
It is also unclear to me if all of this is even a problem. I just tested the different file type with a forbidden filename and both txzchk and mkiocccentry flagged it.

If however someone has a directory as say README.md it can cause an error that is confusing. I will thus have the add unique filename function not abort with an error if it's found. Instead it'll simply do nothing.

xexyl:
Oh great. I see the XXX comment but I haven't the slightest clue what it was all about ... I should have included a command that triggered it. I'll see if git blame helps out but only a quick check. I suggested there might be an issue with the find_paths() routine but I'm not sure of that now. It might have been an error in thinking when writing the code. 

I don't actually think there is a bug in the routine. If that's so then it might be that there is no problem and I can hopefully look at another issue - though I've resolved a number of these issues that are open whilst you're judging.

xexyl:
It has to do also with this log:

    Add temporary workaround issue #1258
    
    Add temporary workaround for issue #1258. It might be that the real fix
    involves an update to a jparse routine but during the IOCCC28 itself
    that is not going to be looked at and it is arguable it is not even a
[...]

PR #1276, opened 2025-06-22
Address issue #1272

xexyl:
Address issue #1272. There indeed was an issue in an FTS routine in 
jparse/util.c. The issue was that the -i path option added to the struct
fts but when reset_fts() was called in all cases the ignore list was set
to NULL, even when it was already initialised AND free_ignored was
false. As of yesterday or the day before this bug was fixed so the
additional code in chkentry was removed as the workaround is no longer
needed.

Updated CHKENTRY_VERSION to "2.0.7 2025-06-22".


lcn2 (approved):

PR #1279, opened 2025-06-23
Fix issues count in txzchk for size of files

xexyl:
Only count total file size > max as one issue (feather) in txzchk. The 
warning for each time it goes over is still printed as the sum and count
function is run per line parsed but once the size limit has been broken
there is no need to raise an issue after the first time.

Updated TXZCHK_VERSION to "2.0.4 2025-06-23".


lcn2 (approved):

Issue #1222, opened 2025-03-08
Enhancement: Remove first_rule_is_all from .info.json for IOCCC29


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

The code to test if `all` is the first rule in a `Makefile` is problematic at best.

### Relevant images, screenshots or other files

Any current **IOCCC28** `.info.json` file.

### Relevant links

Issue #1221 

### Anything else?

For **IOCCC29** (and later contests):

- Amend the code to **NOT** check if all is the first rule.
- Remove `first_rule_is_all` from the `.info.json` file format
- Update `IOCCC_info_version`
- Remove, as needed, references to `first_rule_is_all` in "the other repo" (such as in the FAQ)

xexyl:
Already did it .. have to test of course but should be good.

lcn2:
> Already did it .. have to test of course but should be good.

Did what "it"?

The `.info.json` version cannot be changed for **IOCCC28**.  Moreover this is an **IOCCC29** matter.

xexyl:
> > Already did it .. have to test of course but should be good.
> 
> Did what "it"?
> 
> The `.info.json` version cannot be changed for **IOCCC28**. Moreover this is an **IOCCC29** matter.

I had thought you said this should be done now. I can certainly undo this although unfortunately that means more juggling with the changes I have made.

xexyl:
And yes I know it says IOCCC29 here - I had already finished it before you opened this issue as it only took a minute.

lcn2:
> > > Already did it .. have to test of course but should be good.
> > 
> > 
> > Did what "it"?
> > The `.info.json` version cannot be changed for **IOCCC28**. Moreover this is an **IOCCC29** matter.
> 
> I had thought you said this should be done now. I can certainly undo this although unfortunately that means more juggling with the changes I have made.

Changing the format of `.info.json` is a **HUGE** change ‼️

One most certainly should **NOT** modify the format of a fundamental file such as `.info.json` while the **IOCC28** is open ‼️

Just force the value of `first_rule_is_all` to be `true` in `.info.json`.   Let the rest of the code base think that all is well with a `true` value for `first_rule_is_all` in the existing `.info.json` files for **IOCCC28**.

[...]

Issue #1171, opened 2025-02-20
Enhancement: make it very easy to patch or use a non-patch method for the chkentry semantics tables


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

A quite problematic issue with the semantic tables of chkentry is that if some new field has to be added or changed we have to jump through hoops to get chkentry(1) to not complain.

There have been numerous ideas we have floated but the most recent one (that both @lcn2 and I have thought of individually) is (as he wrote):

> Instead a more descriptive file / file format should be used to indicate what needs to be changed. A reference JSON file can produce a table, but then something needs to modify that table to indicate that something might occur with a wider count range, for example.

### Anything else?

Not yet.

xexyl:
Although this is technically more of an enhancement I'm almost tempted to call it a 'bug': or at least a most annoying mis-feature. But of course it works well as long (as we don't have to modify anything) and something had to be done.

Of course working on this is post IOCCC28 but now we can at least discuss it (or not forget it as easily). 

I do wonder if the jsemtblgen related code in jparse will have to change (I hope not) but I guess we'll see about that later on.

lcn2:
We believe we have addressed all of the current questions that still need answering at this time. If we've missed something or something else needs to be clarified, please ask again.

lcn2:
# Important checkpoint

Please see [GH-issuecomment-2762004417](https://github.com/ioccc-src/mkiocccentry/issues/723#issuecomment-2762004417) for a recent comment on existing issues, priorities, and on a deadline for going into pending state for **IOCCC29**.

xexyl:
Okay this issue is a hairy one ... it requires A LOT of thought and care and it might be better to have it in jparse as the semantics tables stuff is from jparse.

I am not closing this as a reminder but I am not sure how critical this is. I would say that this is not critical before IOCCC29. It is a nice to have but we both know how to do it otherwise (though I have to think about it every time it's still doable - although having to patch is still annoying and not very user friendly).

xexyl:
> **TODO**: When this issue is resolved, add a tables for the semantic check of `.entry.json` and then update `chkentry.c` to pass a non-NULL `struct json_sem` pointer to the `chk_json()` function.

I believe that you were going to take care of this one as you have some ideas on it?

Earlier today I replied to your replies to my questions in the other issue. 

[...]

PR #1280, opened 2025-06-24
Resolve issue #1222

xexyl:
Resolve issue #1222. Removed the JSON members (from .info.json file):
first_rule_is_all, highbit_warning and nul_warning. The template
.info.json file
(test_ioccc/template.test_JSON/info.json/good/info.reference.json)  was
updated to no longer have these members and the other JSON files under
test_ioccc/template.test_JSON/ that referred to the members no longer do (this
was done with my [sgit](https://github.com/xexyl/sgit) and manual editing of the
two files without spaces outside of strings).

The JSON files under jparse which have these members were not updated as that's
out of scope and is entirely unnecessary (although a couple minor updates were
made to its README.md file).

The files under soup/ including the semantics tables (base and patch files
were updated, some automatically and some by hand - see below on this point).
Other code was updated as well for this update.

mkiocccentry no longer checks, warns or writes these members to the .info.json
file. The struct info no longer has these bools. If chkentry were to
encounter any of these members it will now flag it as an error.

I observe that the iocccsize related file soup/rule_count.c does in fact still
refer to nul_warning but it does not refer to highbit_warning. This is
because the NUL one came later; @SirWumpus directed me in this after Landon
requested I remove it. But as neither mkiocccentry nor chkentry care about this
it won't cause any problems.

This took a lot of work and all the more reason does issue #1171 need to be
resolved. However that will very likely require an update to the [jparse
repo](https://github.com/xexyl/jparse) as well. But in any case the issue #1222
should now be resolved.


xexyl:
As for you @SirWumpus: I won't update that code without Landon's permission and your guidance but if that is helpful (and it is helpful to him) I'd appreciate it. But I guess we have to wait for judging to conclude first.

I almost did not mention you here but it felt worth noting as that file is mostly read-only to all but you and Landon.

PR #1282, opened 2025-06-25
Sync jparse/ from jparse repo

xexyl:
UTF-8 library updates. This was not said in its changelog but a new 
function was added that might prove useful down the road but in any case
belongs with the code.


xexyl:
Oh dear .. I just noticed something here. Some surrogate pairs are being rejected as out of range. Hmmm .... have to look into this one.

lcn2 (approved):

PR #1283, opened 2025-06-26
Sync jparse/ from jparse repo

xexyl:
Sync jparse/ from the jparse repo (https://github.com/xexyl/jparse/).  
Made last calls to malloc() use calloc() (outside of the generated code,
of course) instead - and do not assume the type either, just use foo =
calloc(size, sizeof(*foo)).


xexyl:
This should have been done ages ago. Unsure if it's done in this repo or any of the others.

PR #1285, opened 2025-07-04
Sync jparse from jparse repo for UTF-8 updates

xexyl:
Sync jparse/ from jparse repo (https://github.com/xexyl/jparse/). This 
cleans up the function surrogate_pair_to_codepoint() (no debug output is 
actually correct, at least in the invalid range) and adds the function
is_surrogate_pair(). We don't need that function as the JSON decode
string functions would be needlessly complicated; it's in case someone
needs to check if a \uxxxx\uxxxx is a valid surrogate pair but not
necessary parse it (like the surrogate_pair_to_codepoint() does).


lcn2 (approved):

PR #1286, opened 2025-07-08
Minor update to chkentry - fix submission_dir

xexyl:
Fix submission_dir in chkentry(1) to be a char *, not a char **.

Updated CHKENTRY_VERSION to "2.0.9 2025-07-08".


xexyl:
This reminds me that the chkentry_test.sh needs to be redesigned some ... or at least it might be wise to do it.

PR #1301, opened 2025-08-30
Fixes from 730fcd70bbdd8beb5d0e3fbf4c843860cc6e6a7d

xexyl:
A good catch was the chksubmit only having -S in the debug log. That slipped through.

A nice hack was for the -v level option. It might be worth (although I don't feel up to doing this) trying to come up with a way that handles the low level routines that shell_cmd() and pipe_open() use so we don't have to use this kind of hack. The reason I did not do the -v level is exactly this problem and the -J level is for the same reason (plus that did not seem useful). If these functions are updated then it would be worth updating the other programs too, perhaps.

However there was an issue in the jparse/test_jparse/Makefile. It should not refer to anything from this repo. As such that Makefile was updated at jparse and synced here. The other changes have to be tested but I presume they will be fine (they are thus not committed over there yet for this reason and they are not even added yet let alone tested).

### DIFF
--- a/jparse/test_jparse/Makefile
+++ b/jparse/test_jparse/Makefile
@@ -665,7 +665,7 @@ clobber: legacy_clobber clean
 	${S} echo "${OUR_NAME}: make $@ starting"
 	${S} echo
 	${Q} ${RM} ${RM_V} -f ${TARGETS}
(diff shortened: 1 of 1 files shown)


lcn2 (approved):
👍

PR #1302, opened 2025-08-30
Fix bugs in txzchk and mkiocccentry

xexyl:
Although it mostly matters for islower() and isupper() it is safer to
check isascii() before passing it on to one of the `is*()` functions. This
is because of how some implementations do the functions.  This probably
does not matter much but it's now safer.

Be more aggressive in errors with txzchk. Although it's still a warning
for most issues with the tarball if a NULL pointer is encountered rather
than skipping to the next line it is an error.

Added test file for txzchk_test.sh that results in a NULL pointer to
test the above (it should have already been there).

Fix assumption about find_utils() in mkiocccentry and txzchk. Before
freeing the char * of each tool, also check that it was found. The calls
to find_utils() do (for the tools needed) do pass non-NULL and if the
tool is already specified at the command line it still strdup()s it. But
the two tools now check found_foo before freeing the strings.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,9 +13,6 @@ Be more aggressive in errors with txzchk. Although it's still a warning for most
 issues with the tarball if a NULL pointer is encountered rather than skipping to
 the next line it is an error.
 
-Added test file for `txzchk_test.sh` that results in a NULL pointer to test the
-above (it should have already been there).
-
 Fix assumption about `find_utils()` in mkiocccentry and txzchk. Before freeing
 the `char *` of each tool, also check that it was found. The calls to
 `find_utils()` do (for the tools needed) do pass non-NULL and if the tool is
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,9 @@ the `char *` of each tool, also check that it was found. The calls to
 already specified at the command line it still strdup()s it. But the two tools
 now check `found_foo` before freeing the strings.
 
+Updated `MKIOCCCENTRY_VERSION` to `"2.1.2 2025-08-30"`.
+Updated `TXZCHK_VERSION` to `"2.0.6 2025-08-30"`.
+
 
 ## Release 2.5.1 2025-08-28
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,27 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 2.5.2 2025-08-30
+
+Fix bugs in txzchk and mkiocccentry.
+
+Although it mostly matters for `islower()` and `isupper()` it is safer to check
+`isascii()` before passing it on to one of the `is*()` functions. This is
+because of how some implementations do the functions. This probably does not
+matter much but it's now safer.
+
+Be more aggressive in errors with txzchk. Although it's still a warning for most
+issues with the tarball if a NULL pointer is encountered rather than skipping to
+the next line it is an error.
+
+Added test file for `txzchk_test.sh` that results in a NULL pointer to test the
+above (it should have already been there).
+
+Fix assumption about `find_utils()` in mkiocccentry and txzchk. Before freeing
+the `char *` of each tool, also check that it was found. The calls to
+`find_utils()` do (for the tools needed) do pass non-NULL and if the tool is
+already specified at the command line it still strdup()s it. But the two tools
+now check `found_foo` before freeing the strings.
+
 
 ## Release 2.5.1 2025-08-28
 
--- a/jparse/test_jparse/Makefile
+++ b/jparse/test_jparse/Makefile
@@ -665,7 +665,7 @@ clobber: legacy_clobber clean
 	${S} echo "${OUR_NAME}: make $@ starting"
 	${S} echo
 	${Q} ${RM} ${RM_V} -f ${TARGETS}
-	${Q} ${RM} ${RM_V} -f jparse_test.log chksubmit_test.log txzchk_test.log
+	${Q} ${RM} ${RM_V} -f jparse_test.log
 	${Q} ${RM} ${RM_V} -f tags ${LOCAL_DIR_TAGS}
 	${Q} ${RM} ${RM_V} -rf a b
 	${S} echo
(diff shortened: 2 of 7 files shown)


xexyl:
`make test` passes here ... don't know what's wrong. Will try looking at log here again but might not have the time or energy for it ..

xexyl:
It's txzchk_test.sh but it passes here and there aren't missing files ? I haven't the foggiest why this is failing.

xexyl:
The log here says the error file does not match but rebuilding it and no changes are here .. what is going on? I committed the file. Rebuilding the file manually it's the same ... running the script manually it's the same ... unless it needs `make seqcexit` but I did that I think .. checking ... this is so weird.

xexyl:
Hopefully this sequencing of exit codes will solve the problem ... have to go soon. Very soon.

xexyl:
What is happening here? This makes no sense at all ... surely having this test file is a good thing but it is the new file according to the log ... the error files do match unless there is an issue with GitHub itself. I suppose that might be the case. But even so it's output from txzchk not tar ... and it's from text files. I'll delete the file even though it's unfortunate. Then I must go.

lcn2:
> What is happening here? This makes no sense at all ... surely having this test file is a good thing but it is the new file according to the log ... the error files do match unless there is an issue with GitHub itself. I suppose that might be the case. But even so it's output from txzchk not tar ... and it's from text files. I'll delete the file even though it's unfortunate. Then I must go.

We see no problems with `make test` after doing either a `make prep` or a `make release`.  On all systems the `txzchk_test.sh` passed:

```
RUNNING: test_ioccc/txzchk_test.sh

test_ioccc/txzchk_test.sh -t /usr/bin/tar -T ./txzchk -F ./test_ioccc/fnamchk -d test_ioccc/test_txzchk -Z /home/chongo/bench/ioccc/mkiocccentry

PASSED: test_ioccc/txzchk_test.sh
```

xexyl:
> > What is happening here? This makes no sense at all ... surely having this test file is a good thing but it is the new file according to the log ... the error files do match unless there is an issue with GitHub itself. I suppose that might be the case. But even so it's output from txzchk not tar ... and it's from text files. I'll delete the file even though it's unfortunate. Then I must go.
> 
> 
> 
> We see no problems with `make test` after doing either a `make prep` or a `make release`.  On all systems the `txzchk_test.sh` passed:
> 
> 
> 
> ```
> 
> RUNNING: test_ioccc/txzchk_test.sh
> 
> 
> 
> test_ioccc/txzchk_test.sh -t /usr/bin/tar -T ./txzchk -F ./test_ioccc/fnamchk -d test_ioccc/test_txzchk -Z /home/chongo/bench/ioccc/mkiocccentry
> 
> 
> 
> PASSED: test_ioccc/txzchk_test.sh
> 
> ```
> 
> 

Well that's the odd thing: when I had the file locally it worked fine. But until I removed it from the repo the same script failed in the workflow!

It makes no sense to me. It's not a huge deal but it would be nice if it could be in the test suite.

lcn2:
> > > What is happening here? This makes no sense at all ... surely having this test file is a good thing but it is the new file according to the log ... the error files do match unless there is an issue with GitHub itself. I suppose that might be the case. But even so it's output from txzchk not tar ... and it's from text files. I'll delete the file even though it's unfortunate. Then I must go.
> > 
> > 
> > 
> > We see no problems with `make test` after doing either a `make prep` or a `make release`.  On all systems the `txzchk_test.sh` passed:
> > 
> > 
> > 
> > ```
> > 
> > RUNNING: test_ioccc/txzchk_test.sh
> > 
> > 
> > 
> > test_ioccc/txzchk_test.sh -t /usr/bin/tar -T ./txzchk -F ./test_ioccc/fnamchk -d test_ioccc/test_txzchk -Z /home/chongo/bench/ioccc/mkiocccentry
> > 
> > 
> > 
> > PASSED: test_ioccc/txzchk_test.sh
> > 
> > ```
> > 
> > 
> 
> Well that's the odd thing: when I had the file locally it worked fine. But until I removed it from the repo the same script failed in the workflow!
> 
> It makes no sense to me. It's not a huge deal but it would be nice if it could be in the test suite. 

Sorry, pronouns were getting in the way of understanding the nature of the technical problem being presented.

We believe you are saying that the following test fails:

``` sh
$ test_ioccc/txzchk_test.sh -t /usr/bin/tar -T ./txzchk -F ./test_ioccc/fnamchk -d test_ioccc/test_txzchk -Z /home/chongo/bench/ioccc/mkiocccentry
```

We can confirm that this is a problem that needs to be fixed.

From what we think you have said above, we need to fix this test first.

xexyl:
> Opened issue #1303 in order to address the test problems identified in the comments above.

Wait .. are you seeing this only in the GitHub workflow? Because I fixed this.

xexyl:
> > > > What is happening here? This makes no sense at all ... surely having this test file is a good thing but it is the new file according to the log ... the error files do match unless there is an issue with GitHub itself. I suppose that might be the case. But even so it's output from txzchk not tar ... and it's from text files. I'll delete the file even though it's unfortunate. Then I must go.
> > > 
> > > 
> > > We see no problems with `make test` after doing either a `make prep` or a `make release`.  On all systems the `txzchk_test.sh` passed:
> > > ```
> > > 
> > > RUNNING: test_ioccc/txzchk_test.sh
> > > 
> > > 
> > > 
> > > test_ioccc/txzchk_test.sh -t /usr/bin/tar -T ./txzchk -F ./test_ioccc/fnamchk -d test_ioccc/test_txzchk -Z /home/chongo/bench/ioccc/mkiocccentry
> > > 
> > > 
> > > 
> > > PASSED: test_ioccc/txzchk_test.sh
> > > ```
> > 
> > 
> > Well that's the odd thing: when I had the file locally it worked fine. But until I removed it from the repo the same script failed in the workflow!
> > It makes no sense to me. It's not a huge deal but it would be nice if it could be in the test suite.
> 
> Sorry, pronouns were getting in the way of understanding the nature of the technical problem being presented.
> 
> We believe you are saying that the following test fails:
> 
> ```shell
> $ test_ioccc/txzchk_test.sh -t /usr/bin/tar -T ./txzchk -F ./test_ioccc/fnamchk -d test_ioccc/test_txzchk -Z /home/chongo/bench/ioccc/mkiocccentry
> ```
> 
> We can confirm that this is a problem that needs to be fixed.
> 
> From what we think you have said above, we need to fix this test first.

No. This was a GitHub issue. I removed the problematic file. Then the check passed. If you look at the commit here (the last one) you'll see verification worked fine. That would be commit https://github.com/ioccc-src/mkiocccentry/pull/1302/commits/0a47e084ddb67f4b7a116ee6fe45131a4a404810.

xexyl:
Anyway I gave you a command in the new issue you can try. Not sure why it would make a difference though as I rebuilt the error files ...

PR #1307, opened 2025-09-06
Fix exit codes in test_file_util.c

xexyl:
Fix exit codes in test_file_util.c and add test_file_util.8.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.6.3 2025-09-06
+
+Fix exit codes in `test_file_util.c` and add `test_file_util.8`.
+
+
 ## Release 2.6.2 2025-09-04
 
 Make adjustments to the new `test_file_util.c` so that it will function in cases
--- a/soup/version.h
+++ b/soup/version.h
@@ -84,7 +84,7 @@
  *
  * NOTE: This should match the latest Release string in CHANGES.md
  */
-#define MKIOCCCENTRY_REPO_VERSION "2.6.2 2025-09-04"	/* special release format: major.minor[.patch] YYYY-MM-DD */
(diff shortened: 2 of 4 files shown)


xexyl:
I observed that a similar problem occurs with dbg_test and dyn_array_test but I can't deal with that at this time.

Anyway now I'm just waiting on your work with the structs (no rush on my behalf .. have lots on my mind). When that is done I guess we'll have some discussion first before the chkentry and mkiocccentry code can be updated.

Best wishes. Doing other things now and probably the rest of the day.

lcn2 (approved):
Well done, thanks @xexyl

xexyl:
> Well done, thanks @xexyl 

Welcome. 

However I noticed a similar problem with dbg and dyn_array test programs. 

At the very least they don't show exit codes in the help string. 

I know I also wrote a demo for the dbg code but that's very specific and doesn't even need a help string. And the exit code is specific to and need not be documented (the entire program does not need any documentation). 

I don't know if I have the energy or inclination at this time to take care of the other repos but in any case right now I am not even able to. 

It is unclear to me how important this is when the point is that non zero means it didn't pass the tests and zero means it did but I did want to mention it in case I didn't somewhere (maybe even here .. the GitHub app makes it impossible to scroll up when posting a comment as you know). 

But as far as my energy level it's been some really hard nights and although I distinctly remember the things I did yesterday some of them feel quite distant. 

This is one of the reasons I am in no particular rush with your work on the structs. I know when you're done with them I will almost certainly have some questions (and even if it all seems clear I will still want to verify that I understand everything you need). 

After that implementation can begin. My big concern is that I think you wanted to do a freeze sometime next month. 

If a lot has to be rewritten that's going to be difficult. But given how modular I made things (for this reason exactly) I guess it won't be all that bad. 

Best wishes!

xexyl:
Ah .. I did say something here about the others. I thought I had.

PR #1312, opened 2025-09-18
Fix bug in hostchk.sh wrt dbg/dyn_array updates

xexyl:
Fix bug in test_ioccc/hostchk.sh where if you did not have dbg and dyn_array installed it would fail as it was looking for new include files. It appears that this did not happen with jparse not being installed but if anyone has an issue (I did a make uninstall but it's possible I missed something) **PLEASE** report it.

Updated HOSTCHK_VERSION to "2.0.2 2025-09-18".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,17 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.6.6 2025-09-18
+
+Fix bug in `test_ioccc/hostchk.sh` where if you did not have `dbg` and
+`dyn_array` installed it would fail as it was looking for new include files. It
+appears that this did not happen with `jparse` not being installed but if anyone
+has an issue (I did a `make uninstall` but it's possible I missed something)
+**PLEASE** report it.
+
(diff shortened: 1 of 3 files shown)


lcn2:
> > > It appears we might need to either (when running `make clobber all` or `make release` (or `make prep`) uninstall the libraries. As this is annoying perhaps the workflow needs to NOT install the libraries if it does?
> 
> > > What do you think @lcn2? Only by a happy chance did I not yet install the new changes in the server.
> 
> > 
> 
> > As other things (outside of IOCCC) use the dbg and dyn_array libraries, uninstalling the libraries is NOT an option.
> 
> > 
> 
> > We need the mkiocccentry toolkit to link to the static libraries in the `dbg/` and `dyn_array` and `jparse/` subdirectories only, independent of what may be installed on the system.
> 
> 
> 
> Oh I agree with that ... but the question is why did the workflow fail here when it does not install the libraries? I only temporarily uninstalled them (well jparse) to test. I installed it back right after.
> 
> 

Perhaps there was a delay between when the GitHub repos were updated and when the installation test occurred?

PR #1316, opened 2025-09-23
Move file_util code to pr/

xexyl:
Moved soup/file_util.[ch] out to the pr/ repo (a pull request was opened some minutes ago) as these functions are file related (hence the name :-) ) and it'll also make that library more useful (it already relied on dyn_array and dbg so this does not create new dependencies for any repo).

This commit required MANY changes as A LOT of references to file_util were in the code.

This commit removes the man pages for the old file_util_test code (under test_ioccc/) which has also been removed.

IMPORTANT NOTE: test_ioccc/ioccc_test.sh has the test_file_test changed to pr/pr_test but due to how it functions it is commented out for now (we probably need a script pr/pr_test.sh) but as make test does a make -C pr test which does run pr/pr_test this is probably okay.

Updated MKIOCCCENTRY_VERSION "2.1.3 2025-09-23".
Updated FNAMCHK_VERSION to "2.1.1 2025-09-23".
Updated SOUP_VERSION "2.0.9 2025-09-23".
Updated IOCCC_TEST_VERSION to "2.1.3 2025-09-23".

### DIFF
--- a/pr/pr.h
+++ b/pr/pr.h
@@ -102,6 +102,7 @@
 #define BITS_IN_BYTE (CHAR_BIT)	    /* assume 8 bit bytes */
 #define MAX_BYTE (0xff)		    /* maximum byte value */
 #define BYTE_VALUES (MAX_BYTE+1)    /* number of different combinations of bytes */
+#define PR_VERSION "1.1.2 2025-09-22"    /* format: major.minor YYYY-MM-DD */
 
 
 /*


xexyl:
> thank you, sorry (tm Canada 🇨🇦) it didn't work out

Not a problem. Like I said I think some of it is ambiguous (as in some functions seem out of place to me). That's kind of where this started.

That uncertainty is also one of the reasons I am not sure about WHAT to even test in the test code. As for the man page I honestly forgot about it (that it's incomplete).

I am rather pre-occupied with other things as well, some that I'm not looking forward to. I'll have to take the CPU fan out of the server and that's a pain with the clips. I'll have to put in a different fan. I hope it's just RAM that's failing but if not I might have to do a complete rebuild. Which of course costs a lot of money and I hate building computers as well. Worse is that if the OS doesn't boot I'm in trouble as annoyingly to this day anaconda does not support md+lvm in installs.

Funny that I'm looking more forward to having my eyes dilated than this. Anyway it seems like you're making progress with the tables over there. That's good. Hopefully it does not take a lot of effort to implement everything that needs to be done. 

Best wishes.

PR #1323, opened 2025-10-21
Add missing author to soup/walk*.[ch]

xexyl:
As the copyright message even says 'authors' but that was actually Landon and me. My name was missing. Not vital but still inconsistent with the rest of soup/ and the copyright message.

### DIFF
--- a/soup/walk.h
+++ b/soup/walk.h
@@ -12,7 +12,8 @@
  *		   might be one application of walking, so to is processing
  *		   the listing of a tar(1) command.
  *
- * Copyright (c) 2025 by Landon Curt Noll.   All Rights Reserved.
+ * Copyright (c) 2025 by Landon Curt Noll and Cody Boone Ferguson. All Rights
+ * Reserved.
  *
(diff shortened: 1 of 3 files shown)


xexyl:
I noticed a similar thing in `cpath` but haven't bothered to do anything (yet?).

What do you think @lcn2 about that?

xexyl:
Did the cpath updates ... question is (I haven't checked) do we have a Makefile rule (or actually do we have the Makefile rules) for syncing cpath like we do for the other repos? I'll check and add if necessary.

lcn2:
> I noticed a similar thing in `cpath` but haven't bothered to do anything (yet?).
> 
> What do you think @lcn2 about that?

That code is our code.  While it is used by this toolkit, it is not part of it.

xexyl:
> We look forward to any contributions you may have to this code as pre issue #1324 , issue #1325, and issue #1326.

I will update you tomorrow morning. 

I will also reply to the other issues. 

Thanks.

Issue #1325, opened 2025-10-21
Feature: improve txzchk to use the new table-driven walk code


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

With commit a901495217f5b8273d8b0eab080a61dc5f561e4c, "_table driven walk code_" has been added to the repo.

## Overview

The recently added "_new table driven walk code_" is found in:

- soup/walk.h
- soup/walk_tbl.c
- soup/walk_util.c

Test code that exercises the "_new table driven walk code_" is found in:

- test_ioccc/try_walk_set.c
- test_ioccc/try_fts_walk.c

The above 2 programs should be viewed as examples and demos of the "_table driven walk code_".

The actual tables in the "_table driven walk code_" are found in `soup/walk_tbl.c`.

Important and **general toolkit purpose** "_table driven walk
code_" macros, definitions, and related data structures are found in
`soup/walk.h`.

## On soup/walk.h and soup/walk_util.c

Important and ** general toolkit purpose** "_table driven walk code_"
functions are found in `soup/walk_util.c`.

We have no doubt that the "_table driven walk code_" is just a general
API, **AND** that there may be special services that this particular
tool requires.

**PLEASE**: Please keep the contents of `soup/walk.h` and
`soup/walk_util.c` for **general toolkit purpose** and **NOT** for
something specific to this particular tool.

## On soup/walk_tbl.c

The contents of `soup/walk_tbl.c` were formed on the basis of
[GH issuecomment-3240531487](https://github.com/ioccc-src/mkiocccentry/issues/1297#issuecomment-3240531487),
with a few modifications.

The contents of `soup/walk_tbl.c` represent a new **IOCCC29 policy** and
should **NOT** be changed lightly: although we acknowledge some minor
tweaking may be needed.  We **DO** acknowledge that incorporating the
contents of `soup/walk_tbl.c` in this toolkit will produce an environment
that is **DIFFERENT** from what was in place for **IOCCC28**.   Most of
those changes are **deliberate**.

If you do find a need to modify the contents of `soup/walk_tbl.c`,
beyond fixing typos and improving comments, **PLEASE** discuss the idea
in a comment before modifying where reassemble.

## Recommended actions

This tool needs to be modified so that it uses the new "_table-driven
walk code_", replacing hard-coded logic that this tool currently uses.
Each tool has its own `struct walk_rule` (and in the case of `chkentry(1)`
[...]

PR #1329, opened 2025-10-29
Make some walk table fixes for txzchk

xexyl:
Make some fixes for the txzchk walk table as well as a fix in some of the others. The dot star had "." instead of ".*". Also some of the entries had the wrong type. This solves one of the issues with updates to txzchk for issue #1325 but it is not complete yet.

Make char const **cpath_ret non-const as txzchk will need to free it.

Typo and format fixes (some) in some of the soup/walk files.

Updated SOUP_VERSION to "2.2.3 2025-10-29".

### DIFF
--- a/chksubmit.c
+++ b/chksubmit.c
@@ -296,7 +296,7 @@ main(int argc, char *argv[])
      */
     if (chkentry == NULL) {
         find_utils(NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, &found_chkentry, &chkentry,
-                NULL, NULL, NULL, NULL, NULL, NULL);
+                   NULL, NULL, NULL, NULL, NULL, NULL);
     }
 
     /*
--- a/soup/walk.h
+++ b/soup/walk.h
@@ -526,4 +526,5 @@ extern struct item *path_in_item_array(struct dyn_array *item_array, char const 
 extern struct item *path_in_walk_stat(struct walk_stat *wstat_p, char const *c_path);
 
 
+
 #endif /* INCLUDE_WALK_H */


lcn2 (commented):
The changes to `soup/version.h` has already been performed by commit d639398c0e2e6b23849a1e9e46c8878e48be392f

lcn2 (commented):
> \- 	    bool *dup_p, char const **cpath_ret)
> \+      bool *dup_p, char **cpath_ret)

See [GH pullrequestreview-3397410366](https://github.com/ioccc-src/mkiocccentry/pull/1329#pullrequestreview-3397410366).

lcn2 (approved):
as modified 👍

xexyl:
> > > > > Make `char const **cpath_ret` non-const as txzchk will need to free it.
> > > > 
> > > > 
> > > > Sorry (tm Canada 🇨🇦), while the `free(3)` function often requires one to cast when the argument is a pointer to constant characters. For example:
> > > > ```c
> > > >     free((void *) cpath_ret);
> > > > ```
> > > 
> > > 
> > > I could have sworn this was not valid. But I'll do that instead then. Not today. There are other issues that have to be answered and it's been a long day. I'll hopefully be able to do more tomorrow.
> > 
> > 
> > This is, incidentally, what I refer to. Verbatim from 6.7.3:
> > If an attempt is made to modify an object defined with a const-qualified type through use of an lvalue with non-const-qualified type, the behavior is undefined. If an attempt is made to refer to an object defined with a volatile-qualified type through use of an lvalue with non-volatile-qualified type, the behavior is undefined.
> 
> The `free(3)` function does **NOT** modify the contents of the object being freed, it modifies storage related structures in the heap.

Right. I actually knew that but as you know the longest list in the standard is the undefined list (and they seem to only add to it) so I'm always very careful. Thanks for the reminder.

xexyl:
> > > > > > Make `char const **cpath_ret` non-const as txzchk will need to free it.
> > > > > 
> > > > > 
> > > > > Sorry (tm Canada 🇨🇦), while the `free(3)` function often requires one to cast when the argument is a pointer to constant characters. For example:
> > > > > ```c
> > > > >     free((void *) cpath_ret);
> > > > > ```
> > > > 
> > > > 
> > > > I could have sworn this was not valid. But I'll do that instead then. Not today. There are other issues that have to be answered and it's been a long day. I'll hopefully be able to do more tomorrow.
> > > 
> > > 
> > > This is, incidentally, what I refer to. Verbatim from 6.7.3:
> > > If an attempt is made to modify an object defined with a const-qualified type through use of an lvalue with non-const-qualified type, the behavior is undefined. If an attempt is made to refer to an object defined with a volatile-qualified type through use of an lvalue with non-volatile-qualified type, the behavior is undefined.
> > 
> > 
> > The `free(3)` function does **NOT** modify the contents of the object being freed, it modifies storage related structures in the heap.
> 
> Right. I actually knew that but as you know the longest list in the standard is the undefined list (and they seem to only add to it) so I'm always very careful. Thanks for the reminder.

# OT comment about UB and the way the language is headed:

Actually I once counted the number of items. J.2 is obscenely long. Scrolling through the list is an entertaining exercise. It's probably longer now with C23 though that might be a cynical view. Certainly some of the things make a lot of sense but a lot of it seems to be just laziness and ironically the wording (in some things) is ambiguous/vague and seems likely to cause more UB, not less. Maybe I'm wrong.

But there are a lot of things that could easily be defined if they wanted to with no negative impact. It seems like they want to just break old programs with things like making `int main()` deprecated in all standards, removing trigraphs etc. but then add things to make it look like they're doing something constructive. Maybe they are but some of the things they're doing are also destructive.

Curiously a book that is given high praise, Head First, does not mention sequence points or UB at all. Not once from what I could tell. They also try passing fgets() off as a replacement for scanf() and being safer. They ignore the fact that you can miscalculate the size there too and they do vastly different things (they might mention that briefly). It looked like a fun book but since I saw that I decided against reading it.

Oh well. At least this issue is sorted. Will try continuing on the weekend if not tomorrow.

Issue #1324, opened 2025-10-21
Feature: improve mkiocccentry to use the new table-driven walk code


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

With commit a901495217f5b8273d8b0eab080a61dc5f561e4c, "_table driven walk code_" has been added to the repo.

## Overview

The recently added "_new table driven walk code_" is found in:

- soup/walk.h
- soup/walk_tbl.c
- soup/walk_util.c

Test code that exercises the "_new table driven walk code_" is found in:

- test_ioccc/try_walk_set.c
- test_ioccc/try_fts_walk.c

The above 2 programs should be viewed as examples and demos of the "_table driven walk code_".

The actual tables in the "_table driven walk code_" are found in `soup/walk_tbl.c`.

Important and **general toolkit purpose** "_table driven walk
code_" macros, definitions, and related data structures are found in
`soup/walk.h`.

## On soup/walk.h and soup/walk_util.c

Important and ** general toolkit purpose** "_table driven walk code_"
functions are found in `soup/walk_util.c`.

We have no doubt that the "_table driven walk code_" is just a general
API, **AND** that there may be special services that this particular
tool requires.

**PLEASE**: Please keep the contents of `soup/walk.h` and
`soup/walk_util.c` for **general toolkit purpose** and **NOT** for
something specific to this particular tool.

## On soup/walk_tbl.c

The contents of `soup/walk_tbl.c` were formed on the basis of
[GH issuecomment-3240531487](https://github.com/ioccc-src/mkiocccentry/issues/1297#issuecomment-3240531487),
with a few modifications.

The contents of `soup/walk_tbl.c` represent a new **IOCCC29 policy** and
should **NOT** be changed lightly: although we acknowledge some minor
tweaking may be needed.  We **DO** acknowledge that incorporating the
contents of `soup/walk_tbl.c` in this toolkit will produce an environment
that is **DIFFERENT** from what was in place for **IOCCC28**.   Most of
those changes are **deliberate**.

If you do find a need to modify the contents of `soup/walk_tbl.c`,
beyond fixing typos and improving comments, **PLEASE** discuss the idea
in a comment before modifying where reassemble.

## Recommended actions

This tool needs to be modified so that it uses the new "_table-driven
walk code_", replacing hard-coded logic that this tool currently uses.
Each tool has its own `struct walk_rule` (and in the case of `chkentry(1)`
[...]

PR #1335, opened 2025-11-02
Address issue #1324

xexyl:
The mkiocccentry tool now uses the walk table code. Some of the struct info was nuked. There might be some room for clean up with this but this can be considered a checkpoint at the least in that case.

Updated MKIOCCCENTRY_VERSION to "2.1.6 2025-11-02".


xexyl:
I know there is some room for clean up but I need to go and I figured getting this important commit in would be better than to hold off. The test script works just fine. I forgot to note I had to remove the -I and -M options and I did not update the man page (again I have to go). I think it's unfortunate but I also guess that few people would use those options. Still maybe worth looking at for IOCCC30?

I'll do some clean up tomorrow on this.

lcn2:
> I know there is some room for clean up but I need to go and I figured getting this important commit in would be better than to hold off. The test script works just fine. I forgot to note I had to remove the -I and -M options and I did not update the man page (again I have to go). I think it's unfortunate but I also guess that few people would use those options. Still maybe worth looking at for IOCCC30?
> 
> I'll do some clean up tomorrow on this.

Both the `-M manifest` and the `-X ignore` flags are supported with only a small amount of effort.  The "table-driven walk code" was designed with the `-M manifest` and the `-X ignore` flags in mind.

See [GH issuecomment-3478292323](https://github.com/ioccc-src/mkiocccentry/issues/1324#issuecomment-3478292323) for details.

xexyl:
As for anything goes in your commit log I am also reminded of the first scene in Indiana Jones and the Temple of Doom. 

I will be looking more at this soon I hope.

lcn2:
> I really appreciate your looking into this whilst I was gone! It looks like I won't be able to put in the missing options unless I can somehow sync this branch to my copy of the branch. Which I suppose should be possible but there might be a risk of conflicts. If nothing else I can at least write the code so that when you're done I can make another pull request.

You are most welcome, @xexyl 

Please review [GH issuecomment-3478356241](https://github.com/ioccc-src/mkiocccentry/pull/1335#issuecomment-3478356241).

Please add the code for the `-M manifest` and `-X ignore` as per [GH issuecomment-3478292323](https://github.com/ioccc-src/mkiocccentry/issues/1324#issuecomment-3478292323).

PR #1340, opened 2025-11-04
Fixes and modularisation for issue #1324

xexyl:
The scanning of the topdir must NOT check for forbidden files (of all kinds). Instead of calling copy_topdir() directly scan_topdir() now calls show_copy_list() (new function) which then calls copy_topdir() which then calls check_submission_dir(). When copying files in copy_topdir() we must skip the list of pruned files. When showing (in the new function show_copy_list()) the list of files and directories to be ignored and copied skip files that are in the prune list. When copying files and making directories in the submission directory skip files and directories in the prune list. When checking files and directories in check_submission_dir() don't check the length of the arrays; instead just check the submission directory arrays. This is because the topdir CAN (depending on the topdir in question) have a different number of files and directories. This happens because files and directories that are in submission directory can be (and are in the case pruned files and/or directories were in the topdir) a different number. The user will have to instead verify the submission directory but given that the user is given multiple chances this should be fine.

Minor format fixes in txzchk.c.

Typo fix in soup/walk_util.c.

Updated MKIOCCCENTRY_VERSION to "2.1.9 2025-11-03".

### DIFF
--- a/mkiocccentry.c
+++ b/mkiocccentry.c
@@ -1588,7 +1588,7 @@ show_copy_list(struct walk_stat *wstat, char const *context, struct info *infop,
      */
     if (wstat == NULL || context == NULL || infop == NULL || make == NULL || submission_dir == NULL ||
 	submit_path == NULL || size == NULL || wstat2 == NULL) {
-        err(48, __func__, "passed NULL pointer(s)");
+        err(52, __func__, "passed NULL pointer(s)");
         not_reached();
     }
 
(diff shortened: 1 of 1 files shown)


xexyl:
Damn ... wonder what failed ... have to check. Might or might not be able to fix right now.

xexyl:
Ah .. now I know why - you did run into problems too. Hopefully I can remember how I fixed it last time and it will work. At least this fix is in though. I have to leave for a bit but when back I'll see if I can work this out.

PR #1341, opened 2025-11-04
Fix semantics errors from commit 63a719a

xexyl:
It should be noted that now that mkiocccentry uses the table code and 
simply skips over the files that are too long in name or for other 
reasons, the mkiocccentry_test.sh script fails as the bad tests no 
longer fail (or some don't if not all). A new issue will be opened to 
discuss this issue. The chksubmit_test.sh has an issue too or so it 
seems; testing it manually here it passed but it seemed like it failed 
in make test. This will be mentioned in the new issue that will be 
opened but the semantics errors should be fixed at least.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,6 +55,9 @@ Typo fix in `soup/walk_util.c`.
 
 Updated `MKIOCCCENTRY_VERSION` to `"2.1.9 2025-11-03"`.
 
+Fix semantics errors from the `.info.json` file updates.
+
+
 ## Release 2.8.8 2025-11-03
 
 Further address #1324. The `-M manifest` and `-X ignore` options were added
--- a/soup/chk.auth.ptch.c
+++ b/soup/chk.auth.ptch.c
@@ -1,5 +1,5 @@
---- ref/auth.reference.json.c	2025-11-03 20:25:40
-+++ chk_sem_auth.c	2025-11-03 20:05:33
+--- ref/auth.reference.json.c	2025-11-04 05:14:32
++++ chk_sem_auth.c	2025-11-04 05:13:30
 @@ -40,9 +40,9 @@
  struct json_sem sem_auth[SEM_AUTH_LEN+1] = {
  /* depth    type        min     max   count   index  name_len validate  name    data */
--- a/soup/chk.auth.ptch.h
+++ b/soup/chk.auth.ptch.h
@@ -1,5 +1,5 @@
---- ref/auth.reference.json.h	2025-11-03 20:25:40
-+++ chk_sem_auth.h	2025-11-03 20:05:32
+--- ref/auth.reference.json.h	2025-11-04 05:14:33
++++ chk_sem_auth.h	2025-11-04 05:12:36
 @@ -33,19 +33,17 @@
  /*
   * json_sem - JSON semantics support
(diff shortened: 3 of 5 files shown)


xexyl:
> > And I really should have documented how I fixed the semantics issues. It's such a pain and I hope we can get a better system in in time.
> 
> Please do document.
> 


I should have said... I am not entirely sure what I did. I can say there were a few things wrong though. One was  you accidentally had the wrong function names in the JSON C files (the ones that end up in soup/ref).

I used `sgit(1)` to make the corrections there. I THINK I then did:

```sh
make all_sem_ref
make all_sem_ref_ptch
```

but I am not 100% sure of this. 

After the other issues are resolved if I have the energy and patience (I'm sure you understand) I could temporarily checkout the commit I had to fix and then try it again, this time documenting it. If I am able to do that would you like that?

On the other hand ...

> The current system of semantics is a failure and needs to be replaced. That is for a post **IOCCC29** update as we don't want to have a late start on **IOCCC29**. Nevertheless, please do document what you did.

Agreed. Since it does seem it'd be useful I'll see if I can figure out what I did after the other issues are resolved: though perhaps not the same day.

Question on this for you @lcn2: where do you want this documented?

Also I FINALLY just realised something. This entire time (in fact when you first told me long before anyone else) about IOCCC29 opening in December and it being once a year I could not figure out how this worked. But now I think on it the contest went into pending on 31 December last year. So it really will have been about a year! And this year in some ways seems to have gone by very fast. In other ways very slow (from a not perfect typo wise copy .. too lazy to pull it out but it looks right and I've read it so many times that I know it EXTREMELY well: so minus the poor sleep ...):

<hr>
'Nay, time does not tarry ever,' he said; 'but change and growth is not in all things and places alike. For the Elves the world moves, and it moves both very swift and very slow. Swift, because they themselves change little, and all else fleets by: it is a grief to them. Slow, because they do not count the running years, not for themselves. The passing seasons are but ripples ever repeated in the long long stream. Yet beneath the Sun all things must wear to an end at last.'
<hr>

Actually like a lot of other things that has applied to my life in many ways.

But that's all another matter entirely and is just a symptom of how tired I have been.

Best wishes! I'm going to look at those issues now (I hope). Hopefully they'll be easy to deal with and then the next time I can look at documenting fixing semantics tables, or at least what I did: adding to it I don't even remember unless we have to add to the files manually? Is that what you did?

lcn2:
> Question on this for you @lcn2: where do you want this documented?

How about `soup/semantics.how-to.md` or something like that?

xexyl:
> > Also I FINALLY just realised something. This entire time (in fact when you first told me long before anyone else) about IOCCC29 opening in December and it being once a year I could not figure out how this worked. But now I think on it the contest went into pending on 31 December last year. So it really will have been about a year! And this year in some ways seems to have gone by very fast. In other ways very slow (from a not perfect typo wise copy .. too lazy to pull it out but it looks right and I've read it so many times that I know it EXTREMELY well: so minus the poor sleep ...):
> 
> Correct. Moreover we see it as **EXTREMELY** critical that **IOCCC29** start in 2025. It will be a clean sign that all of the **HARD WORD** you (@xexyl) and others put in to rebuild the way IOCCC is managed has paid off: namely that the IOCCC has returned to an annual contest.

Oh I put in lots of hard words :-)

(But on a more serious note: you're very welcome and on my behalf I can't see how it won't start in this year.)

PR #1344, opened 2025-11-04
Updated FAQ to be more up to date

xexyl:
Remove references to chkentry (use chksubmit instead) and add references
to location (as that can be useful for those who don't know their
country code, even though we provide a URL of a list as well).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,10 @@ Updated `MKIOCCCENTRY_VERSION` to `"2.1.9 2025-11-03"`.
 
 Fix semantics errors from the `.info.json` file updates.
 
+Updated FAQ.md - remove references to chkentry (use chksubmit instead) and add
+references to location (as that can be useful for those who don't know their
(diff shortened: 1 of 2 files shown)


xexyl:
This will of course fail due to the fixes that cause the scripts to fail ... but none of that should prevent a merge being possible.

When you have figured out what you want those scripts to do instead I can hopefully look at them (or else if you have ideas you can). Leaving shortly.

lcn2 (approved):
Thank you @xexyl 

Due to physical issues and local matters helping others who need our physical assistance where we live, we had to "dump what we had been doing into a checkpoint" without a high level of testing.  From the comments we read, we see that there were issues beyond our original commit that need to be addressed: not surprising since we saw some pending issues ourselves when we were doing testing.

It is our hope to return to this repo towards the end of this week.  In the meantime we hope you can make progress on matters to the nest that you can using your best judgment. 

Thank you for helping and thank you in advance for whatever help you can provide over the next few (2?) days.

Issue #1345, opened 2025-11-05
Bug: fts_walk() does not like getting a path which ends in `/`


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

If one does:

```sh
$ ./chksubmit test_ioccc/workdir/good/test-0
$
```

then all is good (as the prompt shows). If however one adds a trailing `/`:

```sh
$ ./chksubmit test_ioccc/workdir/good/test-0/
Warning: fts_walk: fts_read() produced fts_pathlen: 41 == 31 for non-wstat_p->topdir: test_ioccc/workdir/good/test-0/.auth.json
ERROR[1]: chkentry: failed to scan: test_ioccc/workdir/good/test-0/
$
```

... all is not good, as can be seen. This is the code:

```c
        /*
         * paranoia - avoid fts_path that does NOT start with canonicalized wstat_p->topdir followed by / (slash)
         *
         * NOTE: This should never happen, but in case there is an fts_read(3) bug, we check.
         */
        } else if (ftsent->fts_path[wstat_p->topdir_len] != '/' ||
                   strncmp(wstat_p->topdir, ftsent->fts_path, wstat_p->topdir_len) != 0) {

            /* warn about fts_path not starting with canonicalized wstat_p->topdir + / (slash) */
            warn(__func__, "fts_read() produced fts_pathlen: %d == %zu for non-wstat_p->topdir: %s",
                 ftsent->fts_pathlen, wstat_p->topdir_len, ftsent->fts_path);
    
```

### What you expect

It probably would be better if the trailing `/` (and even multiple `/`s as in `///`) was ignored as sometimes people do add a trailing `/` with directory names. The error might make some people think something is wrong, more wrong than it actually is.

### Environment

- OS: n/a
- Device: n/a
- Compiler: n/a

### bug_report.sh output

n/a

### Anything else?

It is not entirely clear to me what the best approach is but that would also depend on if we deal with multiple trailing slashes or not. To me it seems like that should be done as it would be confusing to some people.

xexyl:
I have fixed t his .. it involved some changes in chkentry (need absolute path) as well as some changes in the walk code .. though after the change in the chkentry it might not be required. Nonetheless the check is obviously iffy. I noticed a missing check too so I added that. I'll do a commit soon ...anyway check this:

```sh
$ ./chkentry test_ioccc/workdir/bad/test-5/
chkentry -S required: Makefile not found
[...]

PR #1349, opened 2025-11-07
Address #1345 - fts_walk() with dir ending with /

xexyl:
Fixed fts_walk() and chkentry(1) to work with a directory ending with /
AND a directory NOT ending with a /.

Updated CHKENTRY_VERSION to "2.1.9 2025-11-07"
Updated SOUP_VERSION to "2.2.10 2025-11-07".

### DIFF
--- a/chkentry.c
+++ b/chkentry.c
@@ -236,9 +236,10 @@ main(int argc, char *argv[])
             errp(44, __func__, "failed to chdir() into submission dir: %s", argv[optind]);
             not_reached();
         }
+        errno = 0; /* pre-clear errno for errp() */
         submission_dir = getcwd(NULL, 0);
         if (submission_dir == NULL) {
-            err(44, __func__, "failed to get absolute path for: %s", argv[optind]);
+            errp(45, __func__, "failed to get absolute path for: %s", argv[optind]);
             not_reached();
         }
 	break;


xexyl:
Incidentally @lcn2 ...

Do we want to construct more than one bad directory for the test script? And should we do like what I did for the txzchk test script - have error files that have to match? I'm mixed on the latter one (maybe partly as it might not be worth it) but the former might be an idea IF it can be contrived easily enough. It could even be that the chksubmit script does it itself (all of them) so that mkiocccentry_test.sh does not have to be run prior to the chksubmit_test.sh script.

What do you think about these issues ?

lcn2 (approved):
We went offsite to find an access point, and using a VPN we inspected this PR by eye (no, not that eye 👁️) and said aye, @xexyl 😉 and thanks 🙏.  

Likely the network resolution may take a few days+ so please 🙏 be patient while we are infrequently online.

lcn2:
> Incidentally @lcn2 ...
> 
> 
> 
> Do we want to construct more than one bad directory for the test script? And should we do like what I did for the txzchk test script - have error files that have to match? I'm mixed on the latter one (maybe partly as it might not be worth it) but the former might be an idea IF it can be contrived easily enough. It could even be that the chksubmit script does it itself (all of them) so that mkiocccentry_test.sh does not have to be run prior to the chksubmit_test.sh script.
> 
> 
> 
> What do you think about these issues ? 

They seem like good 👍 ideas, @xexyl .. please proceed.

xexyl:
> > Incidentally @lcn2 ...
> > Do we want to construct more than one bad directory for the test script? And should we do like what I did for the txzchk test script - have error files that have to match? I'm mixed on the latter one (maybe partly as it might not be worth it) but the former might be an idea IF it can be contrived easily enough. It could even be that the chksubmit script does it itself (all of them) so that mkiocccentry_test.sh does not have to be run prior to the chksubmit_test.sh script.
> > What do you think about these issues ?
> 
> They seem like good 👍 ideas, @xexyl .. please proceed.

Will try doing that in the next few days .. hope you went to sleep after you merged!

xexyl:
> We went offsite to find an access point, and using a VPN we inspected this PR by eye (no, not that eye 👁️) and said aye, @xexyl 😉 and thanks 🙏.
> 

An aye for an aye and a nay for a nay?

Anyway you're welcome.

> Likely the network resolution may take a few days+ so please 🙏 be patient while we are infrequently online.

No worries at all of course! Thanks for the note however. Good luck with it! I know how frustrating a network being down can be. It happened when I didn't have the static IP allocated with the new service (which is AMAZING!) earlier this year.

xexyl:
Oh damn ... I found a possible NULL pointer dereference .. good thing is the script was updated to create bad directories and the chksubmit test script works.

I might have to stop before I try tracking down the possible NULL dereference.

xexyl:
Ah .. I found it already. That is quite unfortunate but I think I can work past it ... stay tuned ...

xexyl:
Fixed. Will update changelog now then commit ..

PR #1351, opened 2025-11-10
Silence assorted compiler warnings on NetBSD gcc 14.3.0.

SirWumpus:
With the approaching code freeze I got tired of sifting through unnecessary warnings concerning `stderr` and format string types.

Test suite passes clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ C_OPT= -O3
 
 # Compiler warnings
 #
-WARN_FLAGS= -Wall -Wextra -Wno-char-subscripts
+WARN_FLAGS= -Wall -Wextra -Wno-char-subscripts -Wno-address
 #WARN_FLAGS= -Wall -Wextra -Werror
 
 # special compiler flags
--- a/jparse/jparse.l
+++ b/jparse/jparse.l
@@ -226,7 +226,7 @@ JSON_COMMA		","
                              * (depending on if the JSON debug level was high
                              * enough).
 			     */
-			    (void) json_dbg(JSON_DBG_VVHIGH, __func__, "\nignoring %d whitespace%s\n",
+			    (void) json_dbg(JSON_DBG_VVHIGH, __func__, "\nignoring %zu whitespace%s\n",
 							     yyleng, yyleng==1?"":"s");
 			}
 
--- a/jparse/jparse.y
+++ b/jparse/jparse.y
@@ -427,7 +427,7 @@ json_value:
 	    json_dbg(JSON_DBG_VHIGH, __func__, "under json_value: starting: "
 					       "json_value: JSON_TRUE");
 	    json_dbg(JSON_DBG_VVHIGH, __func__, "under json_value: yytext: <%s>", jparse_get_text(scanner));
-	    json_dbg(JSON_DBG_VVHIGH, __func__, "under json_value: yyleng: <%d>", jparse_get_leng(scanner));
+	    json_dbg(JSON_DBG_VVHIGH, __func__, "under json_value: yyleng: <%zu>", jparse_get_leng(scanner));
 	    json_dbg(JSON_DBG_VHIGH, __func__, "under json_value: about to perform: "
 					       "$json_value = json_parse_bool(jparse_get_text(scanner));");
 	}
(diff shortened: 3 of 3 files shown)


xexyl:
No this can't work as is. 

First jparse itself has to be updated for this and secondly yyleng is not unsigned. It is an int.  

I can look at it soon.

xexyl:
.. but any fix would conflict with this pull request. 

You might want to close this one so I can do it properly. 

It would help if you would tell me why you think you have to change the format specifier for yyleng though because `%d` is actually correct. 

Maybe changing that Makefile will help for non conforming systems (if that's what is happening).

SirWumpus:
NetBSD reports it as a size, so I treated it as such.

I figure there would be some cross-over issues.  I wasn't keen on chasing across repos.

xexyl:
> NetBSD reports it as a size, so I treated it as such.
> 
> 

Well it's wrong to do that. Better to silence the warning for systems like that. 


> 
> I figure there would be some cross-over issues.  I wasn't keen on chasing across repos.

The code here has to be synced from over there and the commit doesn't complete it either. 

Please tell me which warning is triggered and I will take care of it. 

Thanks.

SirWumpus:
[bug-report.20251110.051019.txt](https://github.com/user-attachments/files/23451163/bug-report.20251110.051019.txt)

Here is the compiler output before the PR.  You can see the jparse ~~errors~~ warnings and figure the best solution.

xexyl:
.. honestly a pain though. There are other places that function call is printed so I have to go through and find them.

SirWumpus:
Look at the output, maybe `long unsigned` is the type, in which case `%lu`.  I wouldn't go fixing with casts, messy.  Maybe if you are passing different variables for the same message in assorted places, you might need two different messages with the correct format string?

xexyl:
BTW:

```c
    int yyleng_r;
```

is in the struct so I'm not sure why it thinks it's a size_t ... anyway updating version and changelog over there and then I'll sync to here (after rebuilding the files). I can silence the other warning you're worried about here too.

xexyl:
Also as for casting in one place we actually have to because the function we pass it to expects a size_t - so casts already exists.

SirWumpus:
Its `yyleng_r` doesn't appear to be the name, `yyleng`.  Which header file for the yacc stuff and I'll check here.

xexyl:
> Ok...
> 
> ```
> protected:
>   char* yytext;
>   int yyleng;
>   int yylineno;       // only maintained if you use %option yylineno
>   int yy_flex_debug;  // only has effect with -d or "%option debug"
> };
> ```
> 
> So I don't understand why the warnings.

Exactly. NetBSD is broken. Hence the casts. I pushed it .. will sync here.

xexyl:
Checking the file .. just saw it. Thanks! I guess if it's just changing the data type that a cast will work. I have to take a brief moment to do something else though.

SirWumpus:
@xexyl is providing a revised PR in place of this one.

xexyl:
> What `-d`? There is `-Dmacro`. `gcc -dM is present.` For example I have in some projects, where CC=gcc
> 
> ```
> predefines:
>         ${CC} ${CPPFLAGS} -dM -E -xc /dev/null
> ```

Well I was pasting from what you pasted. However I do see that the above is slightly different from what `cpp -dM` shows. In particular:

```diff
345d344
< #define __STDC__ 1
```

It seems like the one you gave is even better. That was from macOS btw. I haven't tested it in linux. I can update the script but then I will have to probably do it in the jparse repo too. Now I can do this before the contest opens (I'd rather not but I can) or because it's probably not important if you wanted to you could open an issue for post IOCCC29?

Maybe @lcn2 will have some ideas too though his service is out right now ...

xexyl:
Okay doing it now to get it done for you .. then hopefully this will be the last thing prior to code freeze. Thank you!

SirWumpus:
> If it did sound grouchy it wasn't meant that way.

I just assume grouchy by default; just easier that way 😁

For me I was up at 4h00 and might have been fuzzy reading some of the jparse warnings.

SirWumpus:
> Oh wow ... changing it to not use cpp made the workflow fail. So maybe it's only in BSD systems it works? Damn.

No.  I've been using that with gcc across platforms for decades.  I'll kick up Rocky and check.

PR #1352, opened 2025-11-10
Fixes for NetBSD (jparse mostly)

xexyl:
Sync jparse/ from jparse repo (https://github.com/xexyl/jparse). This should hopefully fix the problem with NetBSD warning about yyleng being unsigned (even though it seems to be an int there). The jparse_bug_report.sh to not rely on cpp to get predefined macros.

Added also to the Makefile -Wno-address to prevent some compilers from whining about our checks for NULL on stderr (it can become NULL in some conditions and it's better make sure of it before trying to dereference it).

Fixed bug_report.sh to not rely on cpp for getting the predefined macros.

Updated BUG_REPORT_VERSION to "1.0.13 2025-11-11".

### DIFF
--- a/jparse/CHANGES.md
+++ b/jparse/CHANGES.md
@@ -1,6 +1,6 @@
 # Significant changes in the JSON parser repo
 
-## Release 2.5.3 2025-11-11
+## Release 2.5.3 2025-11-10
 
 Cast `yyleng` (and `jparse_get_leng()`) to an int for systems like NetBSD that
 try making it an unsigned int despite the fact the struct has it as an `int`.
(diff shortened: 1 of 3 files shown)


xexyl:
Please @SirWumpus test this and that includes the bug report script update. Clever trick for the cpp workaround. The `-x` option of `cc` is certainly very useful! I have used it numerous times and I think one of the fixes for one of the entries that no longer worked right I had that .. and maybe some of my previous entries though I have no idea on that now.

lcn2 (changes requested):
We are viewing this via a satellite link and CANNOT download and compile to test it.

Unfortunately this PR seems incomplete as it only makes changes to version strings.
It does NOT contain any of the recent changes to the "other repo" such as `(int)yyleng` stuff.

Please update this PR, @xexyl, and then run `make release` to verify.

Thanks.

PR #1364, opened 2025-11-19
Fix additional critical validation errors

xexyl:
The following validation errors were fixed:

- if a filename ends in .sh then it MUST be a shell_script.
- an extra_file many not match a shell_script filename.
- a shell_script file must not match another shell_script filename.
- an extra_file filename must NOT end in .sh.
- add check for too many files with just shell_script list.
- add check for too many files with shell_script and extra_file
combined.

The above fixes required a new array in the struct manifest - one for shell
script files.

Fix typos.


lcn2 (approved):
Thanks 🙏 @xexyl

Issue #1380, opened 2026-03-12
Bug: answer file response to Previous Winner always forces "winner"


SirWumpus:
An unconfirmed bug from [csmoak on Discord](https://discord.com/channels/1327029557718417438/1332205820925972541/1481591367648936028):

> when i saved answers to a file after saying i'm not a previous winner and then used that answers file when continuing to build a submission, it reported i claim to be a past winner

Is this related to #1375 ?

xexyl:
Possibly not related. But I haven't looked at that one in a while.  I think not unless their answers file got messed up. That actually seems possible now I think on it but the question would be why?

Did they actually complete the first packaging?

It seems odd this would be the case though, unless the file was written in part (which I didn't think is the case). Is it possible for them to comment here? I have some other things I wish to fix after the contest closes and I could look at this as well. The easiest way to test would be (not now) to attempt this but with more feedback it'd be easier.

xexyl:
It can't always force it because we have tests that do otherwise.

xexyl:
I'd like to see their answers file. That would help.

SirWumpus:
I've supplied the link and asked them to follow-up here, but you might have to follow on Discord just in case.

xexyl:
> I've supplied the link and asked them to follow-up here, but you might have to follow on Discord just in case.

I saw but that might be an issue esp if they can't attach the file.

cesmoak:
Howdy. I'm attaching a redacted version of the answers file here. I believe I exited just before it wrote the submit .txz by pressing N at one of the final choices

[answers.txt](https://github.com/user-attachments/files/25932043/answers.txt)

xexyl:
Ugh.

Something **might've** recently change. Looking at the test script it seems all are claiming to be past winners.

My guess is when we went to table driven mode or code about that time did it.

But it might be that that was in the test script already. Please confirm by doing it again but completing the entire process.

I'll look at it after the contest closes. Thanks and apologies if it is a bug. I unfortunately don't have time right now to do a test.

xexyl:
Oh bugger me. It looks like this is not being written to the answers file. The amazing thing is it even works.

I'll fix after the contest closes. Sorry .. not even sure when it happened :(

xexyl:
I can't get my head round this one.

The answers file USED to have 'y' or 'n' added to it for several lines.

But now the tool does not do that. I guess it happened when the seeding option was introduced but I do not know.

[...]

PR #1382, opened 2026-03-14
Fix issue #1380

xexyl:
Fix issue #1380 - except the mkiocccentry_test.sh script as that has to do with the slots. There were two things that had to be done to get this to work (though I recommend more testing than what I did, even though I did try various combinations) is the past_winning_author has to be written to the answers file and in yes_or_no() the check that returns true if answer_yes or force_yes are true now only returns true if !read_answers_flag_used. This way the answers file controls the yes or no.

The mkiocccentry_test.sh fails which means the workflow in GitHub will fail! I wanted this in though as it should fix mkiocccentry itself. As for the test script I BELIEVE it is due to the slots. There is an answers file read error which might make sense because there is more data in it now (or is supposed to be more data). A quick attempt at changing the slots file did not work (though the mkiocccentry_slots.sh exited 0) but as this is a good point to commit I am doing that as the test script can be fixed in another commit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,9 @@ answers file controls the yes or no.
 To fix the test script the good slots answers files had to be updated and
 `mkiocccentry_test.sh` had to be updated too (answers function).
 
+Updated `MKIOCCCENTRY_VERSION` to `"2.3.0 2026-03-14"`.
+Updated `MKIOCCCENTRY_TEST_VERSION` to `"2.1.3 2026-03-14"`.
+
 
 ## Release 2.11.0 2025-11-30
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,25 @@
 # Major changes to the IOCCC entry toolkit
 
 
+## Release 2.11.1 2026-03-14
+
+Fix issue #1380 - except the `mkiocccentry_test.sh` script as that has to do
+with the slots. There were two things that had to be done to get this to work
+(though I recommend more testing than what I did, even though I did try various
+combinations) is the `past_winning_author` has to be written to the answers file
+and in `yes_or_no()` the check that returns true if `answer_yes` or `force_yes`
+are true now only returns true if `!read_answers_flag_used`. This way the
+answers file controls the yes or no.
+
+The `mkiocccentry_test.sh` fails which means the workflow in GitHub will fail! I
+wanted this in though as it should fix mkiocccentry itself. As for the test
+script I BELIEVE it is due to the slots. There is an answers file read error
+which might make sense because there is more data in it now (or is supposed to
+be more data). A quick attempt at changing the slots file did not work (though
+the `mkiocccentry_slots.sh` exited 0) but as this is a good point to commit I am
+doing that as the test script can be fixed in another commit.
+
+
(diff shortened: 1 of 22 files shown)


xexyl:
The workflow WILL fail. Maybe you can address is @lcn2 when you're back from judging? Assuming I've not fixed it yet of course. I'll maybe do that but not now .. and perhaps not today even or even for a while.

xexyl:
Oh I'm a bloody idiot .. I forgot that I also had to update the mkiocccentry_test.sh script. Let's see if this works. It works locally.

lcn2 (changes requested):
There are merge conflicts with this PR, and as such this PR needs to be modified.

xexyl:
I am sorry but I can't deal with this right now .. I thought I could but I'm unable to right now. Same goes for the others. I'm also not sure if I got the one (that I 'merged') right. The problem is I'm not sure what the versions should be.

Given this can wait it does not matter much (even if it would be nice if it were resolved).

xexyl:
I believe that I fixed the pull requests. However .. an entry was lost in the CHANGES.md file. That change was the following, in markdown here so you can copy paste in the right location (alternatively I can):

```
## Release 2.11.1 2026-05-06

Update as per [jparse repo](https://github.com/xexyl/jparse) 2.5.8 2026-05-06.

This is an internal process update to simply sync with that repo.

Changed `MKIOCCCENTRY_REPO_VERSION` to "2.11.1 2026-05-06".
Changed `MKIOCCCENTRY_VERSION` to "2.3.1 2026-05-06".

Kept `MIN_MKIOCCCENTRY_VERSION` at "2.3.0 2015-11-30" because while
`make release` passed, this release while nice, isn't critical.

```

Note however that with the `-funsigned-char` commit the `jparse` was further updated.

xexyl:
See also [comment 4582728423](https://github.com/ioccc-src/mkiocccentry/issues/723#issuecomment-4582728423).

lcn2 (approved):
thank you @xexyl 👍

This PR will be processed after the winning entries of **IOCCC29** have been published.

lcn2:
# Schedule notice

About two or three weeks after the **IOCCC29** winning entries
have been posted, and we process some of the early
[pull requests](https://github.com/ioccc-src/winner/pulls)
against the **2025 directory tree**, the
[IOCCC Judges](https://www.ioccc.org/judges.html) plan to go on an **IOCCC vacation**.

We had intended to go on an **IOCCC vacation** after releasing the winners of
**IOCCC28**, but then the efforts to process bug fixes and enhancements
to the [mkiocccentry repo](https://github.com/ioccc-src/mkiocccentry)
took so much time that by the time that repo was stable, it was time to
open **IOCCC29**.  Therefore, this time, we plan to wait until after
the end of our post-**IOCCC29** **IOCCC vacation** before working on any
[mkiocccentry repo PRs](https://github.com/ioccc-src/mkiocccentry/pulls).

Folks will have time to refine any PRs and/or to submit new PRs against the
[mkiocccentry repo](https://github.com/ioccc-src/mkiocccentry).  We ask for your patience in advance.

lcn2 (approved):
👍 and thank you @xexyl

Issue #1389, opened 2026-06-03
Bug: the total file size check does not work right (see below for details)


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

The following code shows the problem:

```c
                if (total_file_size > MAX_SUM_FILELEN) {
                    /*
                     * We probably should check >= because we have to create the
                     * JSON files as well. But since those are an indeterminate
                     * size but are also > 1 byte it doesn't really matter.
                     */
                    print("\nWARNING: the %lld estimate exceeds the Rule 17 limit of %d.\n",
                            (long long)total_file_size, MAX_SUM_FILELEN);
                }

```

Obviously that comment doesn't make that much sense but part of that is because the value being checked against is the max size itself and not that value - some amount.

However this code is more buggy than it seems. See below.

### What you expect

It should maybe warn if it's approaching it instead of over it? Again see below.

### Environment

- OS: n/a
- Device: n/a
- Compiler: n/a

### bug_report.sh output

n/a

### Anything else?

The funny thing is, and this makes me think we should not even bother with this 'helpful' thing, is that the above code is the only place that macro (the max one) is tested. Later on we test the tarball length. But we cannot know the tarball length. Okay there is one thing ... see below (last part of this section).

There's another thing. By waiting until the end the answers file is completed. This means that the only thing that would have to be fixed is the directory they're submitting. In the end why have two places where it warns where one is just a warning and also which is not even the value tested in the end?

I know it was suggested (not by me) that this code might be useful to users but given we can't even determine the value until after the tarball is formed and given that the JSON files might be a different size, what good does this do? As noted above at least if they go the whole way they can create the answers file! That means later they just have to rerun the command and not worry about entering the contest UUID and everything else (assuming they use the answers file - but some certainly do).

I suggest that this is simply removed. I will do that (so feel free to assign it to me) - if this is decided on - after the current pull requests are merged. If this is not desired however we need to come up with something that actually works. Right now this does not work at all.

I did say that there is no check on that. This is not strictly true! It is used in txzchk of course. But in mkiocccentry it is not used and we still can't know how big the file will be. And again we would then force the users (who use an answer file) to re-enter the information they already entered. That is actually less user-friendly.

lcn2:
> I suggest that this is simply removed. I will do that (so feel free to assign it to me) - if this is decided on - after the current pull requests are merged. 

Please 🙏 remove as suggested.

[...]

Issue #1379, opened 2026-03-04
Enhancement: Some people are confused by the term "IOCCC contest ID"


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

Some people are confused when `mkiocccentry` prompts:

```
Enter IOCCC contest ID or test: IOCCC29
IOCCC contest IDs are in the form:

    xxxxxxxx-xxxx-4xxx-Nxxx-xxxxxxxxxxxx

where 'x' is a hex character, 4 is the UUID version, and N is one of 8, 9, a, or b.
```

Perhaps the tool should just ask for "IOCCC submit server UUID"?

### Relevant links

See:

https://discord.com/channels/1327029557718417438/1332198374249594920/1478558841724469380

xexyl:
Yup. That would simplify it and it would make it easier for us. 

And also we MIGHT not even need to say the test mode because that's for us and the test suite. Also there are the options we added for testing a submission without having to answer any questions. 

But it probably (?) wouldn't hurt to have the 'or "test"'. 

Feel free to assign this to me if you want although I guess it won't matter until next time?

xexyl:
And have a good night my friend! I am just getting up. My sleep is entirely upside down. 

My surgery date has also been changed but not sure when it will be yet. Hopefully not later than it already was going to be. 

I also have a question for you and perhaps Leo about something that I will get to by email eventually. To do with 2024/ferguson1 actually. 

Anyway sleep well!

xexyl:
Well @lcn2 I am looking at this but I'm not sure exactly what it should say. Or maybe it's how it should be presented.

We definitely want to still show the test option. The question is do we change the entire text to just say something like:

> Enter IOCCC submission server UUID or 'test':

... or if you prefer double quotes we could do that.

But it could be something else too. There are many simple options.

> Enter your IOCCC submission server UUID or 'test':

... that might be a bit clearer?

If we wanted to be a bit more verbose:

> Enter the UUID you received in the welcome email or 'test':

... and so on. The last one is not quite right either. I would have to look at exactly what it's called but that's the idea. The idea of the last one is that it is more explicit of where they get it whereas the others don't clarify where it is from which I believe might actually cause a different kind of confusion.

Perhaps we should ask Dave since he brought it up (which kind of surprised me as he submitted and won last time)? Still it would be good if it was simplified.

I know you're going to be inactive for a while but I had a chance to actually address this now.

[...]

Issue #1373, opened 2025-12-22
Enhancement: ALL tools (except perhaps iocccsize) should show the repo release version


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

Observe the `mkiocccentry -V` command:

```sh
$ mkiocccentry -V
mkiocccentry version: 2.3.0 2015-11-30
mkiocccentry repo release: 2.11.0 2015-11-30
soup library version: 2.4.0 2015-11-30
jparse utils version: 2.1.7 2025-11-02
jparse UTF-8 version: 2.1.3 2025-07-04
jparse library version: 2.4.3 2025-11-10
```

But look at the other tools:

```sh
$ for i in txzchk chksubmit chkentry ; do "$i" -V |grep repo ; done
$
```

The other less important (to most contestants) tools should also have the repo version (commands including the test commands and also the commands like location(1) and anything else you, @lcn2, and I wrote).

I would also argue that the repo release line should be either last or first.

The benefit of having these in all the commands is one can more easily look at the release and make SURE they have it installed. I have another thought for an enhancement that might be related but one I had thought of at first but decided against: I have an idea that might work instead (whereas the other one would not). I'll open a new issue.

### Anything else?

I might also say that I'm making https://github.com/ioccc-src/mkiocccentry/pull/1372 a DRAFT until after IOCCC29 concludes. Draft because that way I don't have to redo the commit. Another issue coming up after this one.

xexyl:
First thought: should the repo release version be ABOVE everything else?

Second thought: do we need this in any of the tools? Perhaps yes and perhaps not.

Third thought: if it is deemed worth it then we probably want it in the usage string as well as the version string.

Although I fixed an issue today I doubt I'll be up to this today especially since that other one was a genuine bug and this one is unclear to me.

I should guess that this PROBABLY is a good idea but there are still minor questions to consider (such as my first thought above).

lcn2:
> First thought: should the repo release version be ABOVE everything else?

No: the tool itself should have top billing.

Perhaps a good order is:

0) the tool version
1) the repo version
2) any relevant linked library versions (? - perhaps this is overkill?)

> Second thought: do we need this in any of the tools? Perhaps yes and perhaps not.

[...]

PR #1390, opened 2026-06-07
Resolve issues #1389, #1379, #1373

xexyl:
Remove the 'total file size' check in mkiocccentry. This does not work as 
intended because it does not account for the JSON files and it does not account 
for the compression (in fact the total file size is validated through txzchk 
and it's a fatal error then but if it was before that it would mean users would 
have to, assuming no answers file, input data again if say on the first time 
they had this problem, and that is just one issue amongst others). 
 
Updated tools (except mkiocccentry which already did this) to show in -V and 
-h option (with the latter it is shown with any usage error) the repo version 
as well (right under the tool's version and above the libraries). The tools are: 
txzchk, fnamchk, location, chkentry and chksubmit. 
 
Type fix in bug_report.sh. 
 
Updated MKIOCCCENTRY_VERSION to "2.3.2 2026-06-07". 
Updated MIN_MKIOCCCENTRY_VERSION to MKIOCCCENTRY_VERSION. This is because 
we're in between contests now and the next contest will require the current 
version. 
 
Because the only other code changes were to print the repo version those tools 
did not have their version updated. The only reason mkiocccentry had its 
version updated is because a functional change was made.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,33 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 2.11.3 2026-06-07
+
+Resolve issues #1389, #1379, 1373.
+
+Remove the 'total file size' check in `mkiocccentry`. This does not work as
+intended because it does not account for the JSON files and it does not account
+for the compression (in fact the total file size is validated through `txzchk`
+and it's a fatal error then but if it was before that it would mean users would
+have to, assuming no answers file, input data again if say on the first time
+they had this problem, and that is just one issue amongst others).
+
+Updated tools (except `mkiocccentry` which already did this) to show in `-V` and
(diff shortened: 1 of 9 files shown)


xexyl:
I know that this is three issues in one but ... only one was functionality change: unless you consider showing versions of the repo functionality change. I mean technically it is and it's why I mentioned it but it's barely worth noting. It certainly is not worth updating a version itself.

I might add (forgot to say) that by accident one of us (surely not me :-) ) changed two tool versions to have 2015 not 2025. They were changed to 2025. Well at least one but I think two (the `chk*` tools).

lcn2 (approved):
👍 and thanks 🙏 @xexyl 

p.s. Congratulations on your **IOCCC29** winning entry.

PR #1391, opened 2026-06-10
Sync jparse repo

xexyl:
Sync jparse repo (https://github.com/xexyl/jparse). The change is mis-feature to misfeature (this is actually in OED). There are other repos and files here that should be corrected, as insignificant as it probably is. These can be done as important fixes are made, of which some are pending (and some have been started but which will likely not be done for a while).

A minor insignificant change in soup/ was made but since it was updated it still suggests a change in version so: updated SOUP_VERSION to "2.4.0 2026-06-10".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,18 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 2.11.4 2026-06-10
+
+Sync [jparse repo](https://github.com/xexyl/jparse). The change is `mis-feature`
+to `misfeature` (this is actually in OED). There are other repos and files here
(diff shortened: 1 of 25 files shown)


xexyl:
I do not intend to do more here today. Indeed I mean not to. 

Whether I do get to anything or not I do not know but I have other things I wish to do and things (of course) I must do and I don't exactly feel up to doing more here.

I did start work on another issue yesterday, although that's the low priority issue. The top priority issue I do not feel up to doing nor should I even try when I'm as tired as I am (which does fluctuate throughout the day of course but I have additional things I must do today and I likely would not feel up to it anyway).

lcn2 (approved):
"feature" 😉👍

Issue #1378, opened 2026-02-25
Enhancement: consider a sanity check on workdir and topdir before doing make clobber


lcn2:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the enhancement

As mentioned by @TheEndCrystal28 in Discord:

> There's an annoying feature where if you accidentally swap workdir and topdir in the arguments, the tool runs or in case both `workdir/remarks.md` and `topdir/remarks.md`(and fails) before checking that workdir and topdir are in the correct order.

### Relevant links

https://discord.com/channels/1327029557718417438/1332205820925972541/1476210418065084577

### Anything else?

Perhaps a simple sanity check on the args could be added?  Maybe if the `workdir` has a `prog.c` file, while the `topdir` does **NOT** have a `prog.c` file, an error suggesting a possible topdir-workdir arg mix-up,  could be raised and `mkiocccentry(1)` could exit non-zero? 

In case there is no `prog.c` file, or in case both `workdir/prog.c` and `topdir/prog.c` are found, then check for `remarks.md`.

In case there is no `remarks.md` file, or in case both `workdir/remarks.md` and `topdir/remarks.md` are found, then check for `Makefile`.

If after all that, there is no `Makefile`, or in case both `workdir/remarks.md` and `topdir/remarks.md` are found, then assume the user knows that they are doing and proceed.

We don't need to go overboard with this sanity check.  The above 3 cases are very likely good enough.

## UPDATE 0

It is probably a good idea to do these sanity checks very early on: particularly before prompting the user or running make.

xexyl:
Oh dear. This is not necessarily a bad idea.

I did check (not sure if it's still there) that the answers file is not put in the tarball but this might be worth considering too.

I am afraid I am much too tired to even consider ideas but you're welcome to assign this to me. Obviously for after this IOCCC. And who knows if I'll manage it before my surgery. Maybe during judging but then you might be absent from discussion at the time.

xexyl:
First point: yes before any prompt in fact. Right after getting the absolute paths.

Presumably we'd use fts_walk().

The question is do we actually swap them? I personally think that would be a mistake. Again this is MY opinion but I believe that if we did that it would be making blatant assumptions and possibly destructive assumptions. Frankly I could see how running make clobber at all is destructive. It's probably pretty rare but I could imagine a scenario where someone accidentally ends up deleting their prog.c or other important files. That would be terrible.

[...]

PR #1393, opened 2026-06-16
Resolve issue #1378 and fixed related issues

xexyl:
The code now does some simple checks to try and ascertain if the user swapped the topdir and workdir on the command line by checking for prog.c, Makefile, remarks.md in the workdir but not in the topdir. It's okay for them to exist in both because in the workdir a subdirectory is created and the files that will be copied there come from topdir (and there is nothing we can do in this case: if the user provided the wrong order and the files differ but they exist in both directories we can't know this). If they do not exist in either we do not report an earlier error because shortly after the walking code will report the error. The related issues are two and are below.

For realpath(3) we used PATH_MAX. However one can construct a path longer than this and although it's unlikely it still should be considered: especially since that's the directory names and did not consider files under them. POSIX allows realpath()'s second arg to take NULL and it will allocate memory in that case.

Also since we do use realpath(3) we did not need to use getcwd(3) for the topdir (it used both the result of getcwd(3) and realpath(3) and the former for topdir was used in places where the realpath(3) for the workdir counterpart is used).

Updated MKIOCCCENTRY_VERSION to "2.3.4 2026-06-16".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,32 @@
 # Major changes to the IOCCC entry toolkit
 
+Resolved issue #1378 and fixed some other issues (somewhat related) that I
+found (as below).
+
+The code now does some simple checks to try and ascertain if the user swapped
+the topdir and workdir on the command line by checking for prog.c, Makefile,
+remarks.md in the workdir but not in the topdir. It's okay for them to exist in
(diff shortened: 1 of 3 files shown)


lcn2 (approved):
👍 and 🙏 @xexyl

Issue #1394, opened 2026-06-17
Bug: code does not allow prog.c, remarks.md or Makefile to exist under subdirectories


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

[Note that I have already fixed t his but I am opening this issue to more easily refer to it. Another bug was found in the process (also fixed) and I will open an issue for that bug for the same reason, prior to committing the fixes.]

As the subject says: if one has say workdir/foo/{prog.c,Makefile,remarks.md} (one, two or three of these files) then it rejects it at the end as a manifest error.

### What you expect

As discussed yesterday 16 June 2026 by design the code should NOT forbid this.

### Environment

Wrong! Do it again!
Wrong! Do it again!
Wrong! Do it again!

:-)

### Anything else?

I will commit the fix after opening the other issue (also fixed already) and then updating the version and changelog.

xexyl:
I believe this issue should be resolved with https://github.com/ioccc-src/mkiocccentry/pull/1396/commits/eea56a268f3129f1d243b2a6e81f4afef1817ed9.

If not please let me know. In any case I will not be looking at it today.

lcn2:
Thank you @xexyl

Issue #1395, opened 2026-06-17
Bug: `mkiocccentry` does not check case-insensitively for Makefile, prog.c, remarks.md and it does not check depth


xexyl:
### Is there an existing issue for this?

- [x] I have searched for existing issues and did not find anything like this

### Describe the bug

[I have already fixed these two related bugs but am opening this to easily refer to the problem in the changelog. Commit which fixes this issue and the related issue #1394 will follow in a bit.]

The consensus a long while back was that because some file systems default to case-insensitive (because they're not so insolent?) such as those under macOS file systems we need to check specific filenames case-insensitively: all the more since `make(1)` checks both `Makefile` and `makefile` as well as others. Plus it would be confusing.

Also the code does not check the level so that if one has say topdir/foo/Makefile it would run the `check_Makefile()` on both which can be confusing: only the top level Makefile needs to be checked (same with prog.c and remarks.md). This has also been fixed.

### What you expect

If `makefile` and `Makefile` (for example) are included in the same directory it should be flagged as an error.

Any Makefile not in the top level directory should NOT be checked with `check_Makefile()`.

### Environment

No.

### Anything else?

Maybe.

xexyl:
I believe this issue should be resolved with https://github.com/ioccc-src/mkiocccentry/pull/1396/commits/eea56a268f3129f1d243b2a6e81f4afef1817ed9.

If not please let me know. In any case I will not be looking at it today.

lcn2:
Thanks @xexyl

PR #1396, opened 2026-06-17
Resolved issues #1394 and #1395

xexyl:
The mkiocccentry code did not allow prog.c, Makefile, remarks.md, try.sh and try.alt.sh to exist in subdirectories, as the maximum allowed is 1 but it did not check depth before writing e.g. "remarks": "remarks.md" etc.). This was actually appearing to be a problem with the manifest but that was because mkiocccentry was creating it (the manifest) wrong. It should write it as the specific names (rather than an extra file) if the level < 2 (which means it's in the top level directory).

A related problem is that it did not check for these filenames case-insensitively which meant that if a person submitted a directory with Makefile and makefile it would not be flagged as two of the same which it is supposed to be not only for file systems which are case-insensitive (default in macOS) but also because it could cause problems in entries (for instance GNU make(1), which is what the contest uses, checks for Makefiles in the order: GNUmakefile, makefile and Makefile, which obviously is a problem) and the website (in various ways).

Somewhat related to the above two problems is that if the user had e.g. foo/Makefile it would run check_Makefile() on both Makefile and foo/Makefile which is not only incorrect but it is confusing as well.

Updated MKIOCCCENTRY_VERSION to "2.3.5 2026-06-17".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,35 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 2.11.5 2026-06-17
+
+Resolved issues #1394 and #1395.
+
+The `mkiocccentry` code did not allow `prog.c`, `Makefile`, `remarks.md`,
+`try.sh` and `try.alt.sh` to exist in subdirectories, as the maximum allowed is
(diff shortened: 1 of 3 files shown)


lcn2 (approved):
🙏 @xexyl

PR #1398, opened 2026-06-27
Fix bug in FTS walking code

xexyl:
Fix bug in fts_icmp() (in soup/walk_util.c): it used strcmp() instead of strcasecmp().

Updated SOUP_VERSION to "2.4.2 2026-06-27".

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -51,7 +51,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v6
+      uses: actions/checkout@v7
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v7
     - name: install ctags
       run: sudo apt-get install -y universal-ctags
     - name: install flex
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Major changes to the IOCCC entry toolkit
 
+## Release 2.11.6 2026-06-27
+
+Fix bug in `fts_icmp()` (in `soup/walk_util.c`): it used `strcmp()` instead of
+`strcasecmp()`.
+
+Updated `SOUP_VERSION` to `"2.4.2 2026-06-27"`.
+
+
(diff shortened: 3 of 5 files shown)


xexyl:
QUESTION for you @lcn2.

I don't recall but since we actually do NOT care about case in filenames (due to macOS for example) do we want to use `fts_cmp()` or `fts_icmp()`?

If we don't want to use `fts_icmp()` I'd be happy to change this so that the function `fts_icmp()` does not even exist - unless you think we might need it in the future. On the other hand since it doesn't harm anything having it in it might just be better to not worry about it?

Let me know.

Nonetheless probably this is all I'll do today (and happily there doesn't seem to be anything else to do, although I have some minor fixes to SECURITY.md - just haven't bothered committing them).

lcn2:
> don't recall but since we actually do NOT care about case in filenames (due to macOS for example) 

Correct.

> do we want to use `fts_cmp()` or `fts_icmp()`?

Seems like `fts_icmp()` to fit with the previous question.

lcn2 (approved):
👍

xexyl:
> > don't recall but since we actually do NOT care about case in filenames (due to macOS for example) 
> 
> 
> 
> Correct.
> 
> 
> 
> > do we want to use `fts_cmp()` or `fts_icmp()`?
> 
> 
> 
> Seems like `fts_icmp()` to fit with the previous question.
> 
> 

Want me to make the change ?

If so I will try doing that tomorrow or the next day.

xexyl:
Now that I think on it .. is there any point in changing it? We made sure that if a file exists in the same directory and there is another file of a different case in the same directory that it's an error:

```
Checking your submission directory for various issues ...
Warning: test_manifest: invalid: manifest shell[1] filename matches manifest shell[0] filename
follows are errors for JSON file: .info.json
JSON semantic validation error ERROR[148]: chk_manifest: manifest is missing required files and/or has invalid permissions/missing and/or has invalid/duplicate extra_file filenames 
.info.json total JSON semantic error(s) found: 1
```

So changing it to use `fts_icmp()` seems like a waste of time.

It would also be a waste of time to remove the function even though it's not used esp as (although this is extremely unlikely) it could be in theory useful in the future (again extremely unlikely given what it's for but even so it would be a waste of time doing so - just an extra commit and pull request for no gain).

Am I right that you agree with these assessments @lcn2 ?

xexyl:
Unfortunately I found a bug (I think) ... \*sigh\* ... I really wanted to read again today.

I'll open an issue and see if it's worth it ... though might look at it first. I doubt I can do much of it now if at all today in which case I'll just open an issue ...

PR #1403, opened 2026-07-04
Add bad test file for txzchk

xexyl:
Added bad test file for txzchk. With the recent commits it now correctly fails.


xexyl:
I believe that the three issues can probably be closed now, @lcn2?

It's possible we missed something but if so I am unable to right now determine that. So if you think it's good after this then I would close them. It does seem to catch the edge cases I thought of and I appreciate you updating the code as I am barely able to do anything right now!

lcn2 (approved):
🙏 @xexyl

